Issue #709, opened 2024-12-18
code: Use a proper test app in integration tests instead of the mock Starknet app


romac:
Depends on #740

PR #747, opened 2025-01-10
feat(code/test): Add channel-based test app for integration tests

romac:
Closes: #709


romac:
Let's merge this already to unblock https://github.com/informalsystems/malachite/pull/836. A couple tests are still failing with the test app, I will fix those in a follow-up PR.

code/crates/test/app/src/app.rs
```diff
@@ -0,0 +1,317 @@
...
+                // with crash recovery and is not strictly necessary in this example app since all our state
+                // is kept in-memory and therefore is not crash tolerant at all.
+                if let Some(proposal) = state.get_previously_built_value(height, round).await? {
+                    info!(value = %proposal.value.id(), "Re-using previously built value");
+
+                    if reply.send(proposal).is_err() {
+                        error!("Failed to send GetValue reply");
+                    }
+
+                    return Ok(());
```
ancazamfir, line 92: In principle we should restream the parts here again.

code/crates/test/app/src/app.rs
```diff
@@ -0,0 +1,317 @@
...
+                }
+
+                // If we have not previously built a value for that very same height and round,
+                // we need to create a new value to propose and send it back to consensus.
+                let proposal = state.propose_value(height, round).await?;
```
ancazamfir, line 97: The way we build the block/ parts vs propose value is strange. The normal order should be:
- create block
- stream parts
- call `propose_value()` and determine the proposal based on the block created before
I understand this app constructs the parts based on the primes of the randomly generated value (u64) but maybe we can hide that deeper.

code/crates/test/app/src/state.rs
```diff
@@ -0,0 +1,385 @@
...
+
+        parts
+    }
+}
+
+/// Re-assemble a [`ProposedValue`] from its [`ProposalParts`].
+///
+/// This is done by multiplying all the factors in the parts.
+fn assemble_value_from_parts(parts: ProposalParts) -> ProposedValue<TestContext> {
+    let value = parts
```
ancazamfir, line 340: Normally there should be a check that the received parts put together match some `ValueId` the proposer computed. I would put the `u64` in the Fin and double check it matches  `value` here. Alternatively, we should hash the parts and make sure they match the current hash in the Fin.
It's not super important for the use of this app in tests but rather if it's looked at as tutorial.

code/crates/test/app/src/state.rs
```diff
@@ -0,0 +1,385 @@
...
+        &mut self,
+        certificate: CommitCertificate<TestContext>,
+    ) -> eyre::Result<()> {
+        let Ok(Some(proposal)) = self
+            .store
+            .get_undecided_proposal(certificate.height, certificate.round)
+            .await
+        else {
+            return Err(eyre!(
+                "Trying to commit a value at height {} and round {} that is not decided: {}",
```
ancazamfir, line 148:
```suggestion
                "Trying to commit a value at height {} and round {} for which there is no proposal: {}",
```

code/crates/test/app/src/state.rs
```diff
@@ -0,0 +1,385 @@
...
+                certificate.height,
+                certificate.round,
+                certificate.value_id
+            ));
+        };
+
+        self.store
+            .store_decided_value(&certificate, proposal.value)
+            .await?;
+
```
ancazamfir, line 158: Maybe we should remove value from the undecided table.

ancazamfir (commented):
Looks good in general, left some comments, we could address in subsequent PRs.

code/crates/test/app/src/state.rs
romac: We already check the signature before assembling the parts, I've refactored it slightly in 38843d80.

romac: Done in a6b77ec6

code/crates/test/app/src/app.rs
romac: Done in c226ec30

code/crates/test/app/src/state.rs
romac: https://github.com/informalsystems/malachite/pull/747/commits/0971317dbc1c0f0276bbf891bc45c4f854c1c4a8

romac:
The only remaining failing test is

```
    informalsystems-malachitebft-test it value_sync::start_late: failure
```

Spent an hour debugging it to no avail. Let's merge this and fix that test in a follow-up PR.

Issue #510, opened 2024-11-01
spec: Formally define the `valid(v)` properties


cason:
When executed on a correct node `valid(getValue())` should always return true. For some application we discussed to lift this property, by requiring that we don't need this always, but only from some round on. This is related to whether `valid` should be a function or a relation... if it is not a function, at least we need a strong enough link to `getValue` to ensure liveness.

_Originally posted by @josef-widder in https://github.com/informalsystems/malachite/pull/389#discussion_r1771312835_

The `valid(v)` function is supposed to be deterministic. In particular, if `v` is produced and proposed by a correct proposer, then `valid(v)` should return `true` at every correct process.

Definition of done: add details to [this section](https://github.com/informalsystems/malachite/blob/main/specs/consensus/overview.md#validation).

josef-widder:
A different way to look at soft-upgrades is as follows: 
The idea is that if `valid` should be changed in that from height `h`on, a function `new_valid` should be used to check validity, while up to height `h-1` the function `old_valid` has been used, then we can get something along these lines:
```
valid(v, chain, height) = 
  if (height >= h)
    new_valid(v, chain, height)
  else
    old_valid(v, chain, height)
```

In this way, we effectively split evaluation of `valid` into different height 
intervals. Whenever a new version is needed, and new interval is added on top. Moreover
if in the future validity of a value is checked with a new version of `valid`, effectively
the logic that was in place for that height is used by using the proper "if" branch. 
This addresses Requirement 1 and 2 from above.

PR #748, opened 2025-01-10
feat(spec): Requirements on valid(v)

josef-widder:
Closes: #510

---

#### For all contributors

- [ ] Reference GitHub issue


specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+3. If invoked on different processes for the same input, the function always 
+returns the same boolean value. 
+
+The correctness of the consensus algorithm depends heavily on these points. 
+Most obviously for termination, we require that all correct processes find 
+the value by a correct proposer valid under all circumstances, because we
+need them to all prevote for the value. A deviation in `valid` would 
+result in some processes
+prevoting nil, and consequently, there might never be enough votes for a 
+value to decide.
```
cason, line 465:
```suggestion
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+
+#### `valid(v)` in implementations
```
cason, line 467:
```suggestion
#### Implementation
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+
+In implementations, the validity check typically is not pure:
```
cason, line 469:
```suggestion
Implementations of the validity check are typically not pure:
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+as already mentioned in the arXiv paper "In the context of blockchain systems,
```
cason, line 470:
```suggestion
as already mentioned in the [Tendermint paper][tendermint-arxiv], "In the context of blockchain systems, 
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+prevoting nil, and consequently, there might never be enough votes for a 
+value to decide.
+
+#### `valid(v)` in implementations
+
+In implementations, the validity check typically is not pure:
+as already mentioned in the arXiv paper "In the context of blockchain systems, 
+for example, a value is not valid if it does not
+contain an appropriate hash of the last value (block) added to the blockchain." That is, 
+rather than `valid(v)` the implementation looks something like `valid(v, chain, height)`,
```
cason, line 473: Regarding `valid(v, chain, height)`, I agree with the discussion but I would use the nomenclature of the pseudo-code.

Instead of `chain`, we have the `decision_p` array for process `p`, and instead of `height` we use `h_p`.

The height here can actually be omitted, as by construction if `h_p = H`, then `decision_p` is filled from `0` to `H-1`. In other words, `len(decision_p) = h_p` is an invariant.

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+the algorithm. The purpose is to restrict the domain of values that consensus 
+may decide. There is the assumption that if `v` is a value returned
+by the  `getValue()` function of a correct process in the current height, then
+`valid(v)` evaluates to `true`. Under this assumption, as long as the proposer
+is correct, `valid(v)` does not serve any purpose. It only becomes relevant if
+there is a faulty proposer; it limits "how bad" proposed values can be.
+
+In the arXiv paper, the pseudo code uses `valid(v)`, that is, a function that
+maps a value to a boolean. Observe that this should be understood in terms
+of a mathematical (pure) function:
```
cason, line 450:
```suggestion
of a mathematical (pure) function, with the following properties:
```

cason: So we can point to the "properties above" instead.

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+
+In implementations, the validity check typically is not pure:
+as already mentioned in the arXiv paper "In the context of blockchain systems, 
+for example, a value is not valid if it does not
+contain an appropriate hash of the last value (block) added to the blockchain." That is, 
+rather than `valid(v)` the implementation looks something like `valid(v, chain, height)`, 
+that is, 
+it depends on the state of the blockchain at a given height and a value. This implies that
+a value that might be valid at blockchain height 5, might be invalid at height 6. Observe 
+that this, strictly speaking, violates Point 1. from above.
```
cason, line 477:
```suggestion
that this, strictly speaking, the above defined property 1.
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+However, as long as all processes agree on 
+the state of the chain up to height 5 when they 
+start consensus for height 6, this still satisfies Points 2 and 3 and it will not harm
```
cason, line 480:
```suggestion
start consensus for height 6, this still satisfies properties 2 and 3 and it will not harm 
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+a value that might be valid at blockchain height 5, might be invalid at height 6. Observe 
+that this, strictly speaking, violates Point 1. from above.
+However, as long as all processes agree on 
+the state of the chain up to height 5 when they 
+start consensus for height 6, this still satisfies Points 2 and 3 and it will not harm 
+liveness. (There have been cases of 
+non-determinism in the application level that led to processes disagreeing on the 
+application state and thus consensus being blocked at a given height)
+
+**Remark.** We have seen slight (ab)uses of valid, that use external data. Consider he toy
```
cason, line 485: I will use a quote block (`> `) for this paragraph.

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+decided some time ago). So this use is not encouraged, and we ignore it in the 
+remainder. However, for some applications this use case might still be beneficial: in
+this case it is important to understand that one needs to make an argument (which
+also needs to involve
+the environment) why it
+can be ensured that if `v` is a value returned
+by the  `getValue()` function of a correct process in the current height, then
+`valid(v)` evaluates to `true`.
+
+**Remark.** Point 1 above also forbids that the current consensus round influences
```
cason, line 501: I will use a quote block (`> `) for this paragraph.

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+validity. For instance, one may say to get out of the liveness issue from the previous
+remark, to just find all blocks valid that have been proposed after, say, round 10. 
+Similarly, one might consider more stricter validity requirements for the proposal in
+round 0 (the lucky path), while in unlucky situations one might want to simplify 
+reaching a decision by weakening validity. In general this leads to complexity down
+the road when someone reads the decisions much later, and needs to understand the 
+different semantics of different blocks based on the decision round. We strongly
+suggest to not use round numbers in validation of values for this reason.
+
+#### Backward compatibility of `valid(v)`
```
cason, line 511:
```suggestion
#### Backwards compatibility
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+
+**Requirement 1 (Fixing bugs).**
+There might be a bug in `valid`. Then it might be possible that due to a bug,
```
cason, line 514:
```suggestion
There might be a bug in the implementation of `valid(v)`. Then it might be possible that due to a bug,
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+`valid` returns `false` 
+for a value `v` proposed by a correct process, and we are stuck at a given height.
```
cason, line 516:
```suggestion
for values proposed by correct processes, and we are stuck at a given height. 
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+suggest to not use round numbers in validation of values for this reason.
+
+#### Backward compatibility of `valid(v)`
+
+**Requirement 1 (Fixing bugs).**
+There might be a bug in `valid`. Then it might be possible that due to a bug,
+`valid` returns `false` 
+for a value `v` proposed by a correct process, and we are stuck at a given height. 
+A way to get out of the 
+situation is to produce a new implementation `valid'` that returns `true` for `v`. So
```
cason, line 518:
```suggestion
situation is to produce a new implementation of `valid(v)` that returns `true` for the values proposed by correct processes.
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+to be prepared for such a scenario we need to allow a change in the function.
```
cason, line 519:
```suggestion
To be prepared for such a scenario we need to allow a change in the function.
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+
+**Requirement 2 (Future use).**
+So if we allow changes to `valid`, we need to understand all uses of this function: Some
```
cason, line 522:
```suggestion
If we allow changes to `valid`, we need to understand all uses of this function. Some
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+synchronization protocols may use `valid` for consistency checks, for instance, if a node
```
cason, line 523:
```suggestion
synchronization protocols may use `valid(v)` for consistency checks, for instance, if a node
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+fell behind, it might need to learn several past decisions. In doing so, it typically also
+uses (the current version) of `valid` to check the decided values before accepting them.
```
cason, line 525:
```suggestion
uses (the current version) of `valid(v)` to check the decided values before accepting them.
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+In this scenario, a value decided in the past (potentially using a now old version of 
+`valid`) should be deemed valid with the current version of the function.
```
cason, line 527:
```suggestion
`valid(v)`) should be deemed valid with the current version of the function. 
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+fell behind, it might need to learn several past decisions. In doing so, it typically also
+uses (the current version) of `valid` to check the decided values before accepting them.
+In this scenario, a value decided in the past (potentially using a now old version of 
+`valid`) should be deemed valid with the current version of the function. 
+
+These two requirements lead us to the following requirement on the implementations:
+we consider a sequence of `valid_i(v)` implementations, with increasing indexes `i`, 
+so that to represent multiple _backwards compatible_ implementations of the validity checks.
+Formally we require that
+
```
cason, line 533:
```suggestion
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+`valid_i(v) == true` implies `valid_j(v) == true `, for `j > i` and every value `v`
```
cason, line 534:
```suggestion
`valid_i(v) == true` implies `valid_j(v) == true `, for `j > i` and every value `v`.
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+
+The logical implications allows to be more permissive in later versions (as required to
```
cason, line 536:
```suggestion
The logical implication allows newer versions of `valid(v)` to be more permissive (as required to
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+fix the bug), while maintaining the need that newer versions allow to check validity
```
cason, line 537:
```suggestion
fix bugs), while ensuring that newer versions allow to check validity
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+**Requirement 2 (Future use).**
+So if we allow changes to `valid`, we need to understand all uses of this function: Some
+synchronization protocols may use `valid` for consistency checks, for instance, if a node
+fell behind, it might need to learn several past decisions. In doing so, it typically also
+uses (the current version) of `valid` to check the decided values before accepting them.
+In this scenario, a value decided in the past (potentially using a now old version of 
+`valid`) should be deemed valid with the current version of the function. 
+
+These two requirements lead us to the following requirement on the implementations:
+we consider a sequence of `valid_i(v)` implementations, with increasing indexes `i`,
```
cason, line 530:
```suggestion
we consider a sequence of `valid_i(v)` implementations, with increasing versions `i`, 
```

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+so that to represent multiple _backwards compatible_ implementations of the validity checks.
+Formally we require that
+
+`valid_i(v) == true` implies `valid_j(v) == true `, for `j > i` and every value `v`
+
+The logical implications allows to be more permissive in later versions (as required to
+fix the bug), while maintaining the need that newer versions allow to check validity
+of previously decided values.
+
+#### Backward compatibility of `valid(v, chain, height)`
```
cason, line 540: I see this section as an example of the requirement of the previous section.

Although I agree that it is only a legit example if `v` somehow included a height or `h_p` is also known by the `valid(v)` implementation. Not sure if we need this. Maybe a pointer to issue 510 is enough here.

specs/consensus/overview.md
```diff
@@ -437,18 +437,134 @@ height of consensus `h_p`.
...
+
+In this way, we effectively split evaluation of `valid` into different height 
+intervals. Whenever
+a new version is needed, and new interval is added on top. Moreover
+if in
+the future validity of a value is checked with a new version of `valid`, effectively
+the logic that was in place for that height is used by using the proper "if" branch. 
+This addresses Requirement 1 and 2 from above.
+
+TODO: I think in this [remark](https://github.com/informalsystems/malachite/issues/510#issuecomment-2505632410), different use cases (soft upgrades) are mixed.
```
cason, line 565: Indeed, the example in the code above is almost the definition of software upgrades.

cason: The linked comment is a short-back of the code snippet and the software upgrades approach. This is the case where a bug identified during a given height forces us to the most general approach described in the "Backwards Compatibility" section.

cason (approved):
Great work.

Most of the suggestions I made were to align the content of this section with the "format" used in the remainder of this document.

Regarding the last sub(*)section, from line 540, I think it should not be part of this document. The same for the linked commend in issue 510.

They end up being too much specific for the level of detail that is considered in this document.

josef-widder:
> Regarding the last sub(*)section, from line 540, I think it should not be part of this document. The same for the linked commend in issue 510.
> 
> They end up being too much specific for the level of detail that is considered in this document.

I agree. We might do another md file with implementation-specific discussions, and link it. 

More generally, I find the document very long. I wonder whether we should cut it in multiple shorter ones that might be easier to navigate if we have a good top-level document...

cason:
> More generally, I find the document very long.

For an "overview" document it is definitely, ehe.

But I wonder whether we should try to split it now or first have on it all content we find relevant, then do the splitting?

josef-widder:
> For an "overview" document it is definitely, ehe.
> 
> But I wonder whether we should try to split it now or first have on it all content we find relevant, then do the splitting?

Let's fix the TODO on soft upgrades on our call tomorrow. Then merge, and open an issue about splitting.

specs/consensus/overview.md
josef-widder: I removed the section and put the discussion in the linked issue

specs/consensus/overview.md
josef-widder: I updated the discussion accordingly.

Issue #742, opened 2025-01-08
feat(code/consensus): Add support for full nodes


romac:
## Background

> All nodes have a validator key. A validator node is a node that has its validator key included in the "current" validator set.
> 
> A validator node signs and broadcasts consensus messages. Every node gossip consensus messages. Every processes sends and receives consensus control/gossip messages.
> 
> Validator nodes may have a more restrictive network setup. But this is not mandatory.
> 
> So, being a validator is not a characteristic, it is a property. A node can be a validator in height `H` but not in height `H + 1`. This is application dependent.
> 
> All nodes process messages in the same way. But when producing a message, a node whose validator key is not in the current validator set implements a noop. An active validator sign the message and "broadcast" it.
> 
> by @cason

## Implementation

Change the logic in `malachitebft-consensus` to not emit `Sign*` and `Publish*` effects when we are not in the validator set.

## Testing

Update the test framework so that we can mark some nodes as not being part of the validator set (turning them into full nodes) ideally with a nice fluent API, e.g.:

```rust
let mut test = TestBuilder::<()>::new();

// validator node (in validator set, voting power 1 by default)
test.add_node().start().wait_until(5).success();

// validator node (in validator set, voting power 5)
test.add_node().with_voting_power(5).start().wait_until(5).success();

// full node (not in validator set)
test.add_node().full_node().start().wait_until(5).success();

test.build().run(Duration::from_secs(30)).await
```

hoank101:
hi, i can do it

romac:
Awesome, thank you! 🙌

PR #750, opened 2025-01-11
feat(code/consensus): Add support for full nodes

hoank101:
Closes: #742

---

#### For all contributors

- [x] Reference GitHub issue


code/crates/starknet/test/src/lib.rs
```diff
@@ -739,10 +759,6 @@ pub fn make_node_config<S>(test: &Test<S>, i: usize) -> NodeConfig {
     }
 }
 
-fn voting_powers<S>(nodes: &[TestNode<S>]) -> Vec<VotingPower> {
```
romac, line 742: Let's keep this function and change the logic accordingly

romac: Last nitpick and then we can merge I think :)

romac (approved):
Great work! 🚀

code/crates/core-consensus/src/handle/driver.rs
```diff
@@ -194,36 +194,39 @@ where
...
-                        signed_proposal.height(),
-                        signed_proposal.round(),
-                        signed_proposal.pol_round(),
-                        signed_proposal.validator_address().clone(),
-                        signed_proposal.value().id(),
-                        Default::default()
-                    )
-                );
+            // Only sign and publish if we're in the validator set
+            if state.is_validator() {
```
ancazamfir, line 198: Not sure about the proposal filtering, seems we are trying to defend against a bug where `get_proposer()` returns a non-validator/ full node address...but in this case the `is_validator()` should also fail.

ancazamfir:
Wondering if we should send `VoteSetRequest` to a non-validator, or accept `VoteSetResponse` from a non-validator.

ancazamfir:
So for the receive side, `verify_signed_vote()` and `verify_signed_proposal()` handle the case where a full node sends messages and we will see warning messages of the form: `Received vote/proposal from unknown validator`. Maybe `non-validator` would be better but really minor.

romac:
> Wondering if we should send VoteSetRequest to a non-validator, or accept VoteSetResponse from a non-validator.

Good question, perhaps there is a higher chance that non-validators might be byzantine and in that case we should restrict those queries to validators only?

romac:
> Can we start a full node from command line?

No, you have to remove the full node address from the validator set in the genesis of all validator nodes.

code/crates/core-consensus/src/handle/driver.rs
romac: Yes it's just defensive coding. Maybe we can change the check to do a `debug_assert` or even an `assert` in order to ensure this code path is never hit.

ancazamfir:
> > Wondering if we should send VoteSetRequest to a non-validator, or accept VoteSetResponse from a non-validator.
> 
> Good question, perhaps there is a higher chance that non-validators might be byzantine and in that case we should restrict those queries to validators only?

Not sure myself, let's discuss next time we meet. On a side note, we should allow vote sync when blocksync is diabled (iirc this is not the case today)

PR #760, opened 2025-01-14
feat(spec): discussion of consensus `getValue()` function

cason:
Part of  #761.

---

#### For all contributors

- [x] Reference GitHub issue

### DIFF
--- a/specs/consensus/overview.md
+++ b/specs/consensus/overview.md
@@ -530,6 +530,9 @@ then `v` was the value returned by the context/application running at a process
 `p` when in invoked, as part of a round `r` where `proposer(h, r) = p`,  the
 `getValue()` function.
 
+> Observe that there is no function that passes decided values in the other direction. In Malachite's
+state machine design, this data is passed as consensus/driver output, 
+
 ### Validation
 
 The external function `valid` is used to guard any value-specific action in 
--- a/specs/consensus/overview.md
+++ b/specs/consensus/overview.md
@@ -229,7 +229,7 @@ thus `⟨PROPOSAL, h, r, validValue_p, validRound_p⟩`.
 
 If the proposer `p` of a round `r` of height `h` has `validValue_p = nil`, `p`
 may propose any value it wants.
-The function `getValue()` is invoked and returns a value to be proposed.
+The [function `getValue()`](#proposal-value) is invoked and returns a value to be proposed.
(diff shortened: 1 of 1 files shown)


specs/consensus/overview.md
```diff
@@ -492,13 +492,37 @@ selection algorithm becomes more challenging.
...
+implements the **propose** consensus primitive.
+Instead of having the context/application that uses the consensus protocol
+invoking it with a value to be proposed, when a process running Tendermint can
+propose a value, it invokes this function so that the context/application can
+provide the value to be proposed.
+
+> Notice that this is a design concern that is encoded in Tendermint's
+> pseudo-code.
+> A priori, every process running a consensus height could provide its proposed
+> value for that height, but this approach has some disadvantages:
```
josef-widder, line 510: I wonder whether we should add a bullet point on the Starknet usecase, where it is expensive to compute a proposal (if it must contain a ZK proof), and therefore only the scheduled proposers of round 0 should do that to optimize on the normal case.

specs/consensus/overview.md
```diff
@@ -492,13 +492,37 @@ selection algorithm becomes more challenging.
...
+> - Proposed values are likely to aggregate multiple inputs received from
+>   clients (e.g., transactions that form a proposed block).
+>   By providing the proposed value at the beginning of a height to a process
+>   that will only be the proposer of a high-numbered round, when this round is
+>   reached the proposal is likely to be incomplete or outdated.
+
+Since only proposed values can be decided, if `v` is decided at height `h`,
+then `v` was the value returned by the context/application running at a process
+`p` when in invoked, as part of a round `r` where `proposer(h, r) = p`,  the
+`getValue()` function.
```
josef-widder, line 525:
```suggestion
Since only proposed values can be decided, if `v` is decided at height `h`,
then `v` was the value returned by the context/application running at a process
`p` when in invoked, as part of a round `r` where `proposer(h, r) = p`,  the
`getValue()` function.

> Observe that there is no function that passes decided values in the other direction. In Malachite's
state machine design, this data is passed as consensus/driver output, 
```

josef-widder (approved):
Looks great. Thanks! I have left two comments...

specs/consensus/overview.md
cason: Added a bullet point with this example.

Issue #761, opened 2025-01-14
spec: conclude the consensus `overview.md` documentation


cason:
The Tendermint consensus algorithm [overview](https://github.com/informalsystems/malachite/blob/main/specs/consensus/overview.md) contains somehow our interpretation of the Tendermint paper and the associated implementation of the algorithm. It should not, however, dive into our implementation details.

However, it has some missing part and some `TODO`s:

- [x] #396
- [x] #760
- [x] #510
- [x] #764
- [x] #765
- [x] Stretch goal: #777

cason:
As also pointed out by @josef-widder, this document that should be an "overview" is getting super long and complex.

We should consider splitting it into parts, once its "full content" is defined.

cason:
I am closing this issue because the protocol overview has been read and reviewed by different people.

@josef-widder, feel free to open an issue if you still want to dismember it into smaller documents, with different levels of complexity (the real basic and details).

PR #764, opened 2025-01-14
feat(spec): Consensus network requirements and gossip property

cason:
Part of #761.

Rendered [section](https://github.com/informalsystems/malachite/blob/cason/761-broadcast/specs/consensus/overview.md#network)

---

#### For all contributors

- [x] Reference GitHub issue

### DIFF
--- a/specs/consensus/overview.md
+++ b/specs/consensus/overview.md
@@ -701,11 +701,11 @@ This property is not trivial to ensure.
 [comment]: <> (I wonder whether the next paragraph is more abou design)
 
 Fortunately, it turns out that the Gossip communication property does not need
-to be ensured for _every_ message broadcast by the algorithm.
+to be ensured for _every_ broadcast message.
 As discussed in [#260][synchronization-issue], only messages whose reception
 produce a relevant action in the consensus state-machine (in particular,
-produce a valid, locked, or decided value) of a correct process need to be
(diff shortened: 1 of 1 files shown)


specs/consensus/overview.md
```diff
@@ -651,11 +651,66 @@ be implemented by the processes running the consensus protocol.
...
+> correct processes, the other correct processes will relay it to them.
+> In the same way, if the sender is Byzantine and purposely does not send the
+> message to a subset of processes, the message will be relayed to them.
+>
+> The complexity added by Property 2 is the need for a protocol for reliably
+> broadcasting messages in a Byzantine setup.
+> In other words, the `broadcast` primitive cannot be just implemented by
+> sending the same message (via "unicast") to every process.
+
+[comment]: <> (I wonder whether the next paragraph is more design related)
```
josef-widder, line 701: I don't think it is design-related. It is related to the fact that there is wiggle room in the assumptions in the arXiv paper that we explore. The point is that there are two relevant assumptions on the message system
1. the one given in the paper (gossip)
2. the one actually needed by the algorithm in the paper (the one that you describe below)

So the point is that instead of considering 1, we actually care about 2, because it is realistic.

josef-widder (approved):
Thanks!

Issue #717, opened 2024-12-19
code: Rework vote extensions


romac:
The way vote extensions work right now is very cumbersome and to be frank quite confusing.

A vote extension is currently added to a precommit by looking up the corresponding full proposal assembled by the application and copying the extension found there into the precommit. This complicates the logic and mixes up the vote extension process with the building of the proposal. Moreover, and because that process is so twisty, I am not even sure that this is implemented either fully or correctly at the moment, cf. this [TODO](https://github.com/informalsystems/malachite/blob/1f16e9f32dfc9cffdea0d81e8a6746dbb3167330/code/crates/core-consensus/src/handle/proposal.rs#L83-L84).

Instead we should add a dedicated effect to the core consensus library which would be relayed to the application, allowing to explicitly extend its vote if it chooses to.

PR #775, opened 2025-01-16
feat(code): Vote extensions v2

romac:
Closes: #717 

---

## Questions

- [x] Should the app generate a plain `Extension` and then have it signed by consensus or generate a `SignedExtension` directly?
  - Let's have consensus sign and verify vote extensions

---

#### For all contributors

- [x] Reference GitHub issue

### DIFF
--- a/code/crates/starknet/host/src/host/proposal.rs
+++ b/code/crates/starknet/host/src/host/proposal.rs
@@ -64,10 +64,11 @@ async fn run_build_proposal_task(
     let build_duration = (deadline - start).mul_f32(params.time_allowance_factor);
 
     let mut sequence = 0;
-    let mut block_size = 0;
     let mut block_tx_count = 0;
     let mut block_hasher = sha3::Keccak256::new();
-    let mut max_block_size_reached = false;
+    let vote_extensions_size =
+        (params.vote_extensions.size.as_u64() * vote_extensions.extensions.len() as u64) as usize;
+    let mut block_size = vote_extensions_size;
 
     // Init
     let init = {
@@ -110,10 +111,9 @@ async fn run_build_proposal_task(
         let mut txes = Vec::new();
         let mut tx_count = 0;
 
-        for tx in reaped_txes {
+        'txes: for tx in reaped_txes {
             if block_size + tx.size_bytes() > max_block_size {
-                max_block_size_reached = true;
-                continue 'reap;
+                continue 'txes;
             }
 
             block_size += tx.size_bytes();
@@ -143,7 +143,7 @@ async fn run_build_proposal_task(
             sequence += 1;
         }
 
-        if max_block_size_reached {
+        if block_size > max_block_size {
             trace!("Max block size reached, stopping tx generation");
             break 'reap;
         } else if start.elapsed() > build_duration {
@@ -153,7 +153,7 @@ async fn run_build_proposal_task(
     }
 
     // Vote extensions
-    if !max_block_size_reached && !vote_extensions.extensions.is_empty() {
+    if !vote_extensions.extensions.is_empty() {
         let transactions = vote_extensions
             .extensions
             .into_iter()
--- a/code/crates/starknet/host/src/host/proposal.rs
+++ b/code/crates/starknet/host/src/host/proposal.rs
@@ -84,7 +84,9 @@ async fn run_build_proposal_task(
         init
     };
 
-    loop {
+    let max_block_size = params.max_block_size.as_u64() as usize;
+
+    'reap: loop {
         trace!(%height, %round, %sequence, "Building local value");
 
         let reaped_txes = mempool
@@ -102,10 +104,8 @@ async fn run_build_proposal_task(
         trace!("Reaped {} transactions from the mempool", reaped_txes.len());
 
         if reaped_txes.is_empty() {
-            break;
+            break 'reap;
         }
-
-        let max_block_size = params.max_block_size.as_u64() as usize;
 
         let mut txes = Vec::new();
         let mut tx_count = 0;
@@ -113,7 +113,7 @@ async fn run_build_proposal_task(
         for tx in reaped_txes {
             if block_size + tx.size_bytes() > max_block_size {
                 max_block_size_reached = true;
-                continue;
+                continue 'reap;
             }
 
             block_size += tx.size_bytes();
@@ -145,10 +145,10 @@ async fn run_build_proposal_task(
 
         if max_block_size_reached {
             trace!("Max block size reached, stopping tx generation");
-            break;
+            break 'reap;
         } else if start.elapsed() > build_duration {
             trace!("Time allowance exceeded, stopping tx generation");
-            break;
+            break 'reap;
         }
     }
 
--- a/code/crates/starknet/host/src/host/proposal.rs
+++ b/code/crates/starknet/host/src/host/proposal.rs
@@ -153,11 +153,19 @@ async fn run_build_proposal_task(
     }
 
     // Vote extensions
-    if !vote_extensions.extensions.is_empty() {
+    if !max_block_size_reached && !vote_extensions.extensions.is_empty() {
         let transactions = vote_extensions
             .extensions
             .into_iter()
             .map(|(_, e)| e.message)
+            .take_while(|e| {
+                let keep_going = block_size + e.len() <= max_block_size;
+                if keep_going {
+                    block_size += e.len();
+                    block_tx_count += 1;
+                }
+                keep_going
+            })
             .map(Transaction::new)
             .collect();
 
--- a/code/crates/starknet/host/src/host/proposal.rs
+++ b/code/crates/starknet/host/src/host/proposal.rs
@@ -12,7 +12,7 @@ use tokio::sync::{mpsc, oneshot};
 use tokio::time::Instant;
 use tracing::{error, trace};
 
-use malachitebft_core_types::Round;
+use malachitebft_core_types::{Round, VoteExtensions};
 
 use crate::host::starknet::StarknetParams;
 use crate::mempool::{MempoolMsg, MempoolRef};
@@ -23,6 +23,7 @@ pub async fn build_proposal_task(
     round: Round,
     proposer: Address,
     private_key: PrivateKey,
+    vote_extensions: VoteExtensions<MockContext>,
     params: StarknetParams,
     deadline: Instant,
     mempool: MempoolRef,
@@ -34,6 +35,7 @@ pub async fn build_proposal_task(
         round,
         proposer,
         private_key,
+        vote_extensions,
         params,
         deadline,
         mempool,
@@ -51,6 +53,7 @@ async fn run_build_proposal_task(
     round: Round,
     proposer: Address,
     private_key: PrivateKey,
+    vote_extensions: VoteExtensions<MockContext>,
     params: StarknetParams,
     deadline: Instant,
     mempool: MempoolRef,
@@ -147,6 +150,22 @@ async fn run_build_proposal_task(
             trace!("Time allowance exceeded, stopping tx generation");
             break;
         }
+    }
+
+    // Vote extensions
+    if !vote_extensions.extensions.is_empty() {
+        let transactions = vote_extensions
+            .extensions
+            .into_iter()
+            .map(|(_, e)| e.message)
+            .map(Transaction::new)
+            .collect();
+
+        let part = ProposalPart::Transactions(Transactions::new(transactions));
+
+        block_hasher.update(part.to_sign_bytes());
+        tx_part.send(part).await?;
+        sequence += 1;
     }
 
     // BlockProof
--- a/code/crates/app-channel/src/connector.rs
+++ b/code/crates/app-channel/src/connector.rs
@@ -98,6 +98,26 @@ where
                 reply_to.send(rx.await?)?;
             }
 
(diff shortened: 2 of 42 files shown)


ancazamfir:
Looks good so far. 
Previous implementation was not really a vote extension ala Comet. And we were not doing anything at all with the extensions. We built that in order to compare the case where all Tx-es are included in the block by the proposer and therefore disseminated by a single sender (the proposer) with the case where the Tx-es (for next block) are included in the votes, i.e. Tx dissemination by multiple senders (potentially all validators). For the latter case, mempool bcast was disabled. We measured and compared the latencies but those measurements did not include Tx execution or block store times (we didn't have it at the time).
Given this, in v2 we need to pass the extensions to the app (starknet) that will do something with them. Here are a few comments along those lines:

- In the starknet app we can have a usecase along the lines of the e2e app in Comet, see:
https://github.com/cometbft/cometbft/blob/63a2a645835fcba5ac7ff6926116d49952e49ca4/test/e2e/app/app.go#L608-L627

- [on_decided()](https://github.com/informalsystems/malachite/blob/f938db651df80177031287fc05f03e5877f1ffc9/code/crates/starknet/host/src/actor.rs#L548)  - Should take an extra argument, the extensions seen by this node. The vote signature should not include the extension (keep the extension signature separate), the certificate should only include/ aggregate the votes. 
The extensions should be stored but not part of the block.

- If we are proposer in the next block, in [on_get_value()](https://github.com/informalsystems/malachite/blob/f938db651df80177031287fc05f03e5877f1ffc9/code/crates/starknet/host/src/actor.rs#L277) we build a transaction with those extensions and include it after the other tx-es. Alternatively we pass the extensions to `on_get_value()`. In this case we have to keep the extensions in consensus (the driver) when moving to the next height.

ancazamfir:
regarding @cason [comment](https://github.com/informalsystems/malachite/issues/717#issuecomment-2572981285)
Maybe we can include that here as well. This way we can clean the `(Signed)Vote` protos and code and add a separate message `(Signed)VoteExtension`. We can diseminate over a different topic for now.

romac:
> regarding @cason [comment](https://github.com/informalsystems/malachite/issues/717#issuecomment-2572981285) Maybe we can include that here as well. This way we can clean the `(Signed)Vote` protos and code and add a separate message `(Signed)VoteExtension`. We can diseminate over a different topic for now.

I agree we should do that, but only in a follow-up PR.

ancazamfir:
> > regarding @cason [comment](https://github.com/informalsystems/malachite/issues/717#issuecomment-2572981285) Maybe we can include that here as well. This way we can clean the `(Signed)Vote` protos and code and add a separate message `(Signed)VoteExtension`. We can diseminate over a different topic for now.
> 
> I agree we should do that, but only in a follow-up PR.

That also works so let's do that if you think it's better. Was about to say the same but then I thought that a lot of vote structure and handling will need to be undone.

romac:
> `on_decided` [...] should take an extra argument, the extensions seen by this node

Aren't the extensions already included in the certificate passed to `on_decided`? ie. `certificate.aggregated_signature.signatures[].extension`

romac:
> Was about to say the same but then I thought that a lot of vote structure and handling will need to be undone.

Yes perhaps but since doing separate propagation of vote extensions might end up being complex I'd be more comfortable doing it in its own PR even if I have to undo a bunch of changes made here.

romac:
> * The extensions should be stored but not part of the block.

You mean not store in the db in the `decided_blocks` table but rather in a separate table, or just kept in memory?

ancazamfir:
> > * The extensions should be stored but not part of the block.
> 
> You mean not stored in the db in the `decided_blocks` table but rather in a separate table, or just kept in memory?

just in memory is fine, given how we use them in the mock app, i don't think there is a problem if we lose them on restart.

romac:
@ancazamfir Let me know if my latest commits have addressed your comments. If so then, I'll get to including the extensions in the value itself (maybe as a special tx for the Starknet app)

ancazamfir:
> @ancazamfir Let me know if my latest commits have addressed your comments. If so then, I'll get to including the extensions in the value itself (maybe as a special tx for the Starknet app)

yes, looks good!

code/crates/starknet/host/src/host/proposal.rs
```diff
@@ -149,6 +152,22 @@ async fn run_build_proposal_task(
...
+    // Vote extensions
+    if !vote_extensions.extensions.is_empty() {
+        let transactions = vote_extensions
+            .extensions
+            .into_iter()
+            .map(|(_, e)| e.message)
+            .map(Transaction::new)
+            .collect();
+
+        let part = ProposalPart::Transactions(Transactions::new(transactions));
```
ancazamfir, line 164: maybe make sure they're in multiple parts if we're over the part size.

cason:
Jumping a little late here, vote extensions are transitional data within a height. They should not be stored on chain, they should not, except under request, be exchanged by peers (via VoteSync). They don't have any use from when the next proposal is produced and/or received.

The idea of having two signatures, one for the vote extension, another for the vote message itself is not great, to be honest. In my opinion, extensions are data produced by the application, that might choose or not to sign it. The vote message could include the hash of the vote extension, a constant-size value, with a special value for no/empty extension. This saves us from signing votes twice _but_ is a problem if we want to consider vote/signatures aggregation, since votes for the same round step and value become different, since the extensions are non-deterministic data. This topic is worth additional discussion. The solution adopted by Comet is sub-optimal, in my view, it is like that for "historical reasons".

cason:
> [on_decided()](https://github.com/informalsystems/malachite/blob/f938db651df80177031287fc05f03e5877f1ffc9/code/crates/starknet/host/src/actor.rs#L548) - Should take an extra argument, the extensions seen by this node.

Notice that in this case, each node is seeing "only" strictly 2/3+ of the produced vote extensions. The solution works, in terms of performance improvements, mostly if we can see more than that, ideally all produced vote extensions.

Also, from the design concept, they are really only useful for the next height, either when producing a block, or when validating a block produced by another node.

romac:
> Notice that in this case, each node is seeing "only" strictly 2/3+ of the produced vote extensions

How about when running with non-zero commit timeout? Shouldn't this let us gather more precommits and therefore more vote extensions?

romac:
> Jumping a little late here, vote extensions are transitional data within a height. They should not be stored on chain, they should not, except under request, be exchanged by peers (via VoteSync). They don't have any use from when the next proposal is produced and/or received.

Yep that's the way they are now implemented as of this PR.

> The idea of having two signatures, one for the vote extension, another for the vote message itself is not great, to be honest. In my opinion, extensions are data produced by the application, that might choose or not to sign it. 

That's a valid point. So you are saying we should only work with the abstract `Extension` type and leave it up to the application to instantiate it with a signed or unsigned extension type?

> The vote message could include the hash of the vote extension, a constant-size value, with a special value for no/empty extension. 

This will only apply when we start disseminating vote extensions over a bespoke channel instead of stuffing them in the votes, right?

I was actually wondering about that, thinking that we need a way to know how to correlate an extension we receive "out-of-band" (ie. not via the consensus gossip topic but over some other topic/channel, eg. DOG) with a precommit. Including the hash of the vote extension in the vote seems like a good idea and would allow us to track whether we have received all extensions for votes we have seen and to discard extensions for votes we have not seen.
Does that makes sense?

code/crates/starknet/host/src/host/proposal.rs
romac: Good point! Moreover, what happens if `value_size + extensions_size > max_block_size`, do we just drop the remaining extensions or do we include them regardless? cc @cason

cason:
> How about when running with non-zero commit timeout? Shouldn't this let us gather more precommits and therefore more vote extensions?

Yes, but we should not wait for the commit timeout to decide, but to start the next height.

cason:
> > How about when running with non-zero commit timeout? Shouldn't this let us gather more precommits and therefore more vote extensions?
> 
> Yes, but we should not wait for the commit timeout to decide, but to start the next height.

And in anyway this is not a really nice design. There are other ways to collect additional precommits without halting the node for 1s.

code/crates/starknet/host/src/host/proposal.rs
ancazamfir: let's just drop. not much else we can do. Note this is just something for our test app so we can do whatever we want here. In a useful app things would be different but wanted something very simple.

cason:
> > The idea of having two signatures, one for the vote extension, another for the vote message itself is not great, to be honest. In my opinion, extensions are data produced by the application, that might choose or not to sign it.
> 
> That's a valid point. So you are saying we should only work with the abstract `Extension` type and leave it up to the application to instantiate it with a signed or unsigned extension type?
> 
> > The vote message could include the hash of the vote extension, a constant-size value, with a special value for no/empty extension.
> 
> This will only apply when we start disseminating vote extensions over a bespoke channel instead of stuffing them in the votes, right?
> 
> I was actually wondering about that, thinking that we need a way to know how to correlate an extension we receive "out-of-band" (ie. not via the consensus gossip topic but over some other topic/channel, eg. DOG) with a precommit. Including the hash of the vote extension in the vote seems like a good idea and would allow us to track whether we have received all extensions for votes we have seen and to discard extensions for votes we have not seen. Does that makes sense?

Yes, this is exactly the point. We can propagate extensions as we want, the most efficient method for it, but we should also be able to link an extension to a vote. This mean that when we receive a vote (not necessarily a precommit, could be also a Prevote) we need to know that there is an extension `X` we should expect from the same sender.

I really don't see how to "ensure" that without having something like `hash(X)` as part of the vote. Which has, on the other hand, the pitfall of rendering vote/signature aggregation more complex.

code/crates/starknet/host/src/host/proposal.rs
cason: Why do we broadcast the extensions in the proposal? It should be enough to include the hashes of them. Nodes are supposed to receive them independently.

ancazamfir:
> Which has, on the other hand, the pitfall of rendering vote/signature aggregation more complex.

I agree with this but not clear to me we need a hash, depends on the requirements. 

When an extension comes, since it would include (height, round, id), we can drop it if it's invalid (e.g. we moved to the next height or not coming from a val in the set). At "decided" time we can exclude the ones that don't match the commit certificate (height, round, id) or if the val has not voted (the extension may arrive before the vote)

Should it be mandatory for a validator to issue an extension (if enabled)? Could we just collect extensions on a best effort basis, meaning there can be less than 2/3rds?

I propose we write a spec for VE v3 and merge this PR for now.

cason:
> I propose we write a spec for VE v3 and merge this PR for now.

I agree.

code/crates/starknet/host/src/host/proposal.rs
romac: @ancazamfir I just realized we do not really have a `max_part_size`, but rather `max_txs_per_part`. So we can either:

- Use some heuristic to compute the `max_part_size` based on txs seen so far
- Add a `max_part_size` setting, remove `max_txs_per_part`
- Have one part per extension
- Leave it as is

What do you think?

romac: > Why do we broadcast the extensions in the proposal? It should be enough to include the hashes of them. Nodes are supposed to receive them independently.

Yes, I'll get to that in a separate PR. For now we are just piggybacking on proposal parts gossip to disseminate the extensions.

cason:
> Should it be mandatory for a validator to issue an extension (if enabled)? Could we just collect extensions on a best effort basis, meaning there can be less than 2/3rds?

We can do in a best-effort in a first effort. But recall that a Byzantine validator can produce multiple extensions for the same round step (height, round, step). And if they are not signed, it can even produce extensions on behalf of other validators...

Plus, if this is a best-effort dissemination of data, what is the actual difference from a mempool? The fact that we expect at most one message per sender?

ancazamfir:
> > Should it be mandatory for a validator to issue an extension (if enabled)? Could we just collect extensions on a best effort basis, meaning there can be less than 2/3rds?
> 
> We can do in a best-effort in a first effort. But recall that a Byzantine validator can produce multiple extensions for the same round step (height, round, step). 

We can just keep one if multiple are sent. 

> And if they are not signed, it can even produce extensions on behalf of other validators...
> 

They should be signed.

> Plus, if this is a best-effort dissemination of data, what is the actual difference from a mempool? The fact that we expect at most one message per sender?

Yes. And if an app wants to collect more it can set a bigger commit timeout. 

But let's write a spec and if possible not include anything in the vote. My concern is that our main customer is starknet and they haven't express any interest in vote extension so ideally we should use their proto definitions.

code/crates/starknet/host/src/host/proposal.rs
ancazamfir: > @ancazamfir I just realized we do not really have a `max_part_size`, but rather `max_txs_per_part`. So we can either:
> 
> * Use some heuristic to compute the `max_part_size` based on txs seen so far
> * Add a `max_part_size` setting, remove `max_txs_per_part`
> * Have one part per extension
> * Leave it as is
> 
> What do you think?

Do we have access to the `vote_extensions.size` config? If not we can just count one extension as one tx for now. And maybe in the app when we generate it we use min(`tx_size`, `vote_extensions.size`) since the extension is used as a tx in this app.

ancazamfir (approved):
looks great! 🚀

PR #777, opened 2025-01-16
feat(spec): Argue that Tendermint is a safe and live consensus algorithm

cason:
Part of: #761.

Rendered [section](https://github.com/informalsystems/malachite/blob/cason/761-correctness/specs/consensus/overview.md#correctness-1).

---

#### For all contributors

- [x] Reference GitHub issue


specs/consensus/overview.md
```diff
@@ -766,6 +766,246 @@ messages broadcast or received by correct processes at times `t < GST`.
...
+
+Thus, based on the assumption that less than 1/3 of total voting power is owned
+by Byzantine processes, and the locking rules, two correct processes cannot
+decide different values in a height of consensus.
+
+### Liveness
+
+This section argues that Tendermint is a live consensus algorithm.
+For a complete proof, please refer to the [Tendermint paper][tendermint-arxiv].
+
```
josef-widder, line 940:
```suggestion

Tendermint is a round-based algorithm. Liveness arguments for such algorithms typically have two ingredients
A. If there is a good round, then all processes decide before or in this round.
B. Eventually, there will be a good round.

Regarding Point A, a good round is one in which
- the proposer sends the same proposal `v` to everyone
- all correct processes receive the proposal before the timeouts expire
- all correct processes accept the proposal and send a `prevote` for `v`
- all correct processes receive the prevotes from all correct processes and send precommit for that value (before the timeouts expire)
- all correct processes receive the precommits from all correct processes (before the timeouts expire) 

In such a round, all correct processes decide in line 51.

The challenge that remains is to prove that eventually there is such a good round. There are two crucial points
- **Synchrony.** Messages should be delivered and timeouts should not expire 
- **Value handling.** Correct processes need to accept `v`. The complication comes from the following features: 
    - for accepting, processes check `v` against `lockedValue`,  
    - the proposer might use `validValue` to choose `v`, 
    - the evaluations of `lockedValue` and `validValue` change in a rather complicated manner at different processes in arbitrary asynchronous prefixes of the computation.

Let's go through these points one-by-one.

```

josef-widder (approved):
Especially the discussions about lockedValue and validValue are super useful. I have suggested some additional text in the hope it improves readability.

PR #786, opened 2025-01-20
feat(code): Make `peer` crate `no_std` compatible

OakenKnight:
Addressed FIXME on [code/crates/peer/src/lib.rs:21](https://github.com/informalsystems/malachite/blob/7e121597d8be94b8a720a4c8fcd4bed832e4d0ab/code/crates/peer/src/lib.rs#L21)

---

#### For all contributors

- [ ] Reference GitHub issue

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -111,7 +111,7 @@ jobs:
         with: 
           cache-workspaces: "code"
           target: thumbv7m-none-eabi
-      - name: Check no_std compatability
+      - name: Check no_std compatibility
         run: cargo build --target thumbv7m-none-eabi -p informalsystems-malachitebft-peer
 
   clippy:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -104,6 +104,8 @@ jobs:
         uses: actions/checkout@v4
       - name: Install Protoc
         uses: arduino/setup-protoc@v3
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/rust.yml
```diff
@@ -93,6 +93,27 @@ jobs:
...
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Setup Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with: 
+          cache-workspaces: "code"
+          target: thumbv7m-none-eabi
+      - name: Check no_std compatability
```
romac, line 114:
```suggestion
      - name: Check no_std compatibility
```

romac (approved):

cason:
Just a random question: how bad is it?

romac:
@cason I am not sure what you are referring to?

cason:
Probably I don't know Rust enough. I've read here that there is no actual impact to users, it should only make binaries smaller?

I mean, it is not that we are not compatible or are "breaking" some stuff, are we?

romac:
The reason is that we want eventually want to make the `core-consensus` crate `no_std` compatible so that it can run in a variety of platforms, e.g. embedded devices, browsers via WASM, etc.

So yes, you are correct, not being that `no_std` compatible does not have any direct impact on typical users.

Issue #773, opened 2025-01-16
code: Check proposal validity in example app


romac:
In the example app, when we assemble a proposal from its parts, we currently do not check that the signature embedded in the final part (`Fin`) matches what we expect:

https://github.com/informalsystems/malachite/blob/77c741e19f24e56e0adb0efd150e63087813cb99/code/examples/channel/src/state.rs#L320

The original signature is computed here:

https://github.com/informalsystems/malachite/blob/77c741e19f24e56e0adb0efd150e63087813cb99/code/examples/channel/src/state.rs#L292-L298

PR #793, opened 2025-01-21
feat(code): Check proposal validity in the example app

nenadmilosevic95:
This PR aims to close: #773. 

---

#### For all contributors

- [x] Reference GitHub issue

### DIFF
--- a/code/examples/channel/src/state.rs
+++ b/code/examples/channel/src/state.rs
@@ -42,6 +42,19 @@ pub struct State {
     pub peers: HashSet<PeerId>,
 }
 
+/// Represents errors that can occur during the verification of a proposal's signature.
+#[derive(Debug)]
+enum SignatureVerificationError {
+    /// Indicates that the `Fin` part of the proposal is missing.
+    MissingFinPart,
+
+    /// Indicates that the proposer was not found in the validator set.
+    ProposerNotFound,
+
+    /// Indicates that the signature in the `Fin` part is invalid.
+    InvalidSignature,
+}
+
 // Make up a seed for the rng based on our address in
 // order for each node to likely propose different values at
 // each round.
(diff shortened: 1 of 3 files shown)


nenadmilosevic95:
The code is not complete yet and does not compile because we need to implement the mapping from the PeerID to the Public Key that we should use to verify the signature.

code/examples/channel/src/state.rs
```diff
@@ -81,6 +82,7 @@ impl State {
     pub async fn received_proposal_part(
         &mut self,
         from: PeerId,
+        from_pub_key: &PublicKey,
```
romac, line 85: Let's remove this and instead pass the validator set (accessible in `app.rs` via `genesis.initial_validator_set`

```suggestion
        validator_set: &ValidatorSet,
```

code/examples/channel/src/state.rs
```diff
@@ -90,6 +92,45 @@ impl State {
...
+
+            let hash = hasher.finalize().to_vec();
+            let signature =
+                signature.expect("Fin part is guaranteed to exist due to checks on lines 90-92");
+
+            // Verify the signature
+            if !self
+                .ctx
+                .signing_provider
+                .verify(&hash, signature, from_pub_key)
```
romac, line 124: Here you can get the public key of the proposer by doing something like:

```rust
let public_key = validator_set
    .validators
    .iter()
    .find(|v| v.address == parts.proposer)
    .map(|v| v.public_key);

let Some(public_key) = public_key else {
    error!(proposer = %parts.proposer, "Proposer not found in validator set");
    return Ok(None);
};
```

nenadmilosevic95: This is great, thanks! This way, there's no need for mapping PeerID, as we only need to verify the proposer's signature for now.

code/examples/channel/src/state.rs
nenadmilosevic95: Perfect. Is there a particular reason why the genesis is not included as part of the state?

nenadmilosevic95: I will do that also then! Thanks!

code/examples/channel/src/state.rs
```diff
@@ -114,19 +121,31 @@ impl State {
...
+                error!(proposer = %parts.proposer, "Fin proposal part is missing");
+                return Ok(None);
+            };
+
+            // Get proposal's public key
+            let public_key = self
+                .get_validator_set()
+                .validators
+                .iter()
+                .find(|v| v.address == parts.proposer)
```
romac, line 134: I just realized we can shorten this by doing:

```suggestion
                .get_by_address(&parts.proposer)
```

romac (approved):

PR #796, opened 2025-01-21
feat(code/app/starknet): Include existing transactions when reaping the mempool not just the ones generated on the fly

OakenKnight:
Addresses FIXME on [/code/crates/starknet/host/src/mempool.rs](https://github.com/informalsystems/malachite/blob/1947b34fafbe372e66a45b06c05d13a5f64f6f3a/code/crates/starknet/host/src/mempool.rs#L220)

---

#### For all contributors

- [ ] Reference GitHub issue

### DIFF
--- a/code/crates/starknet/host/src/mempool.rs
+++ b/code/crates/starknet/host/src/mempool.rs
@@ -217,6 +217,12 @@ impl Actor for Mempool {
             }
 
             Msg::Update { .. } => {
+                // Clear all transactions from the mempool, given that we consume
+                // the full mempool when proposing a block.
+                //
+                // This reduces the mempool protocol overhead and allow us to
+                // observe issues strictly related to consensus.
+                // It also bumps performance, as we reduce the mempool's background traffic.
                 state.transactions.clear();
             }
         }
--- a/code/crates/starknet/host/src/mempool.rs
+++ b/code/crates/starknet/host/src/mempool.rs
@@ -216,8 +216,11 @@ impl Actor for Mempool {
                 reply.send(txes)?;
             }
 
-            Msg::Update { tx_hashes } => {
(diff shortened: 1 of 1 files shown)


romac:
@ancazamfir Do you remember if there is a reason we were not doing this already?

cason:
A reason is that we always consume the full mempool when proposing a block.

We use this hack in Comet's e2e to reduce the mempool protocol overhead and observe issues strictly related to consensus. This also bumps performance, as we reduce the mempool's background traffic.

romac:
Thanks for the insights @cason, then let's leave it as is. 

@OakenKnight Can you please revert the change in this PR and remove the TODO comment instead?

ancazamfir:
sorry i missed this earlier. when we "reap" N transactions for a block part, we generate and disseminate them (if dissemination is enabled) and add them to the mempool at this point. When we decide on a block we clear the mempool which might be empty or have a subset.

cason:
>  When we decide on a block we clear the mempool which might be empty or have a subset.

But we are potentially removing transactions not included in the decided block, aren't we?

romac:
> But we are potentially removing transactions not included in the decided block, aren't we?

Yes, if there were more of those than the `max_tx_count`. Not a great design I'll admit, but this is not a real mempool and we wanted to keep things simple and easy to configure for testing.

ancazamfir:
> > When we decide on a block we clear the mempool which might be empty or have a subset.
> 
> But we are potentially removing transactions not included in the decided block, aren't we?

yes, should have added "or superset". Our mempool is just dumb, we put tx-es in mempool just to bcast something in the mempool network so we have some traffic there if we want.

romac (approved):

PR #797, opened 2025-01-21
chore: Remove unused PrecommitValue from Input

hvanz:
`PrecommitValue` is not used anywhere in the code. The comment references line 51, which is actually part of line 49, covered already by input `ProposalAndPrecommitValue`.


romac (approved):

ancazamfir (approved):
Good catch!

PR #804, opened 2025-01-24
docs(code): Document the three consensus modes and how to pick a `Value` type accordingly

romac:


code/crates/core-types/src/value.rs
```diff
@@ -54,24 +54,69 @@ impl<Value> NilOrVal<Value> {
     }
 }
 
-/// Defines the requirements for the type of value to decide on.
+/// The `Value` type denotes the value `v` carried by the `Proposal`
```
cason, line 57: Proposal is a message from the implementation?

If referring to the paper, I would use `PROPOSAL` instead.

code/crates/core-types/src/value.rs
```diff
@@ -54,24 +54,69 @@ impl<Value> NilOrVal<Value> {
...
+/// consensus message that is gossiped to other nodes by the proposer.
```
cason, line 58:
```suggestion
/// consensus message broadcast by the proposer of a round of consensus.
```

code/crates/core-types/src/value.rs
```diff
@@ -54,24 +54,69 @@ impl<Value> NilOrVal<Value> {
...
+    type Id: Clone + Debug + Display + Eq + Ord + Send + Sync;
 
     /// The ID of the value.
     fn id(&self) -> Self::Id;
 }
 
+/// The possible messages used to deliver proposals
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ValuePayload {
+    /// The proposer publishes a `Proposal` message carrying the full value `v`, and does not stream any proposal parts at all.
```
cason, line 78:
```suggestion
    /// The proposer broadcasts a `Proposal` consensus message carrying the full proposed value `v`. There is no proposal part streaming.
```

code/crates/core-types/src/value.rs
```diff
@@ -54,24 +54,69 @@ impl<Value> NilOrVal<Value> {
...
+    /// Better suited for small blocks to avoid overhead of gossiping parts.
```
cason, line 79:
```suggestion
    /// Better suited for small proposed values when there are no benefits of gossiping proposal parts.
```

code/crates/core-types/src/value.rs
```diff
@@ -54,24 +54,69 @@ impl<Value> NilOrVal<Value> {
...
+    /// In this case `Value` is typically set to be the block and `Id` is its hash.
+    ProposalOnly,
+
+    /// The proposer does not publish a `Proposal` message at all, it only streams the proposed value as proposal parts.
```
cason, line 83:
```suggestion
    /// The proposer does not broadcast a `Proposal` consensus message at all. The proposer only streams the proposed value as proposal parts.
```

code/crates/core-types/src/value.rs
```diff
@@ -54,24 +54,69 @@ impl<Value> NilOrVal<Value> {
...
+    /// In this case `Value` is typically set to the same type as `Id`.
+    PartsOnly,
+
+    /// The proposer publishes a `Proposal` message carrying only `id(v)`, and streams the full value as proposal parts.
```
cason, line 87:
```suggestion
    /// The proposer broadcasts a `Proposal` message carrying `id(v)` and streams the full proposed value `v` as proposal parts.
```

cason (commented):
Some suggestions to render code comments/documentation more clear.

code/crates/core-types/src/value.rs
romac: Yes it refers to the abstract `Proposal` type/trait. Will add a link to the trait.

Issue #520, opened 2024-11-05
code: Add PoC application where `Value` is full block


ancazamfir:
Follow up on value payload messages introduced for Implicit Proposal. Implement the `proposal-only` case:
- [ ] Write a PoC app that provides a one shot value for `getValue` to be included in the (explicit) Proposal message, no parts. To clarify, this is not a `ValueId` as done today (concretely `BlockHash`, even if presented to Consensus as `Value`)
   - on the Proposal receiver side the `value` should be punted to the app (single BlockPart) and `validity` returned in `ReceivedProposedValue`. In the code, this will be naturally muxed with the actual proposal.
- [ ] Test how big (wrt value size) we can go with the Proposal payload

PR #812, opened 2025-01-28
chore(test): Add blocks proto

greg-szabo:
Closes: #520

This PR expands the test crate with the Block proto struct so example apps and tests can handle block transfers over the wire.

---

#### For all contributors

- [ ] Reference GitHub issue

### DIFF
--- a/code/crates/test/proto/transaction.proto
+++ b/code/crates/test/proto/transaction.proto
@@ -1,0 +1,40 @@
+syntax = "proto3";
+
+import "consensus.proto";
+
+package test;
+
+// This is a transaction that is already accepted in a block.
+message Transaction
+{
+    message Dummy {
+        bytes bytes = 1;
+    }
+
+    message DeclareV0 {
+        Address sender = 1;
+        Signature signature = 3;
+        Hash class_hash = 4;
+    }
+
+    oneof txn {
+        Dummy dummy = 42;
+
+        DeclareV0 declare_v0 = 1;
+    }
+    Hash transaction_hash = 41;
+}
+
+message Transactions {
+    repeated Transaction transactions = 1;
+}
+
+message Hash {
+    bytes elements = 1;
+}
+
+message Block {
+    uint64 block_number = 1;
+    Transactions transactions = 2;
+    Hash block_hash = 3;
+}
--- a/code/crates/test/src/block/hash.rs
+++ b/code/crates/test/src/block/hash.rs
@@ -1,0 +1,92 @@
+use core::{fmt, str};
+
+use bytes::Bytes;
+use serde::{Deserialize, Serialize};
+
+use malachitebft_proto;
+
+pub type MessageHash = Hash;
+pub type BlockHash = Hash;
+
+impl malachitebft_core_types::Value for BlockHash {
+    type Id = BlockHash;
+
+    fn id(&self) -> Self::Id {
+        *self
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct Hash([u8; 32]);
+
+impl Hash {
+    pub const fn new(bytes: [u8; 32]) -> Self {
+        Self(bytes)
+    }
+
+    pub fn as_bytes(&self) -> &[u8; 32] {
+        &self.0
+    }
+
+    #[allow(clippy::len_without_is_empty)]
+    pub fn len(&self) -> usize {
+        self.as_bytes().len()
+    }
+}
+
+impl PartialOrd for Hash {
+    #[cfg_attr(coverage_nightly, coverage(off))]
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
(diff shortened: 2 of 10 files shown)


code/crates/test/proto/transaction.proto
```diff
@@ -0,0 +1,40 @@
...
+{
+    message Dummy {
+        bytes bytes = 1;
+    }
+
+    message DeclareV0 {
+        Address sender = 1;
+        Signature signature = 3;
+        Hash class_hash = 4;
+    }
```
romac, line 18: We already have a dummy tx type so we can remove this I think, it's a starknet-specific tx type.

code/crates/test/src/block/transaction.rs
```diff
@@ -0,0 +1,174 @@
...
+
+use bytes::Bytes;
+use malachitebft_proto::{self as proto};
+
+use crate::hash::Hash;
+
+/// Transaction
+#[derive(Clone, PartialEq, Eq, Ord, PartialOrd)]
+pub struct Transaction {
+    data: Bytes,
```
romac, line 11: Let's have the data be an enum with the Dummy tx as a variant, so that we can add more interesting tx types later down the road.

code/crates/test/src/block/transaction.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /// Hash of this transaction
+    pub fn hash(&self) -> Hash {
+        self.hash
+    }
+
+    /// Compute the hash of a transaction
+    ///
+    /// TODO: Use hash function from Context
```
romac, line 45: We can remove this todo, it's actually not needed since hashing only happens on the app side and the Context does not have to care about it.

code/crates/test/src/block/types.rs
```diff
@@ -0,0 +1,53 @@
+use crate::{hash::BlockHash, transaction::Transactions, Height};
```
romac, line 1: Let's re-order these imports

code/crates/test/src/block/hash.rs
```diff
@@ -0,0 +1,92 @@
+use core::{fmt, str};
+
+use bytes::Bytes;
+use serde::{Deserialize, Serialize};
+
+use malachitebft_proto;
+
+pub type MessageHash = Hash;
```
romac, line 8: This is not needed I think.

code/examples/channel/src/state.rs
```diff
@@ -5,25 +5,28 @@ use std::collections::HashSet;
 
 use bytes::Bytes;
 use eyre::eyre;
+use rand::distributions::Alphanumeric;
 use rand::rngs::StdRng;
 use rand::{Rng, SeedableRng};
 use sha3::Digest;
 use tracing::{debug, error};
 
+use crate::store::{DecidedValue, Store};
```
romac, line 14: Let's re-order these imports

code/examples/channel/src/state.rs
```diff
@@ -223,7 +226,11 @@ impl State {
         assert_eq!(round, self.current_round);
 
         // We create a new value.
-        let value = self.make_value();
+        let block = self.fetch_block(height);
+        // Simplified value creation. In a real application, use the whole hash.
+        let mut block_hash_short = [0; 8];
+        block_hash_short.copy_from_slice(&block.block_hash.as_bytes()[0..8]);
+        let value = Value::new(u64::from_be_bytes(block_hash_short));
```
romac, line 233: Since this code appears twice, let's perhaps pull it out in a free-standing function?

code/examples/channel/src/app.rs
```diff
@@ -176,7 +176,12 @@ pub async fn run(state: &mut State, channels: &mut Channels<TestContext>) -> eyr
             } => {
                 info!(%height, %round, "Processing synced value");
 
-                let value = decode_value(value_bytes);
+                let block = decode_block(value_bytes);
+
+                // Simplified value creation. In a real application, use the whole hash.
+                let mut block_hash_short = [0; 8];
+                block_hash_short.copy_from_slice(&block.block_hash.as_bytes()[0..8]);
+                let value = Value::new(u64::from_be_bytes(block_hash_short));
```
romac, line 184: Let's add a TODO to change `Value` to hold a byte array in a follow-up PR.

romac (changes requested):

code/examples/channel/src/state.rs
```diff
@@ -223,7 +226,11 @@ impl State {
         assert_eq!(round, self.current_round);
 
         // We create a new value.
-        let value = self.make_value();
+        let block = self.fetch_block(height);
```
ancazamfir, line 229: since the block doesn't exist:
```suggestion
        let block = self.make_block(height);
```
also the block should be saved in some UNDECIDED_BLOCK table. And we should stream here the block and not the proposal (as it's done now [here](https://github.com/informalsystems/malachite/blob/fd54d2a9f028fbec645b8b432c8c840f991e9f14/code/examples/channel/src/app.rs#L90)). The `ProposalPart` in context is also wrong now. And reusing the test protos (`ProposalData` should be something like `Transactions transactions ..` or bytes) doesn't work.

Issue #763, opened 2025-01-14
test: MBT for Part Streaming


hvanz:
[Part Streaming](https://github.com/informalsystems/malachite/blob/c2aa59d69aa00fff9f1fe270cc278b4e65ec5f42/specs/starknet/block-streaming/README.md) is the protocol utilised for block propagation in the Starknet app. When sending a block proposal to a set of peers, a node partitions the block into multiple parts and send each part to the peers in `StreamMessage`s, which may arrive out of order. Then the peers must re-assemble the received parts into a block.

This is the [implementation](https://github.com/informalsystems/malachite/blob/c2aa59d69aa00fff9f1fe270cc278b4e65ec5f42/code/crates/starknet/host/src/streaming.rs). And this is the [Quint spec](https://github.com/informalsystems/malachite/blob/c2aa59d69aa00fff9f1fe270cc278b4e65ec5f42/specs/starknet/block-streaming/part_stream.qnt).

The goal is to write an MBT driver and tests for Part Streaming.

Tasks:
- [ ] Update and expand the documentation of the code and the spec, if needed.
- [ ] Extend the spec with new tests and invariants, if needed.
- [ ] Write in Rust the types and data structures that will be used to map the spec's state variables ([example](https://github.com/informalsystems/malachite/blob/8ae80b1a08ae751e71ac231f6574dbe7003988b9/code/crates/test/mbt/src/consensus.rs) of State struct for consensus-core module)
- [ ] Write implementation of `ItfRunner` for ITF traces generated from the spec (for example, [ConsensusRunner](https://github.com/informalsystems/malachite/blob/ad2e5df4194b968ac1b9897885aa9dcb72d230b4/code/crates/test/mbt/src/tests/consensus/runner.rs))
- [ ] Write MBT test, which generates ITF traces from the Quint spec and runs the traces as Rust tests ([example](https://github.com/informalsystems/malachite/blob/ad2e5df4194b968ac1b9897885aa9dcb72d230b4/code/crates/test/mbt/src/tests/consensus.rs) for consensus-core)  
- [ ] Integrate the new MBT tests in CI

PR #814, opened 2025-01-28
feat(spec/test): Model-based tests for part streaming

stojanovic00:
* Document proposal part streaming quint specification

* Document proposal part streaming implementation

* Remove redundant state variables (=> update invariants)

* Add concrete tests in quint spec
  * Fix symbolinc link from mbt/test to specs

* Generate and parse itf-trace

Closes: #763


stojanovic00:
* Update and expand the documentation of the code and the spec

* Extend the spec with new tests and invariants
    * Fix runtime bugs and remove redundant invariants

* Write in Rust the types and data structures that will be used to map the spec's state variables

* Write MBT test, which generates ITF traces from the Quint spec and runs the traces as Rust tests
    * Update/fix symbolic link from test/mbt to specs

* Write implementation of ItfRunner for ITF traces generated from the spec
    * Fix dependency issues with used crates

* Integrate the new MBT tests in CI

Closes: #763

code/crates/engine/src/util/streaming.rs
```diff
@@ -39,6 +39,7 @@ pub enum StreamContent<T> {
     Data(T),
 
     /// Fin must be set to true.
+    /// Q: Why not just leave out the bool then?
```
romac, line 42: This was done to match the original Starknet protos, but those have since removed the boolean, meaning we can do so as well.

code/crates/starknet/host/src/streaming.rs
```diff
@@ -8,63 +10,69 @@ use malachitebft_core_types::Round;
...
 use crate::types::{Address, Height, ProposalInit, ProposalPart};
+use std::fmt::Debug;
 
-struct MinSeq<T>(StreamMessage<T>);
+// Wraps StreamMessage to implement custom ordering for a BinaryHeap
+// The default BinaryHeap is a max-heap, so we reverse the ordering
+// by implementing Ord in reverse to make it min-heap, which suits the purpose of efficiently
+// providing available proposal part with smallest sequence number
+#[derive_where(Debug)]
+pub struct MinSeq<T: Debug>(pub StreamMessage<T>);
```
romac, line 20: No need to include an explicit bound on `T`, instead you can do:

```suggestion
#[derive(Debug)]
pub struct MinSeq<T>(pub StreamMessage<T>);
```

and remove all the `Debug` bounds below.

code/crates/starknet/host/src/streaming.rs
```diff
@@ -8,63 +10,69 @@ use malachitebft_core_types::Round;
...
-struct StreamState<T> {
-    buffer: MinHeap<T>,
-    init_info: Option<ProposalInit>,
-    seen_sequences: HashSet<Sequence>,
-    next_sequence: Sequence,
-    total_messages: usize,
-    fin_received: bool,
-    emitted_messages: usize,
+#[derive_where(Default, Debug)]
+pub struct StreamState<T: Debug> {
```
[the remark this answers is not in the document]
romac: The difference between `derive` and `derive_where` is that the former will add a conditional bound on type parameters, but the latter will not, ie. `#[derive(Debug)]` generates

```rust
impl<T: Debug> Debug for StreamState<T> { ... }
```

and `#[derive_where(Default)]` generates

```rust
impl<T> Default for StreamState<T> { ... }
```

code/crates/test/mempool/proto/google/protobuf/any.proto
```diff
@@ -0,0 +1,162 @@
+// Protocol Buffers - Google's data interchange format
```
romac, line 1: Where does this file come from? Can we remove it?

romac (commented):
Left some minor comments, looks great otherwise! 🚀 

Will do a deeper pass later today or tomorrow.

code/crates/starknet/host/src/streaming.rs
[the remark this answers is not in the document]
stojanovic00: Thanks. I didn't know exact difference between those two. Now it's much clearer.

code/crates/test/mbt/src/tests/streaming/runner.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    // There is no result in the model, so it is empty
+    type Result = ();
+
+    type ExpectedState = SpecificationState;
+
+    type Error = ();
+
+    fn init(&mut self, expected: &Self::ExpectedState) -> Result<Self::ActualState, Self::Error> {
+        println!("🔵 init: expected state={:?}", expected.state);
```
hvanz, line 34: `expected.state` should be used to initialize the actual state.

code/crates/test/mbt/src/tests/streaming/runner.rs
```diff
@@ -0,0 +1,196 @@
...
+                    "unexpected fin received value"
+                );
+
+                assert_eq!(
+                    stream_state.emitted_messages,
+                    expected.state.emitted.len(),
+                    "unexpected emitted messages value"
+                );
+
+                Ok(true)
```
hvanz, line 181: The invariant `inv` in the spec states that 
```
state.finReceived implies (state.totalMessages > 0 and state.nextSequence <= state.totalMessages),
state.received.subseteq(messages),
state.emitted.toSet().subseteq(messages),
```
when `isDone` is false.
Probably it would be better to check these properties here in the MBT runner.

hvanz:
@romac Now I realize that the Starknet related specs were moved to its own directory. Should these new MBT tests be also moved to its own directory, so it will be easier to move then later to a separate repo?

code/crates/starknet/host/src/streaming.rs
```diff
@@ -49,7 +49,7 @@ impl<T> Default for MinHeap<T> {
 }
 
 impl<T> MinHeap<T> {
-    fn push(&mut self, msg: StreamMessage<T>) {
+    pub fn push(&mut self, msg: StreamMessage<T>) {
```
stojanovic00, line 52: I don’t like that I had to set a function to public just for testing purposes. Is this considered bad practice? If so, what are the ways to avoid it?

stojanovic00, line 52: I don’t like that I had to set a function to public just for testing purposes. Is this considered bad practice? If so, what are the ways to avoid it?

romac: Agreed, but unfortunately integration tests don't have privileged access to private methods. We could mark the method with `#[doc(hidden)]` to hide it from the docs, or add a parameter of a private type, but the latter is a bit overkill imho. We can leave it as-is, it's fine I think.

romac:
Yes, let's move that spec under the appropriate folder under `specs/`

code/crates/test/mbt/src/tests/streaming/runner.rs
stojanovic00: I updated state invariant check for both cases (isDone true and false). Checks that I couldn't realize are the ones that are using `messages` and `orderedMessages` vals because they are not accessible to MBT runner, they are accessible only in Quint invariant checks.

specs/starknet/block-streaming/part_stream.qnt
```diff
@@ -20,16 +28,17 @@ module part_stream {
...
         nextSequence: Sequence,
         totalMessages: int,
+        finReceived: bool,
         emitted: List[Message],
-        received: Set[Message]
     }
 
     var state: State
+    // Required for MBT
+    var incomingMessage: Option[Message]
```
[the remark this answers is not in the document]
stojanovic00: I did some research on how this should be implemented. The `quint run --mbt ...` command does indeed produce `action_taken` and `nondet_picks` metadata, which could be used to obtain `incomingMessage` in the MBT runner without introducing an `incomingMessage` state variable.

However, based on my understanding of the `quint run` command from the specification, it does not execute specific tests defined in the specification. Instead, it generates random traces to try to break defined invariants. If it finds a sequence of steps that break an invariant, it outputs that trace in ITF format. If no invariant is violated, it outputs the longest successful trace produced. There is also an option to output all random traces it produced.

While this is generally useful for testing specification, it does not produce ITF output for specific tests defined in the specification. Because of this, I don't think it can be used for MBT testing in its current form. It could, however, be useful for the random trace tests I recently implemented, but not for the concrete tests defined in the Quint spec.

I also checked whether `quint test` supports the `--mbt` flag, but it does not.

code/crates/test/mbt/src/tests/streaming/runner.rs
hvanz: I see. Both `messages` and `orderedMessages` could be defined as part of the runner's state. That would be easy because they are constants.

specs/starknet/block-streaming/part_stream.qnt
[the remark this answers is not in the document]
hvanz: Oh you're right. The `--mbt` flag is currently implemented for `quint run`, not for `quint test`. We'll have to stick to `incomingMessage` until this feature is fully implemented.

code/crates/starknet/test/mbt/src/tests/streaming.rs
```diff
@@ -61,6 +62,7 @@ fn test_mbt_part_streaming_random_traces() {
     let quint_seed = quint_seed();
 
     print!("{}\n", temp_path.to_string_lossy());
+    // WARNING: After moving starknet related specs to separate repo this probably won't  work
     generate_random_traces(
         "starknet/block-streaming/part_stream.qnt",
         &temp_path.to_string_lossy(),
```
stojanovic00, line 68: These two calls cause integration tests to fail. Interesting part is that they are passing when they are being run using `cargo test` and `nextest`, but when they are run with maelstorm using this command in pipeline (`rust.yml`):
`
cargo maelstrom \
            --slots 16 \
            --include 'package.match(informalsystems-malachitebft-starknet-test) || package.match(informalsystems-malachitebft-discovery-test)'
`
they fail. I suppose it has something to do with how maelstorm handles relative paths but I don't know how to fix this issue. 

Also, I noticed that inside `coverage.yml` mbt tests are also being run:
`
      - name: Generate code coverage
        run: cargo llvm-cov nextest -p informalsystems-malachitebft-test-mbt --all-features --lcov --output-path lcov-mbt.info
`

Should I also add there running of starknet mbt tests?

stojanovic00: Since this comment I tweaked workflow files more and separated starknet test (trace generation) utils with hope it will resolve issue with relative paths but that didn't help either.

hvanz (approved):
I see that now all CI tests are passing and the MBT tests was moved under the starknet crate. Thanks!

.github/workflows/mbt.yml
```diff
@@ -54,11 +54,17 @@ jobs:
...
-        working-directory: code/crates/test/mbt
-        run: cargo nextest run -p informalsystems-malachitebft-test-mbt --all-features --no-run
       - name: Current time as random seed for Quint
         run: echo "QUINT_SEED=$(date +%s)" >> $GITHUB_ENV
+      - name: Build code
+        working-directory: code/crates/test/mbt
+        run: |
+          cargo nextest run \
+           -p informalsystems-malachitebft-test-mbt \
+           -p informalsystems-malachitebft-starknet-test-mbt --all-features --no-run
```
romac, line 64:
```suggestion
           -p informalsystems-malachitebft-starknet-test-mbt \
           --all-features --no-run
```

.github/workflows/mbt.yml
```diff
@@ -54,11 +54,17 @@ jobs:
...
       - name: Run tests
         working-directory: code/crates/test/mbt
-        run: cargo nextest run -p informalsystems-malachitebft-test-mbt --all-features
+        run: |
+          cargo nextest run \
+           -p informalsystems-malachitebft-test-mbt \
+           -p informalsystems-malachitebft-starknet-test-mbt --all-features
```
romac, line 70:
```suggestion
           -p informalsystems-malachitebft-starknet-test-mbt \
           --all-features
```

code/Cargo.toml
```diff
@@ -31,6 +31,7 @@ members = [
 
   # Starknet
   "crates/starknet/*",
+  "crates/starknet/test/mbt", # wildcard from line above didn't work
```
romac, line 34:
```suggestion
  "crates/starknet/test/mbt",
```

code/crates/starknet/host/src/streaming.rs
```diff
@@ -1,3 +1,5 @@
+/// The most of algorithm logic is explained in '/malachite/code/crates/starknet/host/src/actor.rs'
+/// Above 'on_received_proposal_part'
```
romac, line 2: Let's do it the other way, ie. add the explanation here and link to it from `actor.rs`

code/crates/starknet/host/src/streaming.rs
```diff
@@ -8,8 +10,14 @@ use malachitebft_core_types::Round;
 use malachitebft_engine::util::streaming::{Sequence, StreamId, StreamMessage};
 
 use crate::types::{Address, Height, ProposalInit, ProposalPart};
+use std::fmt::Debug;
```
romac, line 13: Let's move this import with the other std imports at the top

code/crates/starknet/host/src/streaming.rs
```diff
@@ -8,8 +10,14 @@ use malachitebft_core_types::Round;
...
 
-struct MinSeq<T>(StreamMessage<T>);
+// Wraps StreamMessage to implement custom ordering for a BinaryHeap
+// The default BinaryHeap is a max-heap, so we reverse the ordering
+// by implementing Ord in reverse to make it min-heap, which suits the purpose of efficiently
+// providing available proposal part with smallest sequence number
```
romac, line 18:
```suggestion
/// Wraps a [`StreamMessage`] to implement custom ordering for a [`BinaryHeap`].
///
/// The default `BinaryHeap` is a max-heap, so we reverse the ordering
/// by implementing `Ord` in reverse to make it a min-heap, which suits the purpose of efficiently
/// providing available proposal part with smallest sequence number.
```

code/crates/starknet/host/src/streaming.rs
romac: Actually, let's just copy-paste the explanation here as well.

stojanovic00:
I appreciate the thorough reviews @hvanz  and @romac ! I hope I have addressed all the comments.

Issue #801, opened 2025-01-23
code: Improve mempool of Starknet test app


romac:
Our current mempool for the Starknet mock app is quite barebones and could use a few improvements, namely:

- [ ] Split out tx generation out of the Mempool actor and into its own actor (`MempoolLoad` or something)
- [ ] After a block is decided, only remove txs that were included in that block
- [ ] Allow configuring the `MempoolLoad` actor to generate load on the mempool in various ways
  - No load
  - Uniform load configured in TPS (transactions per second)
  - Non-uniform load to simulate peaks in tx ingestion, the frequency and amplitude of these peaks should be configurable
- [ ] Allow size of generated transactions to be configured (constant, uniformly distributed, other distributions)

cason:
Another way to produce load is to emulate what we call a "closed-loop client".

Namely, we tell the number of clients we want to emulate. A client submits a tx, wait until it is decided, then submits another txs, possibly after a (configurable, random, small?) sleep.

In other words, the number of emulated clients is the maximum number of outstanding requests in the system at any time.

PR #821, opened 2025-01-30
feat(code/starknet): Add mempool load generator to the Starknet test app

oakenknight:
Closes: #801

---

#### For all contributors

- [ ] Reference GitHub issue


code/crates/starknet/host/src/mempool_load.rs
```diff
@@ -0,0 +1,163 @@
...
+            MempoolLoadType::NonUniformLoad => {
+                debug!("entered nonuniform load branch");
+
+                let mut rng = rand::thread_rng();
+                let interval = Duration::from_secs(rng.gen_range(1..10));
+                let count = rng.gen_range(500..=10000) as usize;
+                let size = rng.gen_range(128..=512) as usize;
+                tokio::spawn(ticker(interval, myself.clone(), move || {
+                    Msg::GenerateTransactions { count, size }
+                }))
```
romac, line 120: This is still uniform load but with random parameters. A non-uniform load would have spikes of high load relative to the normal. 

Maybe we could do something like this (in pseudo-Rust):

```rust
MempoolLoadType::NonUniformLoad {
    /// Base transaction count
    base_count: usize,

    /// Base transaction size
    base_size: usize,

    /// How much the transaction count can vary
    count_variation: std::ops::Range<i32>,

    size_variation: std::ops::Range<i32>,

    /// Chance of generating a spike.
    /// e.g. 0.1 = 10% chance of spike
    spike_probability: f64,

    /// Multiplier for spike transactions
    /// e.g. 10 = 10x more transactions during spike
    spike_multiplier: usize,

    /// Range of intervals between generating load, in milliseconds
    sleep_interval: std::ops::Range<u64>,
}
```

```rust
tokio::spawn(move async {
// Determine if this iteration should generate a spike
let is_spike = rng.gen_bool(params.spike_probability);

// Vary transaction count and size
let count_variation = rng.gen_range(params.count_variation);
let size_variation = rng.gen_range(params.size_variation);

let count = if is_spike {
    (params.base_count + count_variation) * params.spike_multiplier
} else {
    params.base_count + count_variation
};

let size = params.base_size + size_variation;

// Create and send the message
let msg = HostMsg::GenerateTransactions {
    count: count.max(1), // Ensure count is at least 1
    size: size.max(1),   // Ensure size is at least 1
};

if myself.send(msg).is_err() {
    println!("Channel closed, stopping load generator");
    break;
}

// Random sleep between 100ms and 1s
let sleep_duration = Duration::from_millis(params.sleep_interval);
sleep(sleep_duration);
```

Note that this does not require `ticker` as the sleep duration is random.

romac (changes requested):

code/crates/config/src/lib.rs
```diff
@@ -339,6 +342,208 @@ mod gossipsub {
...
+impl Default for MempoolLoadType {
+    fn default() -> Self {
+        Self::NonUniformLoad(NonUniformLoadConfig::default())
+    }
+}
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(from = "nonuniformload::RawConfig", default)]
+pub struct NonUniformLoadConfig {
+    /// Base transaction count
+    base_count: usize,
```
romac, line 362: These fields should probably be public

code/crates/config/src/lib.rs
```diff
@@ -339,6 +342,208 @@ mod gossipsub {
...
+    NonUniformLoad(NonUniformLoadConfig),
+}
+
+impl Default for MempoolLoadType {
+    fn default() -> Self {
+        Self::NonUniformLoad(NonUniformLoadConfig::default())
+    }
+}
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(from = "nonuniformload::RawConfig", default)]
```
romac, line 359: Why not derive Deserialize directly on this structure instead of passing by a `RawConfig`?

code/crates/config/src/lib.rs
```diff
@@ -339,6 +342,208 @@ mod gossipsub {
...
+                raw.size_variation,
+                raw.spike_probability,
+                raw.spike_multiplier,
+                raw.sleep_interval,
+            )
+        }
+    }
+}
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(from = "uniformload::RawConfig", default)]
```
romac, line 487: Same remark and question as above

code/crates/starknet/host/src/mempool_load.rs
```diff
@@ -0,0 +1,178 @@
...
+                };
+
+                if let Err(er) = myself.cast(msg) {
+                    tracing::error!(?er, ?myself, "Failed to send tick message");
+                    // break;
+                }
+                // Random sleep between 100ms and 1s
+                let sleep_duration =
+                    Duration::from_millis(params.sleep_interval().choose(&mut rng).unwrap());
+                sleep(sleep_duration);
```
romac, line 137:
```suggestion
                tokio::time::sleep(sleep_duration).await;
```

code/crates/config/src/lib.rs
OakenKnight: I saw that it was done like that for `GossipSubConfig` so I thought that was a standard.

code/crates/starknet/host/src/mempool_load.rs
```diff
@@ -0,0 +1,167 @@
...
+            MempoolLoadType::UniformLoad(uniform_load_config) => tokio::spawn(ticker(
+                uniform_load_config.interval,
+                myself.clone(),
+                move || Msg::GenerateTransactions {
+                    count: uniform_load_config.count,
+                    size: uniform_load_config.size,
+                },
+            )),
+            MempoolLoadType::NoLoad => tokio::spawn(async {}),
+            MempoolLoadType::NonUniformLoad(params) => tokio::spawn(async move {
```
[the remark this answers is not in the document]
OakenKnight: The one from UniformLoad is calling [ticker](https://github.com/informalsystems/malachite/blob/cdedc723f43b222a2d68217574fb1a20ccaae470/code/crates/starknet/host/src/utils/ticker.rs#L5) async function.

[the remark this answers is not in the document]
romac: That's fine, you can move the whole `ticker(...)` expression into its own function that takes `UniformLoadConfig` as a parameter.

code/crates/config/src/lib.rs
romac: Ah I see! This was only done for GossipSubConfig to make sure we would go through it's constructor in order to fine-tune the parameters

adizere:
Thanks @oakenknight ! Do you consider this ready for review?

oakenknight:
> @oakenknight I think the issue was that when there was no load, we would generate blocks too quickly for other nodes to catch up with us in the full nodes tests.
> 
> See [a7ac34d](https://github.com/informalsystems/malachite/commit/a7ac34d8590e248afd310df528ac885a6468aafd) for the fix.

Ah, I see. That makes sense.

Issue #830, opened 2025-02-05
tutorial: Optimize proposal part handling logic in `received_proposal_part`


romac:
## Summary

The current implementation checks for full proposal before validating height, which is inefficient as outdated parts are still being inserted into the map.

```rust
    pub fn received_proposal_part(
        &mut self,
        from: PeerId,
        part: StreamMessage<ProposalPart>,
    ) -> Option<ProposedValue<TestContext>> {
        let sequence = part.sequence;

        // Check if we have a full proposal
        let parts = self.streams_map.insert(from, part)?;

        // Check if the proposal is outdated
        if parts.height < self.current_height {
```

## Action items

- [ ] Reorder checks to validate height before inserting into `streams_map`

```rust
        // Check if the part is outdated
        if part.height < self.current_height {
            debug!(
                height = %self.current_height,
                round = %self.current_round,
                part.height = %parts.height,
                part.round = %parts.round,
                part.sequence = %part.sequence,
                "Received outdated proposal part, ignoring"
            );

            return None;
        }

        // Check if we have a full proposal
        let parts = self.streams_map.insert(from, part)?;

        // Re-assemble the proposal from its parts
        let value = assemble_value_from_parts(parts);

        self.undecided_proposals
            .insert((value.height, value.round), value.clone());

        Some(value)
    }
```

varun-doshi:
I'd like to take this one

romac:
@varun-doshi Awesome, thank you!

romac:
I am afraid that the solution I outlined above does not actually solve the problem, as it only filters out an outdated `Init` part but not any other part types.

I was thinking we could maybe use a bloom-filter of some kind to store stream ids of "outdated streams" (ie. stream ids for which we have seen an outdated `Init` part), but that does not work either in general because parts can arrive in any order and the Init part might arrive last. Therefore, an attack could still DoS a node by sending lots of outdated parts with the Init message at the end only.

PR #837, opened 2025-02-07
feat: Optimize proposal part handling logic in received_proposal_part

varun-doshi:
Closes: #830 

---

#### For all contributors

- [x] Reference GitHub issue

### DIFF
--- a/code/examples/channel/src/state.rs
+++ b/code/examples/channel/src/state.rs
@@ -115,6 +115,28 @@ impl State {
     ) -> eyre::Result<Option<ProposedValue<TestContext>>> {
         let sequence = part.sequence;
 
+        if let Some(proposal) = part
+            .content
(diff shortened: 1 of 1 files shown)


romac:
Thanks for the PR! I need to think a bit more on it, because I am afraid that the solution I outlined in the PR and that you've implemented does not actually solve the problem, as it only filters out an outdated Init part but not any other part types.

I was thinking we could maybe use a bloom-filter of some kind to store stream ids of "outdated streams" (ie. stream ids for which we have seen an outdated Init part), but that does not work either in general because parts can arrive in any order and the Init part might arrive last. Therefore, an attack could still DoS a node by sending lots of outdated parts with the Init message at the end only.

romac:
Closing until we figure out how to address this.

PR #840, opened 2025-02-07
chore(code): Some WAL entries fail to decode

ancazamfir:
Closes: #XXX
Sometime wal entries fail to decode with errors like this:
```
2025-02-07T20:30:44.054030Z ERROR node{moniker=test-2}:wal{height=144}: Failed to decode WAL entry 4: failed to fill whole buffer
```
Looks like for `PrevoteTimeLimit` and `PrecommitTimeLimit` we still write something to the WAL (just the tag) and later fail to decode.
 - [x] Add test that shows the error
 - [x] Fix the error
      
Note: A better fix might be to move the entry filtering logic in consensus who would decide which entries should be logged. For timeouts it could check the type [here](https://github.com/informalsystems/malachite/blob/fec4b6ed39637db75d742a7bf002eb9b56008e9f/code/crates/engine/src/consensus.rs#L1049-L1053)

---

#### For all contributors

- [ ] Reference GitHub issue

### DIFF
--- a/code/crates/engine/src/wal/entry.rs
+++ b/code/crates/engine/src/wal/entry.rs
@@ -60,15 +60,15 @@ where
     {
         match self {
             WalEntry::ConsensusMsg(msg) => {
-                // Write tag
-                buf.write_u8(Self::TAG_CONSENSUS)?;
-
                 let bytes = codec.encode(msg).map_err(|e| {
                     io::Error::new(
                         io::ErrorKind::InvalidData,
                         format!("failed to encode consensus message: {e}"),
                     )
                 })?;
+
+                // Write tag
+                buf.write_u8(Self::TAG_CONSENSUS)?;
 
                 // Write encoded length
                 buf.write_u64::<BE>(bytes.len() as u64)?;
--- a/code/crates/engine/src/wal/entry.rs
+++ b/code/crates/engine/src/wal/entry.rs
@@ -80,11 +80,8 @@ where
             }
 
             WalEntry::Timeout(timeout) => {
-                // Write tag
-                buf.write_u8(Self::TAG_TIMEOUT)?;
-
-                // Write timeout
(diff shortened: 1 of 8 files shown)


code/crates/engine/src/wal/entry.rs
```diff
@@ -80,11 +80,8 @@ where
             }
 
             WalEntry::Timeout(timeout) => {
-                // Write tag
-                buf.write_u8(Self::TAG_TIMEOUT)?;
-
-                // Write timeout
-                encode_timeout(timeout, &mut buf)?;
+                // Write tag and timeout if applicable
+                encode_timeout(Self::TAG_TIMEOUT, timeout, &mut buf)?;
```
romac, line 84: Nice catch! Thanks so much for figuring this out 🙌

ancazamfir: Wondering if we should have consensus decide which events should be logged and have wal write everything it's told to (by consensus).

code/crates/starknet/test/src/lib.rs
```diff
@@ -530,14 +530,22 @@ async fn run_node<S>(
...
+                    match &event {
+                        Event::StartedHeight(height) => {
+                            info!("Node started height {height}");
 
-                    info!("Node started height {height}");
+                            if height.as_u64() == target_height {
+                                break 'inner;
+                            }
+                        }
+                        Event::WalReplayError => {
```
ancazamfir, line 541: Not sure this is the best place to catch errors (WAL or others) in tests.

code/crates/engine/src/wal/entry.rs
romac: 80afd66d6d78321b39a9e65442a75c1c2a880421 + ea04fba76a3df74089432a607d494a51a268b0b3

romac (approved):

PR #853, opened 2025-02-13
fix(code/core-consensus): Reproduce and fix consensus being stuck with byzantine proposer

ancazamfir:
---

To be merged via #859

#### For all contributors

- [ ] Reference GitHub issue


code/crates/test/app/src/state.rs
```diff
@@ -50,12 +50,17 @@ pub struct State {
 // order for each node to likely propose different values at
 // each round.
 fn seed_from_address(address: &Address) -> u64 {
-    address.into_inner().chunks(8).fold(0u64, |acc, chunk| {
```
romac, line 53: Good idea! Maybe we should allow the test framework to trigger such behavior, what do you think? Will think about how to implement that properly

romac, line 53: Good idea! Maybe we should allow the test framework to trigger such behavior, what do you think? Will think about how to implement that properly

ancazamfir:
@romac what do you think about making a minimal PR with just the changes required to eliminate the vote equivocation.
Meaning only with the changes in:
```
core-consensus/src/handle/driver.rs
core-consensus/src/state.rs
core-driver/src/driver.rs
```
We can merge to main if CI is green and then we can continue on this PR.

romac:
Merged in #859

Issue #850, opened 2025-02-12
Consensus stuck in votesync upon restart


aditiharini:
On restarts where the proposer has to restream a precommited value (i.e. hits`RestreamValue`) consensus halts in votesync. 

The non-proposing nodes don't have the precommit votes for the restreamed value though the value has a pol round on it. The proposing node, never gets votes from its peers and gets stuck in votesync indefinitely.

Logs attached:

[node1-shard8.txt](https://github.com/user-attachments/files/18774646/node1-shard8.txt)

[node2-shard8.txt](https://github.com/user-attachments/files/18774657/node2-shard8.txt)

[node3-shard8.txt](https://github.com/user-attachments/files/18774668/node3-shard8.txt)

ancazamfir:
Thanks @aditiharini for the issue.

Looked at the logs and here is a preliminary analysis
Looks like nodes restart, with last decided height `3290` and they move to `height 3291, round 0`
There are two values that show up:
- `v1 = 2e32197974e8513749480d257026d4806465b4167ea50b841f5b2abd2aa2f76f`
- `v2 = 02879ba954a442c6f60ed9f4061c938396068ce65f7ba09268fea685a91ad761`
They are being proposed for the same height and round by Node 2 so we deal with a byzantine proposer. 
We are also not handling correctly equivocation, causing Node 1 to lock on `v1` which breaks consensus.

Node 1:
=======
- `id = 51141fcc0a9b6d81f5702a12150e066c9e2998100aedf250703ccac6cbfac22b`
- proposer at (3291, 2)
- 0 WAL entries
- receives proposal for `v2` from Node 2
- votes for `v2`
- receives vote for `v2` from Node 2
- receives vote for `v2` from Node 3 => ProposalAndPolkaCurrent
**- locks on `v2`**

Node 2:
======
- `id = 75d3277a1b4bb12f9ccd873a4792a8a50c65b91e44edd069e189428fe1c90052`
- proposer at (3291, 0)
- replays 3 WAL entries:
  - replays  its proposal for `v1` 
  - replays its prevote for `v1`
  - replays prevote for `v1` from Node 3
- **proposes `v2`,  a different value for same height and round**
- **votes for `v2` => goes out but dropped internally, considered equivocation**
- receives prevote for `v2` from Node 1 => PolkaAny
- receives prevote for `v2` from Node 3, discards it


Node 3:
=======
[...]

PR #859, opened 2025-02-18
feat(code): Add polka certificates to VoteSync response

romac:
Closes: #850

## Summary
This PR adds support for polka certificates to the VoteSync protocol. Polka certificates encapsulate votes that demonstrate a polka (quorum of prevotes) for a specific value at a given height and round.

## Changes

- Introduce a new `PolkaCertificate` struct alongside the existing `CommitCertificate` to capture evidence of polkas
- Modify the VoteSync protocol to include polka certificates in vote set responses
- Implement certificate-based recovery mechanisms in the consensus driver
- Update consensus state handling to track and use polka certificates during rounds
- Improve proposal evaluation logic to consider polka certificates when validating proposals
- Add storage and retrieval mechanisms for polka certificates
- Update message multiplexer to handle polka certificates for proper round state transitions


code/crates/core-consensus/src/handle/vote_set.rs
```diff
@@ -33,18 +37,31 @@ pub async fn on_vote_set_response<Ctx>(
...
+    if let Some(certificate) = polka_certificate {
+        apply_driver_input(
+            co,
+            state,
+            metrics,
+            DriverInput::PolkaCertificate(certificate),
+        )
+        .await?;
+    } else {
+        for vote in vote_set.votes {
```
romac, line 62: @ancazamfir I don't believe we need to process both the votes and the certificate, but perhaps I am mistaken?

code/crates/core-consensus/src/state.rs
```diff
@@ -134,17 +134,30 @@ where
...
-        } else {
-            vec![]
+        let per_round = self.driver.votes().per_round(round)?;
+        let votes = per_round
+            .received_votes()
+            .iter()
+            .cloned()
+            .collect::<Vec<_>>();
+
+        if votes.is_empty() {
```
romac, line 155: I wonder if this is wrong, in case we have a polka certificate for a previous round?

code/crates/core-driver/src/driver.rs
```diff
@@ -197,16 +204,24 @@ where
...
             .iter()
             .find(|c| c.round == round && c.value_id == value_id)
     }
 
+    /// Get a polka certificate for the given round.
+    pub fn get_polka_certificate(&self, round: Round) -> Option<&PolkaCertificate<Ctx>> {
+        self.polka_certificates
+            .iter()
+            .filter(|c| c.round <= round)
+            .last()
```
romac, line 222: Here I did something similar as we do for commit certificates but it feels very hacky and I am not convinced it is right. Will check the Quint spec again to see whether or not this is correct.

code/crates/core-types/src/certificate.rs
```diff
@@ -86,6 +86,19 @@ impl<Ctx: Context> CommitCertificate<Ctx> {
...
+#[derive_where(Clone, Debug, PartialEq, Eq)]
+pub struct PolkaCertificate<Ctx: Context> {
+    /// The height at which a Polka was witnessed
+    pub height: Ctx::Height,
+    /// The round at which a Polka that was witnessed
+    pub round: Round,
+    /// The value that the Polka is for
+    pub value_id: ValueId<Ctx>,
+    /// The votes that make up the Polka
+    pub votes: Vec<SignedVote<Ctx>>,
```
romac, line 99: Right now I am just storing the votes like this, we can switch to an aggregated signature at a later stage if needed.

code/crates/core-driver/src/driver.rs
```diff
@@ -395,15 +438,39 @@ where
...
+    fn store_polka_certificate(&mut self, vote_round: Round, value_id: &ValueId<Ctx>) {
+        let Some(per_round) = self.vote_keeper.per_round(vote_round) else {
+            return;
+        };
+
+        self.polka_certificates.push(PolkaCertificate {
+            height: self.height(),
+            round: vote_round,
+            value_id: value_id.clone(),
+            votes: per_round
```
romac, line 462: TDOO: Like in the [Quint spec](https://github.com/informalsystems/malachite/pull/364/files#diff-ef09dc528370554b2e2e64658d529bd0666761d785a190f2fe6bd0105fa719ccR545), we could double-check that there is enough voting power in this polka, but maybe that's overkill?

romac:
Ignore the _Rust / Standalone_ CI failure, it is expected because of the debug statements I added in the driver.

romac:
Looks like I broke the Starknet tests, which is unexpected.

code/crates/core-driver/src/driver.rs
```diff
@@ -395,15 +438,39 @@ where
...
 
+        let round_input = self.multiplex_vote_threshold(output, vote_round);
         if round_input == RoundInput::NoInput {
             return Ok(None);
         }
 
         self.apply_input(vote_round, round_input)
     }
 
+    fn store_polka_certificate(&mut self, vote_round: Round, value_id: &ValueId<Ctx>) {
```
romac, line 453: Ref: https://github.com/informalsystems/malachite/blob/ad7350e81006dd28a1351c4eb33a2349208951cf/specs/consensus/quint/votekeeper.qnt#L541

code/crates/core-driver/src/mux.rs
```diff
@@ -216,6 +216,41 @@ where
         None
     }
 
+    pub(crate) fn store_and_multiplex_polka_certificate(
```
romac, line 219: Ref: https://github.com/informalsystems/malachite/blob/ad7350e81006dd28a1351c4eb33a2349208951cf/specs/consensus/quint/votekeeper.qnt#L445

romac: Looking at the spec again, I believe this method is not correct. Instead of sending a `ProposalAndPolkaCurrent` to the state machine, we should instead follow the same logic as the spec and override the votes in the vote keeper.

We should add the `apply_certificate` method to the `VoteKeeper` struct and implement that logic there.

code/crates/core-driver/src/driver.rs
romac: TODO: Check that the certificate we have constructed is valid.

romac: TODO: Rename this method to `get_polka_certificate`

romac: TODO: Move this method to the `VoteKeeper` and only store the certificate in `polka_certificates` here

romac:
Note: The two relevant integration tests for this PR are:
- `wal::restart_with_byzantine_proposer_1_request_response`
- `wal::restart_with_byzantine_proposer_2_request_response`

They can be run with:

```
$ cargo nextest run -p informalsystems-malachitebft-test --no-capture restart_with_byzantine_proposer_1_request_response --retries 0
$ cargo nextest run -p informalsystems-malachitebft-test --no-capture restart_with_byzantine_proposer_2_request_response --retries 0
```

(Requires `cargo-nextest`, can be installed with `cargo install cargo-nextest`)

nenadmilosevic95:
Hey @romac, I think it will be very useful if we add a description to this PR. As from what I understood it is not a bug fix but rather a change that should not only unstuck consensus in case of a Byzantine behavior observed 2nd scenario in #850 but also lead to the consensus decision. 

I can maybe help with writing this description but I need more context, for example where can I learn how vote sync works and when it is triggered?

romac:
> I can maybe help with writing this description but I need more context, for example where can I learn how vote sync works and when it is triggered?

We do not habe a spec for this yet, but the tracking issue for it has some background info: https://github.com/informalsystems/malachite/issues/576 Let me know if that helps, otherwise I can provide more details.

cason:
Recall that we might need a Polka certificate for a previous round (`vr < round_p`) in case of line 28.

code/crates/core-consensus/src/state.rs
ancazamfir: Yes, we need certificates from previous rounds as well.

code/crates/core-driver/src/mux.rs
ancazamfir: We can do that. But we can also store/ retrieve the certificate as you do here. In this case we also have to change `multiplex_proposal()` and how we determine `polka_current` and `polka_previous`.

code/crates/core-consensus/src/handle/vote_set.rs
romac: Do you mean it's fine to process both?

romac:
It's also surprising that the rebroadcast test has passed on CI, where I would have expected that we'd need to include the polka certificates in the rebroadcast for it to pass.

ancazamfir:
> The `wal::restart_with_byzantine_proposer_1_request_response` test has passed on CI but it does not pass reliably locally. Need to investigate further.
> 
> @ancazamfir Can you give it a go when you have time and see if you see the same?
> 
> ```
> $ cargo nextest run -p informalsystems-malachitebft-test --no-capture --retries 0 wal::restart_with_byzantine_proposer_1_request_response
> ```

It passes for me, it was passing even when CI was failing it. I tried today again 10 times and was about to ask you to try it :)

code/crates/test/tests/it/wal.rs
```diff
@@ -289,7 +291,7 @@ async fn byzantine_proposer_crashes_after_proposing_1(params: TestParams) {
             TestParams {
                 enable_value_sync: true,
                 timeout_step: Duration::from_secs(5),
-                value_payload: ValuePayload::ProposalAndParts,
+                value_payload: ValuePayload::PartsOnly,
```
ancazamfir, line 294: Not sure why we need this change. Looking at the test app I think in general it's more accurate to use `ProposalAndParts` because for re-streaming the `Init` doesn't seem to carry `valid_round` so the receiver cannot reconstruct the `valid_round` without an explicit Proposal message.
For the purpose of these tests it's fine but I think for other cases we should be able to test with both.

romac: I had to do this because of this check in the test app:

https://github.com/informalsystems/malachite/blob/7e6c49dc8f39dcc840688da7fbbcae44c14fe5b4/code/crates/test/app/src/app.rs#L127-L131

romac: But this looks arbitrary, let me attempt to remove it and re-enable to proposal-and-parts tests.

ancazamfir:
On this branch we have these failure. Not sure entirely if polka certificates would help all these cases but for now only the first one (`wal::restart_with_byzantine_proposer_1_request_response_proposal_and_parts`) is important to look at and fix.

The polka cert is not yet implemented for rebroadcast and the it app we use doesn't support parts only properly when it comes to L28.

I opened #899 that hacks the driver to always do vote equivocation if proposer in initial height and round 0. That uncovered a few issues:
- parts-only is not properly implemented in the test app (missing `valid_round` in `Init`)
- restreaming is broken in the test app
- missing polka certificate in the driver when a proposer is received, L28 not handled
- incorrect `input-round` when muxing the polka certificate, L28 not handled

I have fixed in that branch all except the first one. Will continue to fix and debug in that branch.

ancazamfir (approved):

PR #860, opened 2025-02-19
feat(code/engine): Buffer consensus messages during startup and recovery

romac:
> @ancazamfir: Votes may be processed before WAL replay (https://github.com/informalsystems/malachite/issues/850#issuecomment-2659910086)

---

This PR introduces message buffering for the consensus engine during startup and recovery phases. This is an important improvement to handle messages that arrive before the consensus engine is fully started or while it is recovering state via the WAL.

- Add a `VecDeque` buffer to store messages that arrive during `Unstarted`/`Recovering` phases
- Implement a message filtering function to determine which messages should be buffered
- Add processing of buffered messages once consensus starts
- Set a maximum buffer size of 1024 messages to prevent unbounded memory growth

Messages that should be buffered include votes, proposals, and proposal parts. Control messages like `StartHeight`, `TimeoutElapsed`, and network connection events are processed immediately.

The buffered messages are replayed in order once:
1. The WAL has been checked and replayed
2. The consensus engine transitions to the `Running` phase

The buffer size is capped at 1024 messages to prevent memory exhaustion attacks. Messages beyond this limit are dropped with a warning log.

---

#### For all contributors

- [ ] Reference GitHub issue

### DIFF
--- a/code/crates/engine/src/consensus.rs
+++ b/code/crates/engine/src/consensus.rs
@@ -2,11 +2,13 @@ use std::collections::BTreeSet;
 use std::sync::Arc;
 use std::time::Duration;
 
+use async_recursion::async_recursion;
 use async_trait::async_trait;
+use derive_where::derive_where;
 use eyre::eyre;
 use ractor::{Actor, ActorProcessingErr, ActorRef, RpcReplyPort};
 use tokio::time::Instant;
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, error_span, info, warn};
 
 use malachitebft_codec as codec;
 use malachitebft_config::TimeoutConfig;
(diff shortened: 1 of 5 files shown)


code/crates/engine/src/consensus.rs
```diff
@@ -1142,9 +1178,19 @@ where
...
-
         Ok(())
     }
 }
+
+fn should_buffer<Ctx: Context>(msg: &Msg<Ctx>) -> bool {
+    !matches!(
+        msg,
+        Msg::StartHeight(..)
+            | Msg::TimeoutElapsed(..)
```
ancazamfir, line 1190: Why do we need to let the `TimeoutElapsed` through?

ancazamfir (approved):
Looks good!

code/crates/engine/src/consensus.rs
romac: No that reason I can think of actually, since no timeouts are even scheduled by that point. I will remove it.

nenadmilosevic95:
Hey @romac , does `StartHeight` influence anyhow previous height? Meaning what is happening if app calls StartHeight in the middle of the previous height which can theoretically happen in this scenario? Or this is not possible?

cason:
Not clear to me if all buffered events refer to the current height, or also to future heights.

And why timeout events are not buffered?

PR #864, opened 2025-02-19
fix(code/core-consensus): Prevent nodes from broadcasting equivocating votes

romac:
Closes: #XXX

---

#### For all contributors

- [ ] Reference GitHub issue


ancazamfir (approved):
Looks great!

nenadmilosevic95:
Hey @ancazamfir @romac, I just have one clarification question with respect to this PR. 

I do not understand why do we need extra logic to prevent vote equivocation in case of recovery. So in Tendermint algo the vote equivocation is prevented because process sets round step to prevote after issuing a vote, and if it receives another proposal it will not accept it. 

This mechanism does not work here because process votes for value v1 and then replay the prevote for v2? If this is the case, shouldn't we prevent processing of messages before we process the replayed messages, and then we should not have this problem? 

Sorry if I am saying something stupid but I am trying to have the full context.

ancazamfir:
> Hey @ancazamfir @romac, I just have one clarification question with respect to this PR.
> 
> I do not understand why do we need extra logic to prevent vote equivocation in case of recovery. So in Tendermint algo the vote equivocation is prevented because process sets round step to prevote after issuing a vote, and if it receives another proposal it will not accept it.

In Malachite things are a bit more complex. The application owns the full value to be proposed. The setup in #850 is very similar to Comet, in the sense that there is block part dissemination and also a `Proposal(h, r, v, vr)` message. We can assume `v = BlockHash(block)` and `id(v) = v`.
Consensus engine manages also the "full proposal":
- At non-proposer, a proposal is considered complete when the application has signaled to consensus that it has all parts (`ProposedValue(v)`) and an explicit Proposal message has been received from the wire.
- At proposer, a proposal is complete when a reply to `GetValue()` has been received. In this case we send a proposal message out on the wire and also store it internally.

In both cases, when we have the propose value message from the app AND the proposal wire message, we send what is shown in the paper as `(PROPOSAL, h, r, v, vr)` to the driver module and then to the round state machine

Now, in the "full proposal" module we store everything. For example we can have:
- consensus receives `Proposal(v2)` wire message
- application sends `ProposedValue(v1)`
- consensus receives `Proposal(v1)` wire message

At this point we match the last two, determine we have a complete proposal and we send `(PROPOSAL, h, r, v,1 vr)` to the driver. If we then get `ProposedValue(v2)` from the app and send `(PROPOSAL, h, r, v,2 vr)` the driver will record this as proposal equivocation and drop the second message.

So to summarize, the full proposal module stores multiple proposal related messages and can store multiple full or partial proposals. The driver accepts only the first proposal, drops the subsequent ones. The driver also store votes and records the vote equivocation (drops subsequent votes for different values and same round)

For restart scenarios, consensus saves the proposal (wire message) and votes in the WAL. It is expected that the application saves the undecided block before notifying consensus.

In the first scenario of #850, consensus saves `Proposal(h, 0, v1, -1)` to WAL but not the block parts since it never sees them. It also saves its own `Prevote(h, 0, id(v1))` and the one from Node 3.
On restart, consensus first replays the proposal and the two vote messages. These replays are "internal", nothing is sent out.
Note also that without the application `ProposedValue(v1)` the proposal is not yet complete and it is on hold in the full proposal module.

Next consensus starts round `0` and, if proposer, asks the application for a value, waiting in `Propose` step. The expectation in this case is that the application disseminates the parts for the block with hash `v1` and returns `v1` to consensus. When that happens we have a "full proposal" and we send `Proposal(h, 0, v1, -1)` out to the peers and also to the driver, then to the round state machine which executes L22 and instructs the engine to send `Prevote(h, 0, v1)`.

Now if the app doesn't replay `v1` but rather `v2` then the full proposal module will send `Proposal(h, 0, v2, -1)` to the driver and the effect is that a `Prevote(h, 0, id(v2))` is proposed by the round state machine. The driver will drop this because it already has a vote for `v1`.
The bug that this PR fixes is for that `Prevote(h, 0, id(v2))` not to be sent on the wire.



> 
> This mechanism does not work here because process votes for value v1 and then replay the prevote for v2? If this is the case, shouldn't we prevent processing of messages before we process the replayed messages, and then we should not have this problem?

This was yet a different bug that we fixed in #860
> 
> Sorry if I am saying something stupid but I am trying to have the full context.

These are all very good questions. We talked about providing details on the architecture and design of malachite in a document, also showing the different flows and interactions. Never got to it but maybe now it's a good time.

Just wanted to add that there is more work to unstuck consensus when dealing with equivocating proposer. This will require dissemination of polka certificates  (#859) but we will also need to allow multiple proposals in the driver.

nenadmilosevic95:
Thanks @ancazamfir ! I think I understand it better now! :)

>Now if the app doesn't replay v1 but rather v2 then the full proposal module will send Proposal(h, 0, v2, -1) to the driver and the effect is that a Prevote(h, 0, id(v2)) is proposed by the round state machine. The driver will drop this because it already has a vote for v1.
The bug that this PR fixes is for that Prevote(h, 0, id(v2)) not to be sent on the wire.

IMHO even if this PR is solving this bug, I think that correct implementation should not prevent this message to be sent on a wire but it should prevent these equivocated messages to be generated in the first place. So correct process should never generate an equivocated proposal or prevote in any scenario. 

So in this specific scenario since node sent Proposal(v1) and Prevote(v1), this means that it received also ProposedValue(v1) from the app? So my question here is why don't we store also this event and when restarting replay also this? Because this way you are calling `getValue` two times for the same round of consensus, and isn't this wrong behavior? 

Just one more thing to check my understanding: 
- Do I understand correctly that consensus only works with hashes and it never stores full values? Instead the app is responsible for disseminating the value and it signals to consensus with `ProposedValue(v)` or `getValue` (in case of a proposer) when it has the full value. 
- How does this replay logic works? When we say it replays a certain message, what does it mean? I saw that you wrote it is "internal". Can you share bit more context about this please? 

>Just wanted to add that there is more work to unstuck consensus when dealing with equivocating proposer. This will require dissemination of polka certificates (https://github.com/informalsystems/malachite/pull/859) but we will also need to allow multiple proposals in the driver.

I discussed this a bit with Daniel, I would really like to understand the progress here. Can you please keep me in the loop about this? And I will check the PR.

cason:
Should we panic when attempting to sign an equivocating vote? Or at least produce a very loud log message?

cason:
This solution, like Comet's ones, also prevents us to rebroadcast "old" votes. Namely, if we have produce a vote for round `(h, r)` we cannot rebroadcast a vote for round `(h', r')` with `h' < h` or `(r'< r)`, even it would be a legit (non-equivocating vote).

Finally, I find confusing here that we care about heights. We are in this case strictly requiring heights to be executed in lock-step, which limits some use cases for Malachite.

PR #870, opened 2025-02-20
feat(code/core-consensus): Add `vote_sync_mode` option to consensus parameters

romac:
Closes: #XXX

Instead of relying on whether or not the Sync actor is enabled, use a bespoke configuration parameter to decide whether or not to rebroadcast votes.

This PR also renamed the `sync` configuration option to `value_sync`:

```toml
#######################################################
###         Sync Configuration Options         ###
#######################################################
[value_sync]
# Enable the synchronization protocol
# Override with MALACHITE__VALUE_SYNC__ENABLED env variable
enabled = true

# Interval at which to update other peers of our status
# Override with MALACHITE__VALUE_SYNC__STATUS_UPDATE_INTERVAL env variable
status_update_interval = "10s"

# Timeout duration for sync requests
# Override with MALACHITE__VALUE_SYNC__REQUEST_TIMEOUT env variable
request_timeout = "10s"
```

---

#### For all contributors

- [ ] Reference GitHub issue


code/crates/engine/src/consensus.rs
```diff
@@ -1047,7 +1046,7 @@ where
                 Ok(r.resume_with(()))
             }
 
-            Effect::GetVoteSet(height, round, r) => {
+            Effect::RequestVoteSet(height, round, r) => {
```
ancazamfir, line 1049: Before this PR, with:
```
[sync]
enabled = false
```
the `rebroadcast` was taking effect, otherwise the `request-response`.
After this PR, with:
```
[consensus.vote_sync]
mode = "request-response"
...
[sync]
enabled = false
```
consensus will halt. It is a misconfiguration and we should maybe detect this earlier or at least log an error here if `self.sync.is_none()`.

romac: How about we change the `sync` section to be named `value_sync`, and spawn the `Sync` actor when either `value_sync` is true or `consensus.vote_sync.mode = "request-response"`?

Or maybe we should split the Sync actor into a ValueSync and VoteSync actor. Is that something you had started working on or not, I can't rembember.

ancazamfir: Wondering if we should we have `enabled` for `vote_sync` as well. I lean against it tbh since consensus will get stuck under some conditions if there is no sync at that level. And it's not easy or expected that an app will develop their own (like in the case of value_sync).

romac: Agreed, let's enforce VoteSync for now.

ancazamfir (approved):

cason:
Can't we have both?

The point here is that rebroadcasting is not really a VoteSync protocol, but anyway.

PR #884, opened 2025-03-04
feat(docs): ADR 003 - Propagation of Proposed Values

cason:
Closes:  ~~365~~ #937 

---

#### For all contributors

- [x] Reference GitHub issue


docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,214 @@
...
+- `PREVOTE` and `PRECOMMIT` messages carry an identifier `id(v)` of a proposed
+  value `v`, or the special value `nil` meaning "no value", and play the role
+  of **Value Decision**.
+  They are referred to as [**votes**][consensus-votes], and are broadcast by
+  all processes, respectively, in the `prevote` and `precommit` round steps.
+
+Note that `PROPOSAL` messages have also a role on **Value Decision**, as they
+carry consensus-related fields that are validated by processes and have
+implications in the algorithm.
+The point is that they are the only messages with **Value Dissemination** role,
```
adizere, line 58:
```suggestion
The point is that they are the only messages with **Value Propagation** role,
```

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,214 @@
...
+    Proposal(Ctx::Proposal, Validity),
+
+    /// Receive a vote
+    Vote(Vote<Ctx>),
+}
+```
+
+The issues that this ADR is meant to address are mainly two:
+
+1. The fact that [ADR 001][adr001] is not any longer in line with the implementation;
```
adizere, line 156: I think we should not focus on this in this ADR.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,214 @@
...
+2. The ways that applications, given Malachite consensus' interface, handle the
```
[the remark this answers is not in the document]
adizere: Note that this this topic is a subset of the broader topic we should eventually cover (elsewhere). The broader topic is, in the words of a Malachite user:

> one thing that would be very valuable is some more information on how to drive the low-level consensus engine.

We should not tackle how to "drive the low-level consensus engine" here. We should tackle "how to drive the low-level consensus engine _specifically_ for the steps of propagation of proposed values." Do I understand correctly this is what you meant by "The ways that applications, given Malachite consensus' interface, handle the propagation of proposed values" ? If so, then we're aligned!

adizere (commented):
Thanks Daniel, really appreciate starting this!

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
adizere: By the way, we should not use the word engine in the ADR. :D

adizere:
Shifted and made the scope of this ADR more narrow. It should simply tackle #937

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Current design
+
+At the moment, Malachite consensus implementation is supporting three
+different modes of operation to handle value propagation: 
+1) **ProposalOnly**
+2) **PartsOnly**
+3) **ProposalAndParts**
+
+A specific mode can be set as a `value_payload` consensus parameter (TODO: not sure if this is right code reference):
```
romac, line 66: Yep, that's the right place

```suggestion
A specific mode can be set as a `value_payload` consensus parameter:
```

ancazamfir (dismissed):

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,351 @@
...
+the network. In contrast, the **Value Decision** phase requires only 
+the transmission of vote messages, which are relatively small and of 
+constant size. As a result, Malachite’s low-level API provides greater 
+flexibility in the value propagation stage to enable more optimized 
+implementations. 
+
+In this document, we focus on the core parts of the consensus 
+implementation responsible for **Value Propagation**.
+
+## Current design
```
cason, line 58: We should first define the blocks used in the following:
- Consensus core
- Network
- Application
- Consensus engine (= Core + network?)

cason (commented):
Look good. Left some comments from our sync meeting.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,351 @@
...
+
+At the moment, Malachite consensus implementation is supporting three
```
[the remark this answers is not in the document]
romac:
```suggestion
At the moment, the Malachite consensus engine is supporting three
```

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,369 @@
...
+
+    box Other nodes
+      participant E2 as Consensus Engine
+      participant C2 as Consensus Core
+      participant A2 as Application
+    end
+
+    C1->>A1: getValue()
+    A1->>C1: Propose(LocallyProposedValue(v))
+    C1--xE1: Effect::Publish(SignedProposal(v))
```
ancazamfir, line 236: TODO: 
- Fix diagram and later the code. The actual drop happens in the consensus core today but couldn't figure out mermaid arrow to point to nothing
- On receive side the drop happens in the consensus engine, this should be shown somehow

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
ancazamfir: I think it's fine "consensus implementation", no? I understand we want the consensus core to (eventually) be agnostic of the mode but even then we can say it does support everything.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,377 @@
...
+
+It is expected to be used by applications that have small values to propose. The maximum value size is defined by the p2p network configuration. For example, if libp2p gossipsub is used, this will be the `max_transmit_size` configured for the gossipsub protocol.
+
+To reduce the size of the vote messages it is recommended that `id(V)` method implementation should return a short representation of the value.
+
+```mermaid
+sequenceDiagram
+    box Proposer
+      participant A1 as Application
+      participant E1 as Consensus Engine
```
ancazamfir, line 130: Should this be 
```suggestion
      participant E1 as Engine
```

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,351 @@
...
+carry the proposed value `v`, but a short representation `id(v)` of it.
+A process cannot sign a vote for `id(v)` if it does not know the value `v`.
+So receiving `PROPOSAL` message carrying the value `v` is a requirement for
+signing a vote for `id(v)`.
+
+If the round of consensus is successful, the value `v` carried by the round's
+`PROPOSAL` message is the value delivered to the application as the decision
+for that height of consensus.
+
+Malachite, in `ProposalOnly` mode, when the application returns a full value in
```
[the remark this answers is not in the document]
ancazamfir: What kind of link?

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,404 @@
...
+
+Malachite provides implementations for the consensus core, engine and networking. Applications can integrate with:
+
+- consensus core - see [ADR-004](./adr-004-coroutine-effect-system.md)
+- engine and core integration - farcaster?
+- engine, core, networking integration - channel and starknet test examples?
+
+### Value Payload Modes
+
+At the moment, Malachite consensus implementation is supporting three
```
cason, line 88:
```suggestion
At the moment, Malachite supports three
```

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
cason: To the section describing this mode.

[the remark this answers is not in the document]
cason:
```suggestion
Malachite, in [`ProposalOnly` mode](#proposalonly), when the application returns a full value in 
```

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,351 @@
...
+actually proposed values, `v` is expected to be a short representation of `V`.
+For this reason, the optimization of having vote messages carrying `id(v)`
+instead of `v` becomes pretty much irrelevant.
+
+If the round of consensus is successful, the reference `v` carried by the
+round's `PROPOSAL` message is the value delivered by the consensus
+implementation as the decision for that height.
+But the actual decision value for that height of consensus is `V`, the value
+referenced by `v`.
+It is `V` and not `v` that should be delivered to the application.
```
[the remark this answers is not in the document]
cason: The application has to implement `Restream`, has to process proposal parts, so it is actually responsible for the value propagation.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,404 @@
...
+flexibility in the value propagation stage to enable more optimized 
+implementations. 
+
+In this document, we focus on the core parts of the consensus 
+implementation responsible for **Value Propagation**.
+
+TODO: Not sure where to mention this but seems relevant. There have been some slight changes to the Tendermint consensus:
+- `getValue()` is asynchronous and the propose timeout is passed as a parameter in order to let the builder know how long it has to build a value.
+- `valid(v)` is checked when a `PROPOSAL` is available but before the algorithm runs. This is equivalent in a way with `PROPOSAL(h, r, v, vr, valid)` where valid = {Valid | Invalid}.
+
```
cason, line 61:
```suggestion
- `upon ⟨PROPOSAL, h, r, v, vr⟩ from proposer(h, r)` represents the receiving of a `PROPOSAL` message for a round `r` of height `h` for vale `v`. This is a necessary condition for the success of round `v` with the decision of the proposed value `v`.
```

cason: Maybe as a second item this one.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,404 @@
...
+constant size. As a result, Malachite’s low-level API provides greater 
+flexibility in the value propagation stage to enable more optimized 
+implementations. 
+
+In this document, we focus on the core parts of the consensus 
+implementation responsible for **Value Propagation**.
+
+TODO: Not sure where to mention this but seems relevant. There have been some slight changes to the Tendermint consensus:
+- `getValue()` is asynchronous and the propose timeout is passed as a parameter in order to let the builder know how long it has to build a value.
+- `valid(v)` is checked when a `PROPOSAL` is available but before the algorithm runs. This is equivalent in a way with `PROPOSAL(h, r, v, vr, valid)` where valid = {Valid | Invalid}.
```
cason, line 60: If we want to refer to the pseudo-code and algorithm, this is actually a callback from consensus to the application upon the reception of a `PROPOSAL`, see the next bullet point I've suggested.

`valid(v) = true` is a requirement for the success of the round as well.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,351 @@
...
+consensus implementation.
+
+Since the values that are proposed and decided by consensus are references to
+actually proposed values, `v` is expected to be a short representation of `V`.
+For this reason, the optimization of having vote messages carrying `id(v)`
+instead of `v` becomes pretty much irrelevant.
+
+If the round of consensus is successful, the reference `v` carried by the
+round's `PROPOSAL` message is the value delivered by the consensus
+implementation as the decision for that height.
```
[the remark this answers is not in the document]
cason: So we always assume that the full value is stored by the application, in a kind of map `id(v) -> v`, possibly with height, round, etc.?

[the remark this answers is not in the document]
ancazamfir: currently yes. But not sure once we finish the `ProposalOnly` case.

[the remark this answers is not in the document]
ancazamfir: ...and it's `map id(v) -> v -> V` or `map id(v) -> V`

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
ancazamfir: agreed but still don't understand why we say "It is `V` and not `v` that should be delivered to the application." 
iiuc, this says consensus core should deliver `V` to the application which is what I challenge.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,404 @@
...
+  - providing the value to be propagated
+  - validating received values
+  - providing data availability for values (both undecided and decided)
+- Consensus Engine - a component that is responsible for managing the interactions between the consensus core and the application, and between the consensus core and the networking layer. For the ones involved in the value propagation process, the consensus engine is responsible for:
+  - propagating values and/or proposals via the networking layer
+  - receiving values and/or proposals from the networking layer
+  - relaying values and/or proposals to the consensus core
+- Consensus Core - this is the component that implements the tendermint consensus protocol.
+- Networking - this is the component that is responsible for transmitting messages between nodes.
+(Currently not shown in the diagrams)
```
ancazamfir, line 78: Should I add back the `Networking` block?

docs/architecture/adr-003-values-propagation.md
ancazamfir: I have a [PR ](https://github.com/informalsystems/malachite/pull/942)for the first as it was easier to fix the code instead vs figuring out the mermaid arrow. (cc @romac)

docs/architecture/adr-003-values-propagation.md
romac: It's fine either way imho, we can leave it like that if you prefer.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -58,7 +58,7 @@ implementation responsible for **Value Propagation**.
 TODO: Not sure where to mention this but seems relevant. There have been some slight changes to the Tendermint consensus:
```
romac, line 58: Remaining TODO

romac (approved):
There's a TODO remaining that needs to either be addressed or removed, looks great otherwise!

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,408 @@
...
+
+    /// The messages required to deliver proposals
+    pub value_payload: ValuePayload,
+
+    ...
+}
+
+```
+
+In the following, we examine each approach to understand how the consensus
```
nenadmilosevic95, line 112: I think that since we introduced the notion of consensus core and consensus engine, we need to be specific each time when we mention consensus

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,408 @@
...
+```
+
+In this mode, the consensus core generates the `GetValue()` effect to request a value from the application. The application responds by sending the `Propose(LocallyProposedValue(V))` input to the consensus core.
+
+The consensus core then generates a `Proposal(SignedProposal(V))` message that includes the actual
+value `V` received via `LocallyProposedValue(V)`, and generates a `Publish`
+effect. This effect is handled by the consensus engine, and the proposal,
+with the full value embedded in it, is disseminated through the network.
+
+Upon receiving a `Proposal(SignedProposal(V))` message from the network, the engine  passes it directly to the consensus core for processing.
```
nenadmilosevic95, line 189:
```suggestion
Upon receiving a `Proposal(SignedProposal(V))` message from the network, the engine passes it directly to the consensus core for processing.
```

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,408 @@
...
+Consensus core verifies the proposal is properly signed by the Proposer for the current height and round.
+
+_(Implementation in progress) The consensus engine (or core?) will also pass the unsigned `Proposal(V)` message to the application for validation. Once validation is performed the application generates the `ProposedValue(V, validity)` input and pass it to the consensus core._
+
+In this setup, the application only needs to provide a value to the consensus core through `Propose(LocallyProposedValue(V))`, and value propagation is entirely handled by the networking module. The consensus core only processes proposal messages that already contain the full value.
+
+This setup is the simplest, as the application is only responsible for providing the value to be ordered. However, if the value is large, the resulting proposal messages will also be large and must be propagated as such through the network. Any optimizations for value propagation, such as chunking the value into smaller parts and reassembling it on the receiving side, must be implemented at the networking level. This is because both the consensus and application remain unaware of how the value is transmitted.
```
nenadmilosevic95, line 196: In the current design descriptions, this can be done also at the Consensus Engine level?

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,408 @@
...
+not the full value. Nevertheless, in this mode, the consensus core generates the `Publish` effect and the `Proposal(SignedProposal(v))` message is disseminated through the network via networking module.
+
+When a node receives the `Proposal(SignedProposal(v))` message it passes it to the consensus. However, in this mode, receiving a proposal alone is not sufficient for the consensus core to consider the proposal complete. The consensus core must also receive confirmation from the application that the corresponding full value is available and was validated.
+
+This confirmation is delivered via the `ProposedValue(ProposedValue(v, validity))` input, which the application generates upon receiving the full value through the network. Only when the consensus core has both `Proposal(SignedProposal(v))` and `ProposedValue(ProposedValue(v, validity))` inputs, it can consider the proposal complete.
+
+### PartsOnly
+
+The same requirements as for `ProposalAndParts` apply. The only difference is that the application must also include `valid_round` in the dissemination protocol so that consensus core can distinguish between L22 and L28 in the consensus algorithm.
```
nenadmilosevic95, line 251: How the application learn what is the `validRound`?

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,408 @@
...
+
+In addition the application doesn't need to define a `Proposal` protocol message as one is not used. The reason is that all the information that would be carried in such a message can be derived from `V` and its metadata.
```
nenadmilosevic95, line 253: You mean the consensus doesn't need to define a `Proposal` message? Because this is the responsability of the consensus core not the app?

docs/architecture/adr-003-values-propagation.md
nenadmilosevic95: On the receive side is only to handle the case when some Byzantine process send the `Proposal` anyway? 
And in the end where do we drop it on the sending side?

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,408 @@
...
+    E2->>C2: ProposedValue(ProposedValue(v, validity))
+
+    Note over C2: Has v and its validity → can proceed
+```
+
+This mode is very similar to `ProposalandParts` but the difference is that when receiving
+`Propose(LocallyProposedValue(v))` , and after processed by the consensus core state machine,
+the `Publish` effect is not emitted and a proposal message is not sent through the network.
+
+At the receiving side, consensus core waits to receive `ProposedValue(ProposedValue(v, validity))` input and when this happens it considers the proposal as complete and proceeds. The application generates this input upon receiving the full value `V` from the network. As a result, in this case value propagation is totally delegated to the application.
```
nenadmilosevic95, line 291: Should we say how the signatures are handled here? I guess the app must also sign somehow V and then check also that V is signed by the proposer?

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,351 @@
...
+    participant N as Network
+
+    box Other nodes
+      participant C2 as Consensus
+      participant A2 as Application
+    end
+
+    C1->>A1: getValue
+    A1->>C1: Propose(LocallyProposedValue<Ctx>) (value ID)
+
```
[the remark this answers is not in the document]
ancazamfir: Added, also receiving our own proposal was missing

docs/architecture/adr-003-values-propagation.md
ancazamfir: ok, will leave it out for now

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
nenadmilosevic95: So in consensus by reference consensus core is working on `v`s, and will decide `v`, and we will have another layer that will ensure that corresponding `V` is available. But we can say that this layer will do the mapping and provide `V` to the app.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,408 @@
...
+round's `PROPOSAL` message is the value delivered by the consensus
+implementation as the decision for that height.
+But the actual decision value for that height of consensus is `V`, the value
+referenced by `v`.
+It is `V` and not `v` that should be delivered to the application.
+
+Notice, however, that a value can only be decided by the consensus
+implementation if a `PROPOSAL` message carrying that value was previously
+decided.
+As already mentioned, in this approach, a `PROPOSAL` message carrying a
```
nenadmilosevic95, line 378: Depending on the implementation, consensus does not need to have `V` in order to accept proposal with `v` it just need to have a guarantee that it will learn `V` eventually. In a sense, `ProposalAndParts` is kinda doing this but the guarantee is not ensured by some middle layer implementing Byzantine Reliable Broadcast, but rather just be trusting the application that it has a full value `V`. So if the application do not ensure this we can decide on `v` and we will not be able to extract `V` since the consensus core is unaware of `V`.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,408 @@
...
+
+    C1->>E1: Effect::GetValue()
+    E1->>A1: GetValue()
+    A1->>E1: Propose(LocallyProposedValue(V))
+    E1->>C1: Propose(LocallyProposedValue(V))
+    C1->>E1: Effect::Publish(SignedProposal(V))
+    E1->>E2: Proposal(SignedProposal(V))
+
+    E2->>C2: Proposal(SignedProposal(V))
+    E2-->>A2: Proposal(V)
```
nenadmilosevic95, line 175: I think this should be triggered by consensus core and not by the engine

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,408 @@
...
+the transmission of vote messages, which are relatively small and of
+constant size. As a result, Malachite’s low-level API provides greater
+flexibility in the value propagation stage to enable more optimized
+implementations.
+
+In this document, we focus on the core parts of the consensus
+implementation responsible for **Value Propagation**.
+
+TODO: Not sure where to mention this but seems relevant. There have been some slight changes to the Tendermint consensus:
+- `getValue()` is asynchronous and the propose timeout is passed as a parameter in order to let the builder know how long it has to build a value.
```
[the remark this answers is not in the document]
ancazamfir: yes and in this case timeout propose will fire.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,408 @@
...
+There are a number of entities that are involved in the value propagation process:
+
+- **Application:** The software that uses the consensus engine. It is responsible for:
+  - Providing the value to be propagated
+  - Validating received values
+  - Providing data availability for values (both undecided and decided)
+- **Consensus Engine:** The component responsible for managing the interactions between the consensus core and the application, and between the consensus core and the networking layer. For the ones involved in the value propagation process, the consensus engine is responsible for:
+  - Propagating values and/or proposals via the networking layer
+  - Receiving values and/or proposals from the networking layer
+  - Relaying values and/or proposals to the consensus core
```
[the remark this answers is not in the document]
ancazamfir: yes, i thought I added that. On the other hand if we get #943 in before we merge this then it's not anymore.

docs/architecture/adr-003-values-propagation.md
ancazamfir: Not in our Consensus Engine impl. There may be core integrators who use their own p2p network and engine equivalent component and they are free to design the way they want.

docs/architecture/adr-003-values-propagation.md
ancazamfir: It has to be taken care by the dissemination protocol. For starknet and farcaster there is a proposal part `Init` that includes `validRound`.

docs/architecture/adr-003-values-propagation.md
ancazamfir: I mean we don't need a wire `Proposal` specification (protobuf). In the core we still emulate a proposal but it's just an internal datastructure.

docs/architecture/adr-003-values-propagation.md
ancazamfir: Yes, let's see how we change the design. At this point we are aligning with current design, and it's in theory ony, as this hasn't been implemented yet. If we move it to the core, we want to do the same in all other cases.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+`Propose(LocallyProposedValue(v))` , and after processed by the consensus core state machine,
+the `Publish` effect is not emitted and a proposal message is not sent through the network.
+
+At the receiving side, consensus core waits to receive `ProposedValue(ProposedValue(v, validity))` input and when this happens it considers the proposal as complete and proceeds. The application generates this input upon receiving the full value `V` from the network. As a result, in this case value propagation is totally delegated to the application.
+
+### Summary
+To sum up, in different modes, different inputs are required to achieve the same effect as receiving
+the `PROPOSAL` message in the original Tendermint algorithm.
+
+- In `ProposalOnly` and `ProposalAndParts`, both `Proposal(SignedProposal(x))` and `ProposedValue(ProposedValue(x, validity))` inputs are needed, with `x == V` for the former and `x == v` for the latter.
```
cason, line 297: Isn't the opposite?

`SingedProposal(v)` and `ProposedValue(V)`?

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
+# ADR 003: Propagation of Proposed Values
+
+## Changelog
+
+* 2025-03-18: Context and description of the problem
+* 2025-03-21: Current design description
+
```
cason, line 7: Adding a new entry here would be good.

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
ancazamfir: Do you want to add some text to clarify this?

docs/architecture/adr-003-values-propagation.md
ancazamfir: > On the receive side is only to handle the case when some Byzantine process send the `Proposal` anyway? 

Yes
>And in the end where do we drop it on the sending side?

We drop it in the core currently. We have a PR up to drop it in the engine, once that goes in we will update the ADR.

docs/architecture/adr-003-values-propagation.md
ancazamfir: we say at beginning of `ProposalAndParts` section and then in this section we say same requirements apply. I can cut/paste

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
nenadmilosevic95: Will it also fire on the proposer node? Because if you see the algorithm this is not the case?

docs/architecture/adr-003-values-propagation.md
nenadmilosevic95: I was targeting this part of the sentence "Any optimizations for value propagation, such as chunking the value into smaller parts and reassembling it on the receiving side, must be implemented at the networking level. " ?

docs/architecture/adr-003-values-propagation.md
nenadmilosevic95: Ok I think it is important to note this here because here we are not sending the signed proposal message

docs/architecture/adr-003-values-propagation.md
nenadmilosevic95: I understood that this is not implemented so we can put what we like more here :D

docs/architecture/adr-003-values-propagation.md
nenadmilosevic95: I think this is fine as it is, for `ProposalOnly` we need `Proposal(SignedProposal(V))` and `ProposedValue(ProposedValue(V))` and for `ProposalAndParts` we need `Proposal(SignedProposal(v))` and `ProposedValue(ProposedValue(v))`

docs/architecture/adr-003-values-propagation.md
adizere: I will take this into account in my review. Good point!

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
adizere: I will make a note that this may change pending #943

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
ancazamfir: yes, we start the timer for proposer node also. The changes are also described in https://github.com/informalsystems/malachite/blob/main/specs/consensus/design.md#asynchronous-getvalue-and-proposevaluev. We could add a pointer here.

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
nenadmilosevic95: In the end I made a pass on the whole `Value Propagation Considerations` section. Please take a look!

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
nenadmilosevic95: This is good, because otherwise we can have a halt if app does not respond! Good job!

docs/architecture/adr-003-values-propagation.md
ancazamfir: I kind of liked it the way it was. But just changed it :)

docs/architecture/adr-003-values-propagation.md
ancazamfir: Ok, should we say "must be implemented outside consensus core"?

docs/architecture/adr-003-values-propagation.md
ancazamfir: Can you check the latest version?

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
ancazamfir: I added references to the design spec.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+There are a number of entities that are involved in the value propagation process:
+
+* **Application:** The software that uses the consensus engine. It is responsible for:
+  * Providing the value to be propagated
+  * Validating received values
+  * Providing data availability for values (both undecided and decided)
+* **Consensus Engine:** The component responsible for managing the interactions between the consensus core and the application, and between the consensus core and the networking layer. For the ones involved in the value propagation process, the consensus engine is responsible for:
+  * Propagating values and/or proposals via the networking layer
+  * Receiving values and/or proposals from the networking layer
+  * Relaying values and/or proposals to the consensus core (pending redesign in #943)
```
cason, line 80: This issue does not exist.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+* **Consensus Core:** The component in Malachite that implements the Tendermint consensus protocol.
+* **Networking:** The component responsible for transmitting messages between nodes (currently not shown in the diagrams).
```
cason, line 82:
```suggestion
* **Networking:** The component responsible for transmitting messages between nodes (not shown in the diagrams below).
```

docs/architecture/adr-003-values-propagation.md
nenadmilosevic95: Yes that should work

docs/architecture/adr-003-values-propagation.md
nenadmilosevic95: I think this is addressed in the new version I pushed

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+* Consensus core - See [ADR-004][adr-004]
+* Engine without networking - [Snapchain](https://github.com/farcasterxyz/snapchain)
+* Engine with networking (channel-based) - [Example application](https://github.com/informalsystems/malachite/tree/main/code/examples/channel)
+* Engine with networking (actor-based) - [Starknet test application](https://github.com/informalsystems/malachite/tree/main/code/crates/starknet)
+
+#### Changes to Tendermint
+
+Malachite deviates from the vanilla Tendermint consensus in these two key methods:
+
+* `getValue()` is asynchronous and the propose timeout is passed as a parameter in order to let the value (`V`) builder know how long it has to create it. See [Malachite Async GetValue][getvalue-changes] for more details.
```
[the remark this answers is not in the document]
cason: We haven't introduced `V` yet.

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
ancazamfir: I think we can solve this as @nenadmilosevic95 has made changes to this section to clarify the expectations for the app.

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
ancazamfir: Looks good. Just wanted to say that even when `Proposal` carries `v` we don't make the assumption that `v == id(v)`. It's again the app that decides that in the concrete implementation of `Value`. But I think we can leave the section as is. Will resolve this.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+Consensus core verifies the proposal is properly signed by the Proposer for the current height and round.
+
+*(Implementation in progress) The consensus engine (or core?) will also pass the unsigned `Proposal(V)` message to the application for validation. Once validation is performed the application generates the `ProposedValue(V, validity)` input and pass it to the consensus core.*
+
+In this setup, the application only needs to provide a value to the consensus core through `Propose(LocallyProposedValue(V))`, and value propagation is entirely handled by the networking module. The consensus core only processes proposal messages that already contain the full value.
+
+This setup is the simplest, as the application is only responsible for providing the value to be ordered. However, if the value is large, the resulting proposal messages will also be large and must be propagated as such through the network. Any optimizations for value propagation, such as chunking the value into smaller parts and reassembling it on the receiving side, must be implemented at the networking level. This is because both the consensus and application remain unaware of how the value is transmitted.
```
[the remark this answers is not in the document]
cason: So, what is not clear to me is: can a user provide its own implementation of the network component?

docs/architecture/adr-003-values-propagation.md
ancazamfir: removed as we decided to not wait for the PRs. please check 90a25d6f0f3f9bd7ab3a20c235158291fb91d1c0.

docs/architecture/adr-003-values-propagation.md
ancazamfir: Suggestion? Past two day work was still on "Current design description"  😄

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+
+This confirmation is delivered via the `ProposedValue(ProposedValue(v, validity))` input, which the application generates upon receiving the full value through the network. Only when the consensus core has both `Proposal(SignedProposal(v))` and `ProposedValue(ProposedValue(v, validity))` inputs, it can consider the proposal complete.
+
+### PartsOnly
+
+In this mode of operation, the application is responsible to define and implement the dissemination protocol for the full value `V` and its metadata (`height`, `round` and possibly `valid_round`). It is expected that the application splits the value `V` into parts, signs each part individually, and disseminates them throughout the network. At the receiving end, the application should verify that the parts are properly signed by the Proposer for the `height` and `round` as derived from the parts and reassemble the full value `V`.
```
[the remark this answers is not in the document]
cason: This is said in the following, btw.

docs/architecture/adr-003-values-propagation.md
cason: But is not the application that defines it. It is always the consensus core.

cason (commented):
Good, lets merge this?

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+
+The other two modes of operation are designed to support such optimizations at the application level rather than at the network level. We will explore how this is achieved in the following sections.
+
+### ProposalAndParts
+
+In this mode of operation, the application is responsible to define and implement the dissemination protocol for the full value `V` and its metadata (`height`, `round` and possibly `valid_round`). It is expected that the application splits the value `V` into parts, signs each part individually, and disseminates them throughout the network. At the receiving end, the application should verify that the parts are properly signed by the Proposer for the `height` and `round` as derived from the parts and reassemble the full value `V`.
```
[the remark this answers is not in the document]
ancazamfir: Parts don't go to the core

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+The application communicates to consensus that a value is available using a reference `v` to `V`. `v` is expected to be short representation of `V` and a possible (but not mandatory) implementation for `id` is `id(v) = v`.
+
+In order to handle restarts the application should persist the undecided values `V`.
```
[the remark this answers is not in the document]
ancazamfir: No.

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
ancazamfir: yeah, cut/paste error. will fix

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+the `Publish` effect is not emitted and a proposal message is not sent through the network.
+
+At the receiving side, consensus core waits to receive `ProposedValue(ProposedValue(v, validity))` input and when this happens it considers the proposal as complete and proceeds. The application generates this input upon receiving the full value `V` from the network. As a result, in this case value propagation is totally delegated to the application.
+
+### Summary
+
+To sum up, in different modes, different inputs are required to achieve the same effect as receiving
+the `PROPOSAL` message in the original Tendermint algorithm.
+
+* In `ProposalOnly` and `ProposalAndParts`, both `Proposal(SignedProposal(x))` and `ProposedValue(ProposedValue(x, validity))` inputs are needed, with `x == V` for the former and `x == v` for the latter.
```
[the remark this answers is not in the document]
ancazamfir: For `ProposalAndParts` the app provides `ProposedValue(ProposedValue(v, valid(V)))`.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+interacts with the environment, depending on the mode of operation adopted.
+Specifically, we focus on the core consensus inputs related to value propagation.
+In general consensus inputs are how the core reacts to the events from
+the environment. A complete overview of all inputs processes by consensus core
+can be found in [ADR-004 Coroutine-Based Effect System for Consensus][adr-004].
+
+| **Input** | **Fields** | **Description** |
+|-----------|------------|-----------------|
+| `Proposal (SignedProposal<Ctx>)` | `height`, `round`, `value`, `value_origin` | This input is generated and passed to consensus core when a proposal message is received from the network. |
+| `Propose (LocallyProposedValue<Ctx>)` | `height`, `round`, `value` | This input is produced by the application in response to a request (`getValue()` specifically) for a value to propose. |
```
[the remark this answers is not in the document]
ancazamfir: It's always `-1` as this is only as response to `GetValue()`.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+    C2->>C2: Proposal(SignedProposal(v))
+
+    Note over C2: Has v and its validity → can proceed
+```
+
+This mode is very similar to `ProposalandParts` but the difference is that when receiving
+`Propose(LocallyProposedValue(v))` , and after processed by the consensus core state machine,
+the `Publish` effect is not emitted and a proposal message is not sent through the network.
+
+At the receiving side, consensus core waits to receive `ProposedValue(ProposedValue(v, validity))` input and when this happens it considers the proposal as complete and proceeds. The application generates this input upon receiving the full value `V` from the network. As a result, in this case value propagation is totally delegated to the application.
```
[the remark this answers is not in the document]
ancazamfir: It's the core that has the "complete proposal" logic.

docs/architecture/adr-003-values-propagation.md
```diff
@@ -0,0 +1,416 @@
...
+### Summary
+
+To sum up, in different modes, different inputs are required to achieve the same effect as receiving
+the `PROPOSAL` message in the original Tendermint algorithm.
+
+* In `ProposalOnly` and `ProposalAndParts`, both `Proposal(SignedProposal(x))` and `ProposedValue(ProposedValue(x, validity))` inputs are needed, with `x == V` for the former and `x == v` for the latter.
+* In `PartsOnly`, only `ProposedValue(ProposedValue(v, validity))` input is enough, as no explicit proposal message is sent over the network.
+
+Regardless of the mode of operation, the value that consensus operates at the proposal level is defined by the application in the `Value` trait concrete implementation.
+The mode of operation is used outside the consensus driver and state machine in order to decide whether to send or accept explicit `Proposal` messages to the caller (via `Effect`).
```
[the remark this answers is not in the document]
ancazamfir: We still have mode check in the core.

romac (approved):

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
cason: So here we need the `ReStream` effect?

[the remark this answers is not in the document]
ancazamfir: Yes, `GetValue` is for L18, `Restream` is for L16. we should probably add diagrams or section for restreaming as the flows are different.

docs/architecture/adr-003-values-propagation.md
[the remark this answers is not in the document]
ancazamfir: Yes, also for the engine

docs/architecture/adr-003-values-propagation.md
```diff
@@ -324,6 +324,25 @@ the `Publish` effect is not emitted and a `Proposal` message is not sent through
...
+The restreaming flow follows a similar pattern to the initial proposal flow, but with these key differences:
+
+1. **Effect Generation**:
+   - Instead of `GetValue`, the consensus core generates a `RestreamProposal` effect
+   - This signals to the application that it should re-disseminate an existing value rather than generate a new one
+
+2. **Proposer Behavior**:
+   - The consensus core already has the valid value from the previous round
+   - Unlike the initial proposal flow, it does not need to wait for a `LocallyProposedValue` input
+   - The application only needs to handle re-dissemination of the value parts through the network
```
nenadmilosevic95, line 340: Why is this in **Proposer Behavior** section?

docs/architecture/adr-003-values-propagation.md
```diff
@@ -324,6 +324,25 @@ the `Publish` effect is not emitted and a `Proposal` message is not sent through
...
 At the receiving side, consensus core waits to receive `ProposedValue(ProposedValue(v, validity))` input and when this happens it considers the proposal as complete and proceeds. The application generates this input upon receiving the full value `V` from the network. As a result, in this case value propagation is totally delegated to the application.
 
+### Value Restreaming
+
+When operating in either `ProposalAndParts` or `PartsOnly` modes, there are cases where a proposer needs to re-propose a value that was previously seen in an earlier round. This occurs when implementing L16 of the Tendermint algorithm, where a proposer enters a new round with a valid value from a previous round.
+
+The restreaming flow follows a similar pattern to the initial proposal flow, but with these key differences:
+
+1. **Effect Generation**:
+   - Instead of `GetValue`, the consensus core generates a `RestreamProposal` effect
```
nenadmilosevic95, line 334: From application side, the only difference is this? Namely, that instead of `getValue` it will process `RestreamProposal` and re-stream the same proposal?

docs/architecture/adr-003-values-propagation.md
```diff
@@ -324,6 +324,25 @@ the `Publish` effect is not emitted and a `Proposal` message is not sent through
...
+   - This signals to the application that it should re-disseminate an existing value rather than generate a new one
+
+2. **Proposer Behavior**:
+   - The consensus core already has the valid value from the previous round
+   - Unlike the initial proposal flow, it does not need to wait for a `LocallyProposedValue` input
+   - The application only needs to handle re-dissemination of the value parts through the network
+
+3. **Non-Proposer Behavior**:
+   - If a non-proposer node has already validated this value in a previous round, it can skip waiting for the `ProposedValue` input
```
nenadmilosevic95, line 343: If you don't wait for this input in `PartsOnly` mode how do you start when you have proposal and when you should send the prevote?

nenadmilosevic95 (commented):
I made a pass on old document and it looks good for me. The only thing I am not sure about is if we should use `V` notation somehow in `Consensus By Value` section. 

For the new section I made some comments.

docs/architecture/adr-003-values-propagation.md
ancazamfir: yes

docs/architecture/adr-003-values-propagation.md
ancazamfir: Because it applies to the proposer node side

docs/architecture/adr-003-values-propagation.md
ancazamfir: ah good catch, this is for `ProposalAndParts` mode. In `PartsOnly` the `Proposal` can only be synthesized when we get `ProposedValue`.

docs/architecture/adr-003-values-propagation.md
ancazamfir: 7be2b78cb9a213097c77103c74aae7271da3b430

docs/architecture/adr-003-values-propagation.md
ancazamfir: c33540f6d53205d987c3d07d9c557510b05f39df

ancazamfir:
> The only thing I am not sure about is if we should use `V` notation somehow in Consensus By Value section.

yes, it's not perfect but I am ok with current version. @cason wdyt?

docs/architecture/adr-003-values-propagation.md
cason: Yes... but the application needs to design a protocol message/dissemination that includes everything which is inside a `Proposal`,  so I don't see advantages here.

The application does not need to care about the `Proposal` in the previous two modes.

ancazamfir (approved):

PR #898, opened 2025-03-07
chore(test): Replay WAL before start round

ancazamfir:
Closes: #XXX

An experiment for simply moving the WAL replay before start round. If we enter new round with a proposal already we should not `getValue()` again.
We should also find all the votes already stored for all rounds, etc. ~~NewRound mux would need more updates though.~~
edit: it's actually `step_change` mux and it looks pretty complete. 

---

#### For all contributors

- [ ] Reference GitHub issue


ancazamfir:
So it look like one thing missing in step change mux, specifically for this case `Unstarted` -> `Propose`,  is looking for conditions at higher rounds (L49 and L55). However this is an optimization that we don't do today (as we replay one by one). We should add multiple round wal tests.

romac:
@ancazamfir I added a new `PrepareHeight` input to avoid mutating the driver state from outside of the `core-consensus` crate, what do you think?

romac:
As discussed today, let's close this in favor of https://github.com/informalsystems/malachite/pull/896

ancazamfir:
@romac just to clarify, we will still replay (proposals) at start height before start round, meaning most of the changes here should still apply, right?

PR #902, opened 2025-03-10
fix(code/core-consensus): Only persist votes in the WAL if they have not yet been seen

romac:
Before adding a vote in the WAL, check that we have not yet seen it, ie. that it has not yet been tallied in the vote keeper.

This avoids the situation where a node is stuck, repeatedly asks for votes via vote sync but cannot make progress, and ends up with a lot of duplicate votes in the WAL.

---

#### For all contributors

- [ ] Reference GitHub issue

### DIFF
--- a/code/crates/core-consensus/src/handle/vote.rs
+++ b/code/crates/core-consensus/src/handle/vote.rs
@@ -75,8 +75,9 @@ where
 
     debug_assert_eq!(consensus_height, vote_height);
 
-    // Only append to WAL and store precommits if we're in the validator set
-    if state.is_validator() {
+    // Only append to WAL and store precommits if we're in the validator set,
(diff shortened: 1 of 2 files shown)


cason:
Generally speaking, we should only add valid/correct votes to the WAL, although this is an optimization.

I wonder, as an option, whether the WAL appending task could not be performed after the vote keeper has processed the vote, and based on its output. Meaning: if the vote keeper has stored the vote, it is valid, therefore we should persist it _before_ processing the vote keeper output. But probably this complicates things a lot.

romac:
I think that's doable without introducing too much complexity by splitting vote validation and processing. Tracked in https://github.com/informalsystems/malachite/issues/918

PR #904, opened 2025-03-10
fix(code/app): Add the POL round to `ProposalInit` message of the example and test applications

romac:
Closes: #XXX

---

#### For all contributors

- [ ] Reference GitHub issue

### DIFF
--- a/code/crates/app-channel/src/msgs.rs
+++ b/code/crates/app-channel/src/msgs.rs
@@ -58,9 +58,9 @@ pub enum AppMsg<Ctx: Context> {
     /// The application MUST reply to this message with the requested value
     /// within the specified timeout duration.
     GetValue {
-        /// Height which consensus is at
+        /// Height for which the value is requested
         height: Ctx::Height,
-        /// Round which consensus is at
+        /// Round for which the value is requested
         round: Round,
         /// Maximum time allowed for the application to respond
         timeout: Duration,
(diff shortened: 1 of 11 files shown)


romac:
@ancazamfir Not sure what is the best name here, in some places we use _valid round_ and in some others we use _POL round_, we should decide on which is best and updates every use site accordingly.

ancazamfir:
> @ancazamfir Not sure what is the best name here, in some places we use _valid round_ and in some others we use _POL round_, we should decide on which is best and updates every use site accordingly.

maybe we can use `valid_round` in the app and we keep (for now) the `pol_round` in the consensus core only?

ancazamfir:
We need to take `valid_round` from `Init` and put it in `ProposedValue` here:
https://github.com/informalsystems/malachite/blob/530fed36e64aa7e4cc4671f764502b94a632afb1/code/crates/test/app/src/state.rs#L416

ancazamfir:
For testing we can try:
```
cargo run --bin informalsystems-malachitebft-example-channel -- testnet --nodes 3 --home nodes
export APP_BINARY="informalsystems-malachitebft-example-channel"; ./scripts/spawn.bash --nodes 3 --home nodes
```
This works with `proposal-and-parts` even with the byzantine proposer, just slow at `height=1` where it will trigger the restreaming and non-nil valid_round. So we should try the same with `parts-only`.

ancazamfir (approved):
🙌

Issue #897, opened 2025-03-07
code: WAL scenarios for Proposal


ancazamfir:
There are few cases to review/ fix:
- `parts-only`: There is no explicit proposal message being sent/ received on the wire. The application streams the full value (block). We don't write anything related to the proposal at either proposer or non-proposer. In this case, at restart the only indication that the app has sent a value is the presence of the vote. But this increases the window where consensus will do a second `getValue()` after restart. ~~We should consider storing the implicit Proposals to the WAL .~~ We should log to WAL the `ProposedValue` received from the application that derived the implicit `Proposal`
- `proposal-and-parts`: The application streams the full value and in addition, consensus sends explicit Proposal messages. We write the proposal to WAL with implication
   - at proposer: it means application has built a full value and we assume it is available after restart. This could be an indication to not call `getValue()` again. 
   - at non-proposer: it is possible that we have received the explicit Proposal message on the wire but the application did not have time to reassemble the full value before restart. Our assumption here is that the proposer may replay the `ProposedValue` message to consensus after restart. In this case consensus will call the driver with the (full) proposal.
 However the application replays the `ProposedValue` after the WAL replay and in some cases it will arrive too late for consensus state machine to restore its state (e.g. locked and valid values). Therefore we should store the `ProposedValue` in the WAL for this case as well.

Here is a brief diagram with current flow showing in blue where we currently save to WAL, or assume application will save values to handle restarts.

![Image](https://github.com/user-attachments/assets/18ad7c73-8350-4155-80eb-5d011cea7d33)

cason:
Thank you for that.

In any case, `Proposal` here contains only the value ID, right?

ancazamfir:
> Thank you for that.
> 
> In any case, `Proposal` here contains only the value ID, right?

`Proposal`, `ProposeValue` and `ProposedValue` contain `Value` as defined by the application in the consensus context. In all usecases we have so far, `Value` is some sort of Hash and `id(v) == v`, with `id(v)` being carried in the votes. 
We do in theory support full value in `Proposal` but we don't have yet an app to test this. In this case we will not have proposal parts on the wire (that's why the interrupted line in the diagram)

ancazamfir:
> > In all usecases we have so far, Value is some sort of Hash and id(v) == v, with id(v) being carried in the votes.
> 
[...]

PR #942, opened 2025-03-25
chore(code): Move mode check on effect handling in engine

ancazamfir:
Closes: #XXX

---

#### For all contributors

- [ ] Reference GitHub issue

### DIFF
--- a/code/crates/engine/src/consensus.rs
+++ b/code/crates/engine/src/consensus.rs
@@ -17,7 +17,7 @@ use malachitebft_core_consensus::{
 };
 use malachitebft_core_types::{
     Context, Round, SigningProvider, SigningProviderExt, Timeout, TimeoutKind, ValidatorSet,
-    ValueId, ValueOrigin,
+    ValueId, ValueOrigin, ValuePayload,
 };
 use malachitebft_metrics::Metrics;
 use malachitebft_sync::{
@@ -243,6 +243,7 @@ where
         input: ConsensusInput<Ctx>,
     ) -> Result<(), ConsensusError<Ctx>> {
         let height = state.height();
+        let value_payload = state.consensus.params.value_payload;
 
         malachitebft_core_consensus::process!(
             input: input,
@@ -255,6 +256,7 @@ where
                     &mut state.timers,
                     &mut state.timeouts,
                     state.phase,
+                    value_payload,
                     effect
                 ).await
             }
@@ -823,6 +825,7 @@ where
         timers: &mut Timers,
         timeouts: &mut Timeouts,
         phase: Phase,
+        value_payload: ValuePayload,
         effect: Effect<Ctx>,
     ) -> Result<Resume<Ctx>, ActorProcessingErr> {
         match effect {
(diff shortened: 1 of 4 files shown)


code/crates/engine/src/consensus.rs
```diff
@@ -243,6 +243,7 @@ where
         input: ConsensusInput<Ctx>,
     ) -> Result<(), ConsensusError<Ctx>> {
         let height = state.height();
+        let value_payload = state.consensus.params.value_payload;
```
romac, line 246: Do we still use `value_payload` in core-consensus? If not, we should remove it from the consensus params.

code/crates/engine/src/consensus.rs
```diff
@@ -951,12 +954,17 @@ where
                 // NOTE: The message has already been append to the WAL by the `WalAppendMessage` effect.
                 self.wal_flush(phase).await?;
 
-                // Notify any subscribers that we are about to publish a message
-                self.tx_event.send(|| Event::Published(msg.clone()));
+                let should_broadcast = match &msg {
+                    SignedConsensusMsg::Vote(_) => true,
+                    _ => value_payload.include_proposal(),
+                };
```
romac, line 960: Let's refactor this into a function that we can use here and for the WAL below

code/crates/engine/src/consensus.rs
romac: We still use it here:
- https://github.com/informalsystems/malachite/blob/anca%2Fengine_effect_handling/code/crates/core-consensus/src/handle/proposed_value.rs#L40
- https://github.com/informalsystems/malachite/blob/anca%2Fengine_effect_handling/code/crates/core-consensus/src/handle/proposal.rs#L78

ancazamfir: yes, we can look into changing that when we complete `ProposalOnly` implementation and maybe reorg based on latest discussions.

code/crates/engine/src/consensus.rs
ancazamfir: I think for the WAL we will have to remove the check when we fix #897. So we have three options
- refactor now and undo when we fix #897 
- leave things as they are
- fix #897 in this PR

code/crates/engine/src/consensus.rs
ancazamfir: For the second one I suggest we fix it in a different PR.

ancazamfir: moved signature verification in the engine, removed value_payload from the consensus parameters and put into the engine.

romac:
I think this approach puts too much burden on users integrating at the core-consensus level, and we should postpone this until we have a higher-level wrapper around the core-consensus API. What do you think @ancazamfir?

ancazamfir:
> I think this approach puts too much burden on users integrating at the core-consensus level, and we should postpone this until we have a higher-level wrapper around the core-consensus API. What do you think @ancazamfir?

We should clarify with the integrators if this is the case. Having less Effects to implement is good but it comes at a cost of integrators having to verify all sigs before sending messages to core. But I assume core API integrators will look at the engine and see what they need to do.

For me the bigger concern is that here we do the expensive checks early, for example before we verify that messages are for lower heights and discard. For that reason I would also move some of the checks in the engine...and then yes it's extra work on integration that need to be done and a wrapper would be good.

ancazamfir:
Let's close this until we have a way to verify signatures for future heights in the engine, or we decide to optimistically verify with current height validator set.

PR #971, opened 2025-04-09
feat(code/sync): Include both tip and sync height in sync status

romac:
Let's say a node is lagging behind by one height whereas other nodes are stuck at the next height (say because there is not enough online voting power to make progress):

- Lagging node: `tip = 5, sync = 5`
- Stuck nodes: `tip = 5, sync = 6`

In this case, we would expect that lagging node would be able to request a value from one of the stuck nodes. This is not the case right now, because the nodes only broadcast their tip height and not their sync height. Therefore we cannot differentiate between a node who is lagging at height 5 or stuck at height 6, and will not consider any of the stuck nodes for ValueSync.

Because of this, [the test](https://github.com/informalsystems/malachite/pull/952/files#diff-73448f07ace47cb4ffc84bd1fc246ff57ad4f34d516f25cd034429951b6b0b63) that I added in https://github.com/informalsystems/malachite/pull/952 fails because node that just reset its height to 5 will not consider syncing with any of the nodes who are stuck at height 6 and therefore gets stuck at height 5.

---

This PR adds the sync height to the status and changes the logic accordingly.

---

#### For all contributors

- [ ] Reference GitHub issue

### DIFF
--- a/code/crates/sync/src/handle.rs
+++ b/code/crates/sync/src/handle.rs
@@ -35,7 +35,12 @@ impl<Ctx: Context> Default for Resume<Ctx> {
 #[derive_where(Debug)]
 pub enum Effect<Ctx: Context> {
     /// Broadcast our status to our direct peers
-    BroadcastStatus(Ctx::Height),
+    BroadcastStatus {
+        // The height of the last value we decided on
+        tip_height: Ctx::Height,
+        // The height we are currently syncing at
+        sync_height: Ctx::Height,
+    },
 
     /// Send a ValueSync request to a peer
     SendValueRequest(PeerId, ValueRequest<Ctx>),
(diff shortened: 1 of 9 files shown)


code/crates/sync/src/handle.rs
```diff
@@ -174,22 +177,21 @@ pub async fn on_status<Ctx>(
...
-    state.update_status(status);
+    state.update_peer_status(status.clone());
 
     if !state.started {
         // Consensus has not started yet, no need to sync (yet).
         return Ok(());
     }
 
-    if peer_height > state.tip_height {
+    if status.sync_height > state.sync_height && status.sync_height > state.tip_height {
```
romac, line 189: TODO: Express this condition better and add comment explaining it.

romac:
Not the actual issue. We should instead update the tip on reset.

Issue #932, opened 2025-03-19
code/specs: Deliver decisions immediately while collecting late precommits and remove Commit timeout


romac:
Here is the way _timeoutCommit_ works in CometBFT:

1. At height H, the decision is delivered to the application immediately upon receiving 2f+1 precommits
2. The _timeoutCommit_ period is used to collect additional precommits while the application processes the decision
3. These additional precommits are included in block H+1
4. The next height starts after _timeoutCommit_ expires

Whereas in Malachite we wait until _timeoutCommit_ elapses before delivering the decision for height H to the application, including any additional precommits we've collected for that same height H.

As @cason suggested in the weekly call we could instead:

1. Deliver the decision for H to the application and start the next height H+1 immediately, rather than waiting for _timeoutCommit_ to elapse
2. Keep collecting additional precommits for H while we do consensus for H+1. We can then deliver these additional precommits for H in the decision for H+1
3. Remove _timeoutCommit_ altogether as will not serve any purpose anymore

> An important remark here that in CometBFT the `timeoutCommit` starts when a block is ready to be delivered to the application, namely when it gets `2f + 1` voting-power equivalent Precommits, and expires when the block is (hopefully) already processed by the application and the next height is started.
> 
> So, we indeed collect extra Precommits for an already decided block, but the block is delivered to the application once it is decided. The Precommits for block `H`, both the required `2f + 1` and the extra ones received during this timeout, are then included in the successive block `H+1`, not delivered to the application together with block `H`.
>
> For this reason, if processing a block takes `X` time, on average, and `timeoutCommit` is set to `X`, there is no actual delay in most cases. The point is that _while the application is processing the block_, the consensus layer collect extra Precommits.

> _Originally posted by @cason in https://github.com/informalsystems/malachite/discussions/930#discussioncomment-12541820_

cason:
In the specs we don't have a `timeoutCommit` and we decide a value once the minimal conditions for it are observed, namely: `2f + 1` voting-power equivalent Precommits for `id(v)` and a Proposal for `v` for the same height and round of the votes.

adizere:
> We can then deliver these additional precommits for H in the decision for H+1

This will involve changing the `decide` effect to provide multiple commit certificates, specifically two certificates, one for H and another for H+1. This will make  the API potentially confusing, because as an app builder, I would want to use the certificate for H before constructing block H+1 (not at the decision step of H+1).

I might be wrong, and some apps may benefit from the proposed API, however.

[...]

PR #976, opened 2025-04-11
feat(code): Removal of timeout commit

adizere:
Closes: #932

---

#### For all contributors

- [ ] Reference GitHub issue


code/crates/core-consensus/src/handle/driver.rs
```diff
@@ -308,10 +309,7 @@ where
                 "Decided",
             );
 
-            perform!(
-                co,
-                Effect::ScheduleTimeout(Timeout::commit(consensus_round), Default::default())
-            );
+            try_decide(co, state, metrics).await?;
```
adizere, line 312: This likely require more in-depth review that it's the right way to do things.

code/crates/engine/src/wal/entry.rs
```diff
@@ -120,7 +120,6 @@ fn encode_timeout(tag: u8, timeout: &Timeout, mut buf: impl Write) -> io::Result
         TimeoutKind::Propose => 1,
         TimeoutKind::Prevote => 2,
         TimeoutKind::Precommit => 3,
-        TimeoutKind::Commit => 4,
```
adizere, line 123: Note that with removing of `TimeoutKind::Commit`, we are now skipping the encoded value `4`. Not sure this is confusing, let me know if I should adjust to avoid skipping 4

romac: We should probably adjust the code to skip entries with tag `4` in order to maintain backward compatibility and preserve the ability to replay a WAL containing commit timeouts.

romac: Actually that's fine because we already just skip entries that fail to decode. But maybe we can yield a specific error to be explicit about commit timeouts having been removed.

romac: 4f1166b0e157e4e8e30a4c2f9674b58743d8e593

code/crates/core-consensus/src/handle/driver.rs
ancazamfir: This seems correct to me. The name `try_decide` looks strange. 
We also need to remove the other calls from the handlers in `sync.rs` and `proposed_value.rs`, the only reason we had them there was to skip the commit timer in case we were syncing.

code/crates/engine/src/wal/entry.rs
```diff
@@ -144,12 +144,22 @@ fn decode_timeout(mut buf: impl Read) -> io::Result<Timeout> {
         2 => TimeoutKind::Prevote,
         3 => TimeoutKind::Precommit,
 
+        // Commit timeouts have been removed in <https://github.com/informalsystems/malachite/pull/976>,
```
adizere, line 147: Nice catch Romain!

ancazamfir (approved):
lgtm!

PR #981, opened 2025-04-14
fix(spec/quint): Move consensus timeout firing into state machine

josef-widder:
In the current spec, a timeout is fired whenever a process takes a step and has nothing else to do (e.g., no incoming messages to process). I refactored this so that the timeouts are now fired as actions. This should eventually allow smoother simulation (e.g., we can define a step action where no timeouts are fired).

~~This is a draft PR, as I would like to check whether people find this useful. As the timeout firing is moved, several tests are failing right now. I fixed a few (simpler) tests. I am happy to fix the remaining ones if people are happy with the change. We can discuss here.~~

As @cason agreed that this modeling is more to the point (giving the ability to fire of not fire to the environment), I have fixed the remaining tests. I think the new tests (scenarios) are slightly better readable because they clearly separate between normal steps, and timeout steps. In general adding more assert statements to the run might improve maintainability, but I don't think this is a priority at the moment.

---

#### For all contributors


specs/consensus/quint/driver.qnt
```diff
@@ -653,9 +653,6 @@ module driver {
             val newstate = { ...state, incomingVotes: state.incomingVotes.exclude(Set(vote))}
             (newstate, VoteDInput(vote))
 
-        else if (state.existsTimeout())
```
cason, line 656: The comment at the begging of this method should be updated, as it mentions timeouts.

specs/consensus/quint/driver.qnt
```diff
@@ -710,9 +707,6 @@ module driver {
             val cert = certificatesForCurrentHeight.fold (Set(), (sum, y) => y)
             val newstate = { ...state, incomingCertificates: state.incomingCertificates.exclude(Set(cert))}
             (newstate, CertificateDInput(cert))
-            
-        else if (command == TimeoutCmd and state.existsTimeout())
```
cason, line 714: Not sure if we should remove this. The `nextActionCommand` is only used by a test unit (`tests/state-machine/statemachineTest.qnt`) and in the case the `TimeoutCmd` is explicitly provided.

specs/consensus/quint/statemachineAsync.qnt
```diff
@@ -179,6 +177,21 @@ action valStep(v: Address) : bool = {
...
+action valTimeout(v: Address) : bool = {
+    if (not(system.get(v).existsTimeout()))
+        unchangedAll
+    else
+        val next_action = system.get(v).pickTimeout()
+        doStep(v, next_action)
+}
+
+action valStep(v: Address) : bool = {
+    // pick action
```
cason, line 189:
```suggestion
    // Pick next action to execute, excluding timeouts
```

cason (commented):
Good, this allows more control on when timeouts are expired.

cason:
I've changed my mind regarding this approach, and I think it is the way to go.

Timeouts should not be the event that is processed (actually, produced, once the timeout is scheduled) once the Driver does not find a higher priority event to feed consensus with. This toggle from the state machine (actions) the control over when timeouts are triggered, that is something the belongs to the state machine, not to the Driver.

Thank you for fixing the broken tests, there still a few to go, maybe I can help with.

As an additional comment, in order to render this change more effective, we should consider a logic to cancel a timeout, what do you think? At the moment, timeouts are produced at certain conditions and added to a set. The timeout triggered is just one of them, picked deterministically. This includes, in longer executions, timeouts which do not have any effect (e.g., because we moved to a different round step or round). An alternative would be to trigger all pending timeouts, maybe?

specs/consensus/quint/tests/state-machine/parameterizedTest.qnt
```diff
@@ -47,7 +51,9 @@ run UnSuccessfulRound(prop: NonNilValue, r: Round): bool =
         N4F0::unchangedAll
     }
     .then(N4F0::deliverProposal(proposer, mkProposal(proposer, 0, r, prop, -1)))
-    .then(everyoneTakesAStep)
+    // .then(everyoneTakesAStep)
```
josef-widder, line 54:
```suggestion
```

specs/consensus/quint/tests/newheightlogic/heightLogicrun.qnt
```diff
@@ -154,7 +154,7 @@ run slowProcessAtHeight1 = {
 run catchUpToHigherRound = {
     init
     .then(fastProcessesDoNotDecide(0, 0, ValueList[0], ValueList[1]))
-    // deliver all prevotes of round 0 and 1
+    // // deliver all prevotes of round 0 and 1
```
josef-widder, line 157:
```suggestion
    // deliver all prevotes of round 0 and 1
```

nenadmilosevic95 (approved):
I had a sync call with @josef-widder and we believe this is ready for merge.

PR #983, opened 2025-04-14
fix(starknet/mempool): Fix behavior of mempool load generator

romac:
Closes: #XXX

---

#### For all contributors


romac:
When running `./scripts/spawn-tmux.bash --nodes 3 --home y --app informalsystems-malachitebft-starknet-app --debug` with these parameters, I get ~30ms block time for 1 MiB blocks:

```
export MALACHITE__CONSENSUS__P2P__PROTOCOL__TYPE="gossipsub"
export MALACHITE__CONSENSUS__TIMEOUT_PROPOSE="5s"
export MALACHITE__CONSENSUS__TIMEOUT_PREVOTE="3s"
export MALACHITE__CONSENSUS__TIMEOUT_PRECOMMIT="3s"
export MALACHITE__CONSENSUS__TIMEOUT_COMMIT="0s"
export MALACHITE__CONSENSUS__TIMEOUT_STEP="2s"
export MALACHITE__CONSENSUS__VOTE_SYNC__MODE="request-response"

export MALACHITE__SYNC__ENABLED=true
export MALACHITE__SYNC__REQUEST_TIMEOUT="30s"

export MALACHITE__MEMPOOL__MAX_TX_COUNT=10000
export MALACHITE__MEMPOOL__GOSSIP_BATCH_SIZE=0

export MALACHITE__TEST__MAX_BLOCK_SIZE="1 MiB"
export MALACHITE__TEST__TX_SIZE="1 KiB"
export MALACHITE__TEST__TXS_PER_PART=256
export MALACHITE__TEST__TIME_ALLOWANCE_FACTOR=0.5
export MALACHITE__TEST__EXEC_TIME_PER_TX="0ms"
export MALACHITE__TEST__MAX_RETAIN_BLOCKS=100
export MALACHITE__TEST__VOTE_EXTENSIONS__ENABLED=false
export MALACHITE__TEST__VOTE_EXTENSIONS__SIZE="1KiB"
```

code/crates/starknet/test/src/lib.rs
```diff
@@ -189,7 +189,10 @@ impl TestRunner {
                     .unwrap(),
             },
             runtime: RuntimeConfig::single_threaded(),
-            test: TestConfig::default(),
+            test: TestConfig {
+                stable_block_times: true,
```
ancazamfir, line 193: If this is set to false, which tests are failing and how? Is there a CI failure that shows this? I looked at the failures in 726b43472c958cae6bc55c5d4b54a88cf0ea0e5c but I think may have been because of the mempool bcast.
```suggestion
                stable_block_times: false,
```

romac: IIRC it's mostly the `::full_node` tests which are failing, the nodes apparently don't have time to catch up via sync. We can try to fix this, but I wanted to get this merged early to unblock @nenadmilosevic95.

ancazamfir: oh we should fix this but yes let's do it in a separate PR then.

code/scripts/spawn-tmux.bash
```diff
@@ -11,15 +11,18 @@ export MALACHITE__CONSENSUS__VOTE_SYNC__MODE="request-response"
 export MALACHITE__SYNC__ENABLED=true
 export MALACHITE__SYNC__REQUEST_TIMEOUT="30s"
 
-export MALACHITE__MEMPOOL__MAX_TX_COUNT=1000
+export MALACHITE__MEMPOOL__MAX_TX_COUNT=10000
 export MALACHITE__MEMPOOL__GOSSIP_BATCH_SIZE=0
+export MALACHITE__MEMPOOL__LOAD__LOAD_TYPE="uniform_load"
+export MALACHITE__MEMPOOL__LOAD__INTERVAL="1ms"
+# export MALACHITE__MEMPOOL__LOAD__COUNT=1000 # For some reason this fails to parse?
```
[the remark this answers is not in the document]
ancazamfir: I tried with a few other env vars and some give the same error (mempool's `mesh_n`), others don't but don't seem to take effect either (mempool's `num_outbound_peers`), others work (`max_tx_count`).

ancazamfir (approved):

code/crates/starknet/host/src/host/proposal.rs
```diff
@@ -157,15 +153,21 @@ async fn run_build_proposal_task(
...
+        } else if start.elapsed() >= build_duration {
+            debug!("Time allowance exceeded, stopping tx generation");
             break 'reap;
         }
     }
 
+    if params.stable_block_times {
+        // Sleep for the remaining time, in order to not break tests
+        // by producing blocks too quickly
+        tokio::time::sleep(build_deadline - Instant::now()).await;
```
nenadmilosevic95, line 168: By default we don't sleep here? what is `build_deadline`?

code/crates/starknet/host/src/mempool.rs
```diff
@@ -154,23 +151,48 @@ impl Mempool {
...
-            Msg::Input(tx) => {
-                if state.transactions.len() < self.max_tx_count {
-                    state.add_tx(tx);
-                } else {
-                    trace!("Mempool is full, dropping transaction");
+            Msg::AddBatch(batch) => {
+                trace!("Received batch of {} transactions", batch.len());
+
+                // If mempool gossip is enabled, broadcast the transactions to the network
+                if self.gossip_batch_size > 0 {
```
nenadmilosevic95, line 158: I would maybe rather have a flag `gossip_enabled` or something similar rather than using batch_size to check if we broadcast tx

code/crates/starknet/host/src/mempool_load.rs
```diff
@@ -175,11 +172,9 @@ impl Actor for MempoolLoad {
         match msg {
             Msg::GenerateTransactions { count, size } => {
                 let transactions = Self::generate_transactions(count, size);
-                let tx_batch = TransactionBatch::new(transactions).to_any().unwrap();
+                let tx_batch = TransactionBatch::new(transactions);
 
-                let mempool_batch: MempoolTransactionBatch = MempoolTransactionBatch::new(tx_batch);
-                self.network
-                    .cast(MempoolNetworkMsg::BroadcastMsg(mempool_batch))?;
+                self.mempool.cast(MempoolMsg::AddBatch(tx_batch))?;
```
nenadmilosevic95, line 177: Nice, now we always broadcast through the mempool?

code/scripts/spawn-tmux.bash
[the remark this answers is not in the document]
nenadmilosevic95: I had "similar" problems when running `qa` with `sconfig` command, I was not specifying the type properly. But I guess here it is different, and you maybe don't use `sconfig` here

nenadmilosevic95 (commented):
LGTM! I might need your help after I merge this into qa branch because I think that config is changed a bit compared to the one I have on my branch

code/crates/starknet/host/src/host/proposal.rs
romac: We don't sleep here by default so that we are fast, we only sleep when "stable block times" (perhaps best named "stable build times" are enabled) is enabled in order to not break some tests. We'll fix this in a follow-up PR, but it won't affect QA as long as it's set to false.

code/crates/starknet/host/src/mempool.rs
romac: We could this but we wanted to avoid introducing one more setting as we have too many already, and this setting contains all the information needed. I would suggest we roll with this for now. Also having both means that some configurations do not make sense, like gossip_enabled = true with gossip_batch_size = zero.

code/crates/starknet/host/src/mempool_load.rs
romac: Yes at the time where we get a batch. But I just realized that we might be re-gossiping batches we get from the network, which is wasteful since GossipSub should already be doing the re-broadcast for us. I will fix this.

Issue #944, opened 2025-03-27
code: Dynamic validator set


romac:
- [x] Add a `get_validator_set(&self, Height, Genesis) -> ValidatorSet` hook to the `Middleware`
- [x] Define a `RotateValidators` middleware that deterministically picks a subset of the genesis validators every N heights.
  - [x] Use it in a happy path integration test (eg. 5 nodes, pick 3 validators out of 5 every 2 heights, run to height 20)
  - [x] Run the `value_sync::start_late` test with the `RotateValidators` middleware
  - [x] Find other existing integrations tests that would be interesting to run with that middleware
- [x] Define an `EpochValidators` middleware that only knows about the validator set up to N heights in the future
  - [x] Find existing integrations tests that would be interesting to run that middleware
- [x] Update the example channel-based app to use a rotating (+ epoch-based?) validator set

PR #987, opened 2025-04-15
feat(code/test): Dynamic validator set

bastienfaivre:
Closes: #944 

---

#### For all contributors


code/crates/test/app/src/app.rs
```diff
@@ -40,7 +40,14 @@ pub async fn run(
                 // We can simply respond by telling the engine to start consensus
                 // at the next height, and provide it with the genesis validator set
                 if reply
-                    .send((start_height, genesis.validator_set.clone()))
+                    .send((
+                        start_height,
+                        state
```
romac, line 45: Let's extract this into a variable

code/crates/test/app/src/app.rs
```diff
@@ -164,8 +171,17 @@ pub async fn run(
...
-            AppMsg::GetValidatorSet { height: _, reply } => {
-                if reply.send(genesis.validator_set.clone()).is_err() {
+            AppMsg::GetValidatorSet { height, reply } => {
+                if reply
+                    .send(
+                        state
+                            .ctx
+                            .middleware()
+                            .get_validator_set(&state.ctx, state.current_height, height, &genesis)
+                            .expect("Validator set should be available"),
```
romac, line 181: Let's extract this into variable

code/crates/test/app/src/app.rs
```diff
@@ -195,7 +211,16 @@ pub async fn run(
...
+                                state
+                                    .ctx
+                                    .middleware()
+                                    .get_validator_set(
+                                        &state.ctx,
+                                        state.current_height,
+                                        state.current_height,
+                                        &genesis,
+                                    )
+                                    .expect("Validator set should be available"),
```
romac, line 223: Let's extract this into a variable

code/crates/test/app/src/app.rs
```diff
@@ -210,7 +235,16 @@ pub async fn run(
...
```
romac, line 247: Let's extract this into a variable

code/crates/test/src/middleware.rs
```diff
@@ -62,3 +72,103 @@ pub trait Middleware: fmt::Debug + Send + Sync {
...
+        let end_index = start_index + self.selection_size;
+
+        let selected = genesis
+            .validator_set
+            .get_all()
+            .into_iter()
+            .cycle()
+            .skip(start_index)
+            .take(end_index - start_index)
+            .collect::<Vec<_>>();
```
romac, line 109: This can be rewritten as 

```rust
        let start_index = height.as_u64() as usize % num_validators;

        let selected = genesis
            .validator_set
            .iter()
            .cycle()
            .skip(start_index)
            .take(self.selection_size)
            .cloned()
            .collect::<Vec<_>>();
```

if you replace `get_all` by this:

```rust
    /// Iterate over the validators in the set
    pub fn iter(&self) -> slice::Iter<Validator> {
        self.validators.iter()
    }
```

romac: This can maybe be extracted into a standalone function to share its code between both middleware

code/crates/test/src/middleware.rs
```diff
@@ -62,3 +72,103 @@ pub trait Middleware: fmt::Debug + Send + Sync {
...
```
romac, line 170: Same comment as above.

code/crates/test/src/validator_set.rs
```diff
@@ -69,6 +69,16 @@ impl ValidatorSet {
         }
     }
 
+    /// Get the number of validators in the set
+    pub fn len(&self) -> usize {
+        self.validators.len()
+    }
+
+    /// Get the validators in the set
+    pub fn get_all(&self) -> Vec<Validator> {
```
romac, line 78: See comment above. In general let's avoid introducing functions or methods which clone a lot of data under the hood. Not really an issue in the test crate but we should be mindful about this in general :)

code/crates/test/tests/it/validator_set.rs
```diff
@@ -0,0 +1,26 @@
+use std::time::Duration;
+
+use informalsystems_malachitebft_test::middleware::RotateValidators;
+
+use crate::TestBuilder;
+
+#[tokio::test]
+async fn rotate_validator_set() {
+    const HEIGHT: u64 = 3;
```
romac, line 9: Let's increase the height a bit just to make sure:

```suggestion
    const HEIGHT: u64 = 9;
```

romac (changes requested):

code/crates/test/tests/it/validator_set.rs
```diff
@@ -0,0 +1,26 @@
...
+
+    for _ in 0..NUM_NODES {
+        test.add_node()
+            .with_middleware(RotateValidators {
+                selection_size: NUM_VALIDATORS_PER_HEIGHT,
+            })
+            .start()
+            .wait_until(HEIGHT)
+            .success();
+    }
```
ancazamfir, line 23: In this test or a new one, can you add a full node that starts late and tries to catch up.

bastienfaivre: I Added `start_late_rotate_validator_set` test in `value_sync.rs`.

ancazamfir: In this test `node-3` is a validator for height 1 so nothing happens until it comes up.
One test you can run with the rotate validator middleware is https://github.com/informalsystems/malachite/blob/cbfc5d92f458ca851876ac3abd8b4302a4809531/code/crates/test/tests/it/full_nodes.rs#L6

The interesting part on this one will be when there is a notion of epoch and we won't be able to verify the votes/ proposals as we can't get the valset for far heights. For example here:
https://github.com/informalsystems/malachite/blob/cbfc5d92f458ca851876ac3abd8b4302a4809531/code/crates/core-consensus/src/handle/vote.rs#L36

Code might change around here, if for example we go with #942 we will encounter this earlier (in the engine) and we should drop messages that we cannot verify hoping value sync will fix it when we reach that height.
As a side note, we buffer messages in consensus state (`input_queue`) and also in the driver (`pending_inputs`)...we might be verifying messages twice. Also we might be moving the `input_queue` messages to `pending_inputs` a bit late. cc @romac 

We can look separately into the issues above but in the meantime we can just add the test.

docs/tutorials/channels.md
```diff
@@ -513,7 +513,7 @@ fn factor_value(value: Value) -> Vec<u64> {
 
 ```
 
-Then, let's introduce _getter_ methods that are quite self-explanatory:
+Then, let's introduce _getter_ methods that are quite self-explanatory. For the validator set, we select a  rotating subset (floor((n+1)/2)) of the validators from the genesis validator set based on the height. This is done to showcase the ability to change the validator set over time (or height, in this case).
```
ancazamfir, line 516: Since there are multiple methods here maybe move the `get_validator_set` comment to the method comment instead and keep the comment here the way it was.

ancazamfir (approved):
Great work @bastienfaivre!

romac (approved):

romac:
Nice work! Can you please add entries in RELEASE_NOTE,md and in BREAKING_CHANGES.md? Then we can merge :)

romac:
> Can you please add entries in RELEASE_NOTE,md and in BREAKING_CHANGES.md? Then we can merge :)
Nevermind, it's actually not necessary.

PR #997, opened 2025-05-01
feat(code): Implement minimal gossip properties to ensure liveness

ancazamfir:
Closes: #998 

Implement minimal gossip properties for liveness.

### Summary of changes

1. **Liveness messages**:
   - Added a new `/liveness` topic/ channel and message types for network communication
   - Introduced new protobuf definitions in `liveness.proto` for handling gossip messages
   - Added support for `PolkaCertificate` and `RoundCertificate` message types

2. **Network Layer Changes**:
   - Added a new `Liveness` channel type alongside existing channels (Consensus, ProposalParts, Sync)
   - Modified network event handling to distinguish between consensus and liveness messages
   - Updated message encoding/decoding to support the new liveness message types

3. **Certificate Handling**:
   - Moved `PolkaCertificate` and related types from sync proto to liveness proto
   - Added new `RoundCertificate` type for handling round-specific certificates

4. **Codec Updates**:
   - Added new codec implementations for liveness messages
   - Updated protobuf codec to handle the new message types

Still to be done:
~~- [ ] Remove vote set mode~~ - done in separate PR #1008
~~- [ ] Move [rebroadcast](https://github.com/informalsystems/malachite/blob/main/code/crates/core-consensus/src/handle/rebroadcast_timeout.rs) in its own module and outside core-consensus~~ - will do in separate PR
- [x] Proper verification for gossip messages (in particular the round certificate)
   - Currently verify vote by vote rather than adding new effect and verification in the engine
   - Let's reconsider depending on when #942 lands
- [x] Decide if we store the round certificate when it occurs (i.e. we got `f+1` quorum or `2f+1` precommit any) similarly with `last_prevote` and `last_precommit` or find the certificate when we enter a round > 0 (current approach)
- [x]  Write IT tests 

---

#### For all contributors


code/crates/core-consensus/src/effect.rs
```diff
@@ -84,6 +84,11 @@ where
     /// Resume with: [`resume::Continue`]
     Publish(SignedConsensusMsg<Ctx>, resume::Continue),
 
+    /// Publish a polka certificate to peers
```
romac, line 87:
```suggestion
    /// Publish a gossip message to peers
```

code/crates/core-consensus/src/effect.rs
```diff
@@ -84,6 +84,11 @@ where
     /// Resume with: [`resume::Continue`]
     Publish(SignedConsensusMsg<Ctx>, resume::Continue),
```
romac, line 85: Should we consider renaming `Publish` to `PublishConsensusMsg`?

code/crates/core-consensus/src/effect.rs
```diff
@@ -84,6 +84,11 @@ where
...
 
+    /// Publish a polka certificate to peers
+    ///
+    /// Resume with: [`resume::Continue`]
+    PublishGossipMessage(GossipMsg<Ctx>, resume::Continue),
```
romac, line 90:
```suggestion
    PublishGossipMsg(GossipMsg<Ctx>, resume::Continue),
```

Since we use _msg_ instead of _message_ all over the codebase already

code/crates/core-consensus/src/types.rs
```diff
@@ -114,3 +115,9 @@ pub enum VoteExtensionError {
     #[error("Invalid vote extension")]
     InvalidVoteExtension,
 }
+
+#[derive_where(Clone, Debug, PartialEq, Eq)]
+pub enum GossipMsg<Ctx: Context> {
```
romac, line 120: I am a bit wary of calling those "gossip" messages and of naming the channel "/gossip" as it might be very confusing for users. How about renaming those to "liveness" instead? It's not great either, and less correct than "gossip" but at least it avoids the confusion. What do you think?

ancazamfir: I am ok with it. @nenadmilosevic95 what do you think?

romac:
@ancazamfir Added a basic integration test, looking good so far :)

code/crates/core-consensus/src/effect.rs
ancazamfir: 2526b2c017db83ed680f800c59a6efea783246d1

code/crates/core-consensus/src/effect.rs
ancazamfir: 2526b2c017db83ed680f800c59a6efea783246d1

code/crates/core-consensus/src/effect.rs
ancazamfir: 2526b2c017db83ed680f800c59a6efea783246d1

code/crates/core-consensus/src/types.rs
ancazamfir: 2526b2c017db83ed680f800c59a6efea783246d1

code/crates/core-consensus/src/handle/driver.rs
```diff
@@ -250,18 +265,63 @@ where
...
+                        co,
+                        Effect::RestreamProposal(
+                            signed_proposal.height(),
+                            signed_proposal.round(),
+                            signed_proposal.pol_round(),
+                            signed_proposal.validator_address().clone(),
+                            signed_proposal.value().id(),
+                            Default::default()
+                        )
+                    );
```
ancazamfir, line 311: We should also broadcast the Proposal message

ancazamfir, line 311: We should also rebroadcast the `signed_proposal`.

code/crates/core-consensus/src/state.rs
```diff
@@ -143,6 +143,15 @@ where
         Some((votes, certificates))
     }
 
+    pub fn polka_certificate_at_round(&self, round: Round) -> Option<PolkaCertificate<Ctx>> {
+        // Get the polka certificate for the specified round if it exists
+        self.driver
+            .polka_certificates()
```
nenadmilosevic95, line 149: Not directly related to this PR, but what polka_certificates() do we store?

code/crates/core-driver/src/driver.rs
```diff
@@ -227,6 +232,11 @@ where
         &self.polka_certificates
     }
 
+    /// Get the round certificate for the given round.
```
nenadmilosevic95, line 235:
```suggestion
    /// Get the round certificate for the current round.
```
Since the round is not passed here?

code/crates/core-driver/src/driver.rs
```diff
@@ -487,6 +500,45 @@ where
...
+        };
+
+        let precommits: Vec<SignedVote<Ctx>> = per_round
+            .received_votes()
+            .iter()
+            .filter(|v| v.vote_type() == VoteType::Precommit)
+            .cloned()
+            .collect();
+
+        self.round_certificate = Some(RoundCertificate::new_from_votes(
```
nenadmilosevic95, line 515: We don't check that RoundCertificate here has 2f+1 Precommits? I guess this is done somewhere else. Do we esnure that it will have exactly 2f+1 Precommits (2f+1 voting-power worth of Precommits) ?

code/crates/core-driver/src/driver.rs
```diff
@@ -487,6 +500,45 @@ where
...
+
+        let mut seen_addresses = BTreeSet::new();
+        let skip_votes: Vec<_> = per_round
+            .received_votes()
+            .iter()
+            .filter(|vote| seen_addresses.insert(vote.validator_address()))
+            .cloned()
+            .collect();
+
+        self.round_certificate = Some(RoundCertificate::new_from_votes(
```
nenadmilosevic95, line 535: Similar comment as above, only here we need f+1 any votes

BREAKING_CHANGES.md
```diff
@@ -2,15 +2,26 @@
 
 ## Unreleased
 
+### `malachitebft-core-consensus`
+- Added new variants to `Input` enum: `PolkaCertificate` and `RoundCertificate`
+- Added new variant to `Effect` enum: `PublishGossipMessage`
+
 ### `malachitebft-engine`
 - Changed the reply channel of `GetValidatorSet` message to take an `Option<Ctx::ValidatorSet>` instead of `Ctx::ValidatorSet`.
+- Added new variant to `Msg` enum: `PublishGossipMsg`
```
nenadmilosevic95, line 11:
```suggestion
- Added new variant to `Msg` enum: `PublishLivenessMsg`
```

BREAKING_CHANGES.md
```diff
@@ -2,15 +2,26 @@
...
+- Added new variants to `NetworkEvent` enum: `PolkaCertificate` and `RoundCertificate`
 - Changed `PartStore::all_parts` to `PartStore::all_parts_by_stream_id`:
   - Renamed method to clarify that, when a new part is received, the contiguous parts should be queried by stream id
   - Added required `StreamId` parameter
 - Added new public API `PartStore::all_parts_by_value_id` to be used instead of `PartStore::all_parts` when a decision is reached
 - Added `&StreamId` parameter to `part_store::PartStore::store`
 - Added `&StreamId` parameter to `part_store::PartStore::store_value_id`
 
+### `malachitebft-network`
+- Added new variant to `Channel` enum: `Gossip`
```
nenadmilosevic95, line 21:
```suggestion
- Added new variant to `Channel` enum: `Liveness`
```

nenadmilosevic95 (commented):
@ancazamfir @romac, I made a full pass and left comments. It looks good, great job!  

I only did a quick scan of changes in `tests/it` but for this I need to understand how these tests work. If you think it makes sense that I invest time there let me know. 

Two questions not related directly to this PR:
1) When do we prune `vote_keeper` ? 
2) What `polka_certificates` do we store? Do we store all that we receive and when do we prune also that?

code/crates/core-consensus/src/handle/driver.rs
ancazamfir: done in 039dd377b1f8b5df1130ab8abed290be1a14654a

code/crates/core-consensus/src/effect.rs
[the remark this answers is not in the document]
romac: They have been renamed to `LivenessMsg`

code/crates/engine/src/consensus.rs
```diff
@@ -1066,27 +1096,59 @@ where
...
-                    .map_err(|e| eyre!("Error when broadcasting gossip message: {e:?}"))?;
+                    .cast(NetworkMsg::PublishConsensusMsg(msg))
+                    .map_err(|e| eyre!("Error when broadcasting consensus message: {e:?}"))?;
+
+                Ok(r.resume_with(()))
+            }
+
+            Effect::PublishLivenessMsg(msg, r) => {
+                // Publish liveness message only if vote sync mode is set to "rebroadcast".
+                if self.params.vote_sync_mode == VoteSyncMode::Rebroadcast {
```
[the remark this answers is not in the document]
romac: We will remove it in https://github.com/informalsystems/malachite/pull/1008, but only after merging this PR in, hence the need for the check in the meantime.

code/crates/core-consensus/src/handle/liveness.rs
```diff
@@ -0,0 +1,105 @@
...
+                ),
+                signature.signature,
+            ),
+        };
+
+        if !verify_signed_vote(co, state, &vote).await? {
+            warn!(?vote, "Invalid vote");
+            continue;
+        }
+        apply_driver_input(co, state, metrics, DriverInput::Vote(vote)).await?;
```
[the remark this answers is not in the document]
ancazamfir: This is on receive side where we don't trust the sender.

code/crates/core-consensus/src/handle/signature.rs
```diff
@@ -78,3 +78,16 @@ where
...
+
+pub async fn verify_round_certificate<Ctx>(
+    _co: &Co<Ctx>,
+    _certificate: RoundCertificate<Ctx>,
+    _validator_set: Ctx::ValidatorSet,
+    _threshold_params: ThresholdParams,
+) -> Result<Result<(), CertificateError<Ctx>>, Error<Ctx>>
+where
+    Ctx: Context,
+{
```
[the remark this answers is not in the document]
romac: - Commit certificates: https://github.com/informalsystems/malachite/blob/main/code/crates/core-types/src/signing.rs#L320
- Polka certificates: https://github.com/informalsystems/malachite/blob/main/code/crates/core-types/src/signing.rs#L380
- For round certificates, each vote it contain is verified independently with `verify_signed_vote` which contains the check: https://github.com/informalsystems/malachite/blob/main/code/crates/core-consensus/src/handle/vote.rs#L121

code/crates/engine/src/network.rs
```diff
@@ -347,7 +362,35 @@ where
...
+                        LivenessMsg::SkipRoundCertificate(round_cert) => {
+                            NetworkEvent::RoundCertificate(from, round_cert)
+                        }
+                    };
+
+                    output_port.send(event);
+                }
+            }
+
+            Msg::NewEvent(Event::ConsensusMessage(Channel::Liveness, from, _)) => {
```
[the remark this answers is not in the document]
romac: Yes it's just a sanity check.

code/crates/core-consensus/src/handle/rebroadcast_timeout.rs
```diff
@@ -15,32 +15,24 @@ where
...
-        TimeoutKind::PrevoteRebroadcast => (
-            state.last_signed_prevote.as_ref(),
-            Timeout::prevote_rebroadcast(round),
-        ),
-        TimeoutKind::PrecommitRebroadcast => (
-            state.last_signed_precommit.as_ref(),
-            Timeout::precommit_rebroadcast(round),
-        ),
-        _ => return Ok(()),
+    if let Some(vote) = state.last_signed_prevote.as_ref() {
```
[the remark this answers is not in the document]
ancazamfir: We overwrite just before we enter new round when the `PrecommitAny` or `SkipRound` conditions occur
https://github.com/informalsystems/malachite/blob/3db290c4fae012cf036053baf4593611c4e6cd03/code/crates/core-driver/src/driver.rs#L468-L469

code/crates/core-driver/src/driver.rs
```diff
@@ -487,6 +500,45 @@ where
         })
     }
 
+    fn store_precommit_any_round_certificate(&mut self, vote_round: Round) {
+        let Some(per_round) = self.vote_keeper.per_round(vote_round) else {
+            return;
```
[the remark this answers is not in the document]
ancazamfir: Yes, in fact we should probably put an assert here since the driver just saw a precommit any and wants to store it.

code/crates/core-driver/src/driver.rs
ancazamfir: Yes, it happened in the vote keeper when a vote is applied. It came back with `PrecommitAny` and we immediately store here. Same for `SkipRound`.

code/crates/core-types/src/certificate.rs
```diff
@@ -166,3 +168,133 @@ pub enum CertificateError<Ctx: Context> {
...
+        Self {
+            vote_type,
+            value_id,
+            address,
+            signature,
+        }
+    }
+}
+
+/// Represents a certificate witnessing a Polka at a given height and round.
```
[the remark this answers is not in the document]
ancazamfir: 53170da3280bd28f3b9beb5deb318901ec4ff2b6

code/crates/core-types/src/certificate.rs
```diff
@@ -166,3 +168,133 @@ pub enum CertificateError<Ctx: Context> {
...
+#[derive_where(Clone, Debug, PartialEq, Eq)]
+pub struct RoundCertificate<Ctx: Context> {
+    /// The height at which a Polka was witnessed
+    pub height: Ctx::Height,
+    /// The round at which a Polka that was witnessed
+    pub round: Round,
+    /// The value that the Polka is for
```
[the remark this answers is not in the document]
ancazamfir: 53170da3280bd28f3b9beb5deb318901ec4ff2b6

code/crates/engine/src/network.rs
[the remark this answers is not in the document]
ancazamfir: Yes, it should not happen.

code/crates/network/src/lib.rs
```diff
@@ -569,15 +570,26 @@ async fn handle_gossipsub_event(
...
+                let event = Event::LivenessMessage(
+                    channel,
+                    PeerId::from_libp2p(&peer_id),
+                    Bytes::from(message.data),
+                );
+                if let Err(e) = tx_event.send(event).await {
+                    error!("Error sending message to handle: {e}");
+                    return ControlFlow::Break(());
+                }
+            } else {
```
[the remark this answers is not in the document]
ancazamfir: It can. All non-liveness messages are bundled under `Event::ConsensusMessage`. Maybe we can cleanup things a bit here.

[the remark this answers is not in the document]
romac: Agreed, we really need to decouple proposal parts streaming from consensus, and make it a pure app-level concern. But for that we need to expose the networking layer to the app somehow.

code/crates/core-driver/src/driver.rs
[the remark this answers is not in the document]
ancazamfir: Actually we have `deny(clippy::panic)` here. @romac should we remove it?

code/crates/core-consensus/src/handle/driver.rs
```diff
@@ -250,18 +265,74 @@ where
...
                         co,
-                        Effect::Publish(
+                        Effect::PublishConsensusMsg(
                             SignedConsensusMsg::Proposal(signed_proposal),
                             Default::default()
                         )
                     );
                 };
+
+                if proposal.pol_round().is_defined() {
```
nenadmilosevic95, line 275:
```suggestion
                // Publishing the polka certificate of the re-proposed value  
                // ensures all validators receive it, which is necessary for 
                // them to accept the re-proposed value.
                if proposal.pol_round().is_defined() {
```

code/crates/core-consensus/src/handle/driver.rs
```diff
@@ -250,18 +265,74 @@ where
...
+                        );
+                    }
+                }
             }
 
             Ok(())
         }
 
         DriverOutput::Vote(vote) => {
+            if vote.vote_type() == VoteType::Precommit
```
nenadmilosevic95, line 294:
```suggestion
            // Upon locking, in addition to publishing a Precommit message,  
            // a validator must request the application to restream the proposal,  
            // publish the proposal message, and publish the polka certificate.  
            // In other words, it must ensure that all validators receive the same events  
            // that led it to lock a value. Together with the timeout mechanisms,  
            // this guarantees that after GST, all correct validators will update  
            // their validValue and validRound to these values in this round. 
            // As a result, Malachite ensures liveness, because all validators  
            // will be aware of the most recently locked value, and whichever validator  
            // becomes the leader in one of the following rounds will propose a value  
            // that all correct validators can accept.  
            // Importantly, this mechanism does not need to be enabled from round 0,  
            // as it is expensive; it can be activated from any round as a last-resort  
            // backup to guarantee liveness.  
            if vote.vote_type() == VoteType::Precommit
```

code/crates/core-driver/src/driver.rs
nenadmilosevic95: Ok great, I was wondering if there is some delay (an event that can occur) in between that can lead to the situation where we add more votes here than necessary

code/crates/core-consensus/src/state.rs
nenadmilosevic95: We actually just need polka certificates that correspond to our locked value and valid value. Right?

nenadmilosevic95 (commented):
Added comments when publishing liveness events

code/crates/network/src/lib.rs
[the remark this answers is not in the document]
ancazamfir: Let's track this in a separate issue.

code/crates/core-consensus/src/handle/liveness.rs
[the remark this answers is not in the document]
ancazamfir: summarizing the discussion we had:
- for polka certificate for non-nil value, we need to apply as single input as it may contain equivocating votes that would be dropped in a vote=by=vote counting.
- for round vertificate which are `Any` type of certificates the result of processing vote-by-vote is the same in the presence of equivocating votes (we still count the stored one)
Will add some comments to reflect this.

code/crates/core-consensus/src/state.rs
ancazamfir: For vote set mode we store and send them all. But this will be removed soon.

For rebroadcast, with the current PR, it will be for the last valid value, may or may not be locked:
- L16: We re-propose with (latest) valid but not necessarily locked, if we skipped L37 in round `vr`.
- L40: For the hidden lock it must be the case that it is locked.

code/crates/core-consensus/src/handle/liveness.rs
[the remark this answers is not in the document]
ancazamfir: Added comments in 1c457d70ea48184b4b19b9de28c4a8c276abf5bc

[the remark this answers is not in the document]
ancazamfir: While writing the comments for the polka certificate, I realized we have the same problem as for commit certificate. We don't store the equivocating proposals in the driver (we keep track in misbehavior but we don't mux them further).
Let's discuss, we might need to fix it for this case.

code/crates/core-driver/src/driver.rs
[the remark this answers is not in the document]
romac: Either that or you can use `unreachable!` instead if you think it conveys the meaning better.

nenadmilosevic95:
> > @ancazamfir @romac, I made a full pass and left comments. It looks good, great job!
> > I only did a quick scan of changes in `tests/it` but for this I need to understand how these tests work. If you think it makes sense that I invest time there let me know.
> > Two questions not related directly to this PR:
> > 
> > 1. When do we prune `vote_keeper` ?
> 
> When we move to a new height. We keep all votes for all rounds until we decide.
> 
> > 2. What `polka_certificates` do we store? Do we store all that we receive and when do we prune also that?
> 
> We store them as they are detected by the vote keeper. ~I think we should remove these when we remove the vote set mode as they are only required there.~ So we still need them for rebroadcast and the code in this PR. Maybe since the cases they are needed for are very rare, we can restore them from votes when needed.

Yes, this is a possibility but we again need to be careful because we would need to allow equivocated votes in the vote keeper. Let's leave it as is for now.

nenadmilosevic95:
One way a node can prune (if we decide to do it at some point) is to remove all polka certificates from round smaller than its `lockedRound` because anyway it knows it will not accept values with such `pol_round`. I guess the same logic we can do also for a `vote_keeper`

code/crates/core-consensus/src/handle/liveness.rs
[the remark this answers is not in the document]
ancazamfir: Opened #1018

ancazamfir:
> One way a node can prune (if we decide to do it at some point) is to remove all polka certificates from round smaller than its `lockedRound` because anyway it knows it will not accept values with such `pol_round`. I guess the same logic we can do also for a `vote_keeper`

Tracked in #1019

code/crates/core-consensus/src/handle/driver.rs
```diff
@@ -250,18 +271,101 @@ where
...
+            // this guarantees that after GST, all correct validators will update
+            // their validValue and validRound to these values in this round.
+            // As a result, Malachite ensures liveness, because all validators
+            // will be aware of the most recently locked value, and whichever validator
+            // becomes the leader in one of the following rounds will propose a value
+            // that all correct validators can accept.
+            // Importantly, this mechanism does not need to be enabled from round 0,
+            // as it is expensive; it can be activated from any round as a last-resort
+            // backup to guarantee liveness.
+            if vote.vote_type() == VoteType::Precommit
```
ancazamfir, line 322: Add a test for this case

ancazamfir: 03ae78fc8050e7f02d812cca2a23bdfdce2639bd

code/crates/test/tests/it/liveness.rs
```diff
@@ -0,0 +1,138 @@
...
+
+fn expect_hidden_lock_messages(node: &mut TestNode<TestContext>, round: u32) {
+    node.expect_polka_certificate(1, round);
+}
+
+#[tokio::test]
+async fn polka_certificate_for_hidden_lock() {
+    const FINAL_HEIGHT: u64 = 3;
+    // TODO: Use the actual hidden lock round constant
+    const HIDDEN_LOCK_ROUND: u32 = 10;
```
ancazamfir, line 100: Get this from the driver

romac: 8e9cba6f01c9f5d4665310d64b3074158c82b22e

ancazamfir:
> Yes, this is a possibility but we again need to be careful because we would need to allow equivocated votes in the vote keeper. Let's leave it as is for now.

I realized that I misread your comment, I thought it was for the locally generated polka certificates. For these we do store them [here](https://github.com/informalsystems/malachite/blob/2a330c37326452195751bc8fd309a8e47b4d7b11/code/crates/core-driver/src/driver.rs#L467)
But not sure we need them as we don't rebroadcast anymore (we need them in the vote set mode but this will go away) and for consensus purposes we have the votes locally. 
And yes, we do need to store the received ones (done [here](https://github.com/informalsystems/malachite/blob/2a330c37326452195751bc8fd309a8e47b4d7b11/code/crates/core-driver/src/mux.rs#L253))

romac (approved):

ancazamfir:
I enabled these rebroadcast mode tests in 0d7321ea497ce7e23d06efbf8038cf448ab27828:
```
restart_with_byzantine_proposer_1_rebroadcast_parts_only
restart_with_byzantine_proposer_1_rebroadcast_proposal_and_parts
restart_with_byzantine_proposer_2_rebroadcast_parts_only
restart_with_byzantine_proposer_2_rebroadcast_proposal_and_parts
```

They all passed locally, however on CI we get
```Error publishing message: InsufficientPeers channel=/liveness```

I will disable them again and open an issue to solve the failures in a separate PR

**Comment by ancazamfir** [comment]
Also `liveness::round_certificate_rebroadcast` fails in commit f7bd6002cb030ed6432d66fac0d94cb9add9fead. Passes locally.

**Comment by nenadmilosevic95** [comment]
> @nenadmilosevic95 @romac one last thing, we need to decide on (from the round sync spec):
> 
> "Open questions:
> 
>     * What should be the value we adopt for rebroadcastTimeout?"

What is the value that you currently use?

**Comment by nenadmilosevic95** [comment]
> > > @nenadmilosevic95 @romac one last thing, we need to decide on (from the round sync spec):
> > > "Open questions:
> > > ```
> > > * What should be the value we adopt for rebroadcastTimeout?"
> > > ```
> > 
> > 
> > What is the value that you currently use?
> 
> I still need to do the change to reflect the round sync proposal. Currently we use the old timeout mechanism. We have two timers, `PrevoteRebroadcast` with duration equal to the `timeout_prevote` and, similarly, `PrecommitRebroadcast` with duration equal to the `timeout_precommit`. When we enter Prevote state we start the `PrevoteRebroadcast` timer and if it expires we resend last prevote and precommit. When we enter Precommit state we start the `PrecommitRebroadcast` timer and we resent the last votes when it expires. In both cases we restart the timer that just expired. The timer duration does not take into account the deltas.
> 
> We will have only one timer as proposed but what should be the duration?

What was the motivation to have two timeouts? 
By "does not take into account the deltas" you mean that it does not increase as we progress in higher rounds?

**Comment by ancazamfir** [comment]
> What was the motivation to have two timeouts?

It was the initial rebroadcast design that starknet proposed. And (in main) we are sending only the last prevote on `PrevoteRebroadcast` expiry and the last precommit on `PrecommitRebroadcast`

**Review by nenadmilosevic95 — COMMENTED** [review]
LGTM, left some minor comments

File: code/crates/config/src/lib.rs [hunk 25]
```diff
@@ -549,8 +554,10 @@ impl TimeoutConfig {
             TimeoutKind::Precommit => self.timeout_precommit,
             TimeoutKind::PrevoteTimeLimit => self.timeout_step,
             TimeoutKind::PrecommitTimeLimit => self.timeout_step,
-            TimeoutKind::PrevoteRebroadcast => self.timeout_prevote,
-            TimeoutKind::PrecommitRebroadcast => self.timeout_precommit,
+            // TODO - clarify the rebroadcast timeout duration
```
nenadmilosevic95, line 557: remove TODO

code/crates/core-consensus/src/handle/liveness.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    // Cancel rebroadcast timer
+    // TODO: Should do only if the round certificate is well formed, i.e. either PrecommitAny or SkipRound
```
nenadmilosevic95, line 148: Do we want to do this?

nenadmilosevic95 (approved):
LGTM

code/crates/core-types/src/timeout.rs
```diff
@@ -67,14 +64,9 @@ impl Timeout {
...
 
-    /// Create a new timeout for rebroadcasting the last prevote.
-    pub const fn prevote_rebroadcast(round: Round) -> Self {
-        Self::new(round, TimeoutKind::PrevoteRebroadcast)
-    }
-
-    /// Create a new timeout for rebroadcasting the last precommit.
-    pub const fn precommit_rebroadcast(round: Round) -> Self {
-        Self::new(round, TimeoutKind::PrecommitRebroadcast)
+    /// Create a new timeout for rebroadcasting the liveness messages.
```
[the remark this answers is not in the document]
ancazamfir: fae6219fb4d92219cab092e6733db7248b6fe705

code/crates/core-types/src/timeout.rs
```diff
@@ -20,11 +20,8 @@ pub enum TimeoutKind {
     /// Timeout for detecting consensus being in the precommit step for too long.
     PrecommitTimeLimit,
 
-    /// Timeout to rebroadcast the last prevote
-    PrevoteRebroadcast,
-
-    /// Timeout to rebroadcast the last precommit
-    PrecommitRebroadcast,
+    /// Timeout to rebroadcast the liveness messages
```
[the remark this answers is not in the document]
ancazamfir: fae6219fb4d92219cab092e6733db7248b6fe705

code/crates/core-driver/src/driver.rs
```diff
@@ -63,6 +64,11 @@ where
 
     last_prevote: Option<Ctx::Vote>,
     last_precommit: Option<Ctx::Vote>,
+
+    /// The certificate that justifies moving to `round`.
+    /// For `PrecommitAny` case it will be `round + 1` of the round signatures.
+    /// For `SkipRound` case it will be the `round` of the round signatures.
```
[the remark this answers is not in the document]
ancazamfir: ah good catch! I will move the comments to `EnterRoundCertificate`, they are meant to explain the `enter_round` vs the round the signatures are for. Let me know if they make more sense with fae6219fb4d92219cab092e6733db7248b6fe705. Forgot to move them when I added `EnterRoundCertificate` (used to be here as a tuple `(Round, RoundCertificate)`

code/crates/core-consensus/src/handle/liveness.rs
ancazamfir: If we get a certificate that will actually not move us to a new round (e.g. a `PrecommitAny` with < 2f+1) we would have cancelled the timer and will not be started again.

nenadmilosevic95: Yes I understand, I am just asking if we want to tackle this in this PR

ancazamfir: Let's do it in a separate one.

code/crates/core-driver/src/driver.rs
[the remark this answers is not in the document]
nenadmilosevic95: Ok I cannot do it directly, so I would replace this 
```
/// The round that is or will be entered due to the certificate
/// For `PrecommitAny` case it will be `round + 1` of the round signatures in `RoundCertificate`.
/// For `SkipRound` case it will be the `round` of the round signatures.
``` 

with 
```
/// The round that will be entered due to the `RoundCertificate`.
/// - If the certificate is `PrecommitAny`, it contains signatures from the previous round,
///   so `enter_round` will be one more than the round of those signatures.
/// - If the certificate is `SkipRound`, it contains signatures from the round being entered,
///   so `enter_round` will be equal to the round of those signatures.
```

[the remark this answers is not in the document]
ancazamfir: nice! e0c3353fb6bd07989cb70ddaf987243471ea97c1

romac:
🎉

code/crates/core-consensus/src/handle/liveness.rs
ancazamfir: opened https://github.com/informalsystems/malachite/issues/1036

Issue #911, opened 2025-03-12
code: Surface proposal and vote evidence


ancazamfir:
We currently store the proposal and vote equivocation evidence in the proposal and vote keepers. It is silent and no reporting is done.
We should
- [ ] surface to consensus engine and log 
- [ ] add to metrics
- [ ] report to the application

We can do the first to right away. For the third we should figure out what we want to do, e.g. send on `Decided` or something else.

PR #999, opened 2025-05-02
feat(code): Surface proposal and vote equivocation evidence

bastienfaivre:
Closes: #911

---

#### For all contributors


bastienfaivre:
If we want to inform the engine and the application of the equivocation that occurred during a height, we can simply add the evidence map to the `Decided` event. This is done in commit [3b83e0f](https://github.com/informalsystems/malachite/pull/999/commits/3b83e0f7ca3c325f00eb9c84ddc15fa0547662b2).
Now, if we want to know about the equivocation as soon as it occurs, we can throw an error for the vote and the proposal, respectively. This works well for the vote, since the engine (or more specifically, the driver) does not have to do anything other than return if it is an equivocation. However, this approach does not work for the proposal since, even if it is an equivocation, the driver still needs to execute `multiplex_proposal` (c.f. [mux.rs](https://github.com/informalsystems/malachite/pull/999/commits/1f4c5ecf616bfc37dd23a5ee0506d15b22343275#diff-d75b52ecb824dd9ad5f471b105565c99ad9eaef34394cd7a677f21809fa0e446)), which, depending on the output, will need to perform something else in the driver that can also lead to an error. Therefore, we may need to carry two errors, which is inconvenient.
Another approach would be to define proper data structures at the same level as the evidence map so that the engine can check and clear them every time it processes a vote or a proposal.

What do you think? @romac @ancazamfir

code/crates/core-consensus/src/handle/decide.rs
```diff
@@ -102,7 +102,13 @@ where
 
     perform!(
         co,
-        Effect::Decide(certificate, extensions, Default::default())
+        Effect::Decide(
```
ancazamfir, line 105: Can you add some evidence metrics above and also some error/warn logs?

ancazamfir:
> If we want to inform the engine and the application of the equivocation that occurred during a height, we can simply add the evidence map to the `Decided` event. This is done in commit [3b83e0f](https://github.com/informalsystems/malachite/pull/999/commits/3b83e0f7ca3c325f00eb9c84ddc15fa0547662b2). 

Let't do this in a separate PR. We need to figure out if we send the actual  equivocating votes/ proposals or just a statement "This node/address has equivocated". It depends on the app -- consensus trust level and if the app is "ready" to handle votes and proposals.

> Now, if we want to know about the equivocation as soon as it occurs, we can throw an error for the vote and the proposal, respectively. This works well for the vote, since the engine (or more specifically, the driver) does not have to do anything other than return if it is an equivocation. However, this approach does not work for the proposal since, even if it is an equivocation, the driver still needs to execute `multiplex_proposal` (c.f. [mux.rs](https://github.com/informalsystems/malachite/pull/999/commits/1f4c5ecf616bfc37dd23a5ee0506d15b22343275#diff-d75b52ecb824dd9ad5f471b105565c99ad9eaef34394cd7a677f21809fa0e446)), which, depending on the output, will need to perform something else in the driver that can also lead to an error. Therefore, we may need to carry two errors, which is inconvenient. Another approach would be to define proper data structures at the same level as the evidence map so that the engine can check and clear them every time it processes a vote or a proposal.

Yes, the latter is better. Not sure about clearing it though. So maybe we can check after applying a proposal or vote if "new" evidence is recorded in the driver. This is just for logging purposes. 
> 
> What do you think? @romac @ancazamfir

code/crates/core-consensus/src/effect.rs
```diff
@@ -133,6 +133,8 @@ where
     Decide(
         CommitCertificate<Ctx>,
         VoteExtensions<Ctx>,
+        malachitebft_core_driver::EvidenceMap<Ctx>,
+        malachitebft_core_votekeeper::EvidenceMap<Ctx>,
```
romac, line 137: Let's define our own `MisbehaviorEvidence` type in `core-consensus` which both evidence maps as fields.

code/crates/core-driver/src/driver.rs
```diff
@@ -45,9 +45,23 @@ where
     /// The proposals to decide on.
     pub(crate) proposal_keeper: ProposalKeeper<Ctx>,
 
+    /// The evidence of equivocation (if any) after processing a proposal.
+    pub(crate) proposal_equivocation_evidence: Option<(
+        Ctx::Height,
+        Ctx::Address,
+        (SignedProposal<Ctx>, SignedProposal<Ctx>),
+    )>,
```
romac, line 53: Let's define a proper type for this instead of using a tuple

code/crates/core-driver/src/driver.rs
```diff
@@ -45,9 +45,23 @@ where
...
+
     /// The vote keeper.
     pub(crate) vote_keeper: VoteKeeper<Ctx>,
 
+    /// The evidence of equivocation (if any) after processing a vote.
+    pub(crate) vote_equivocation_evidence: Option<(
+        Ctx::Height,
+        Ctx::Address,
+        (SignedVote<Ctx>, SignedVote<Ctx>),
```
romac, line 62: Same here

code/crates/core-driver/src/driver.rs
```diff
@@ -197,6 +213,38 @@ where
...
 
+    /// Return proposal equivocation evidence, if any.
+    pub fn proposal_equivocation_evidence(
+        &self,
+    ) -> Option<(
+        Ctx::Height,
+        Ctx::Address,
+        (SignedProposal<Ctx>, SignedProposal<Ctx>),
+    )> {
+        self.proposal_equivocation_evidence.clone()
```
romac, line 224: Let's return a reference here

code/crates/core-driver/src/driver.rs
```diff
@@ -197,6 +213,38 @@ where
...
+
+    /// Return vote equivocation evidence, if any.
+    pub fn vote_equivocation_evidence(
+        &self,
+    ) -> Option<(
+        Ctx::Height,
+        Ctx::Address,
+        (SignedVote<Ctx>, SignedVote<Ctx>),
+    )> {
+        self.vote_equivocation_evidence.clone()
```
romac, line 240: Same here

code/crates/engine/src/util/events.rs
```diff
@@ -47,14 +49,34 @@ pub enum Event<Ctx: Context> {
     Published(SignedConsensusMsg<Ctx>),
     ProposedValue(LocallyProposedValue<Ctx>),
     ReceivedProposedValue(ProposedValue<Ctx>, ValueOrigin),
-    Decided(CommitCertificate<Ctx>),
+    Decided {
+        commit_certificate: CommitCertificate<Ctx>,
+        proposal_equivocation_evidence_map: malachitebft_core_driver::EvidenceMap<Ctx>,
+        vote_equivocation_evidence_map: malachitebft_core_votekeeper::EvidenceMap<Ctx>,
```
romac, line 55: Here, let's also use the compound type that we will introduce in core-consensus

code/crates/engine/src/util/events.rs
```diff
@@ -47,14 +49,34 @@ pub enum Event<Ctx: Context> {
...
     WalReplayBegin(Ctx::Height, usize),
     WalReplayEntry(WalEntry<Ctx>),
     WalReplayDone(Ctx::Height),
     WalReplayError(Arc<ActorProcessingErr>),
+    ProposalEquivocationEvidence {
+        proposal_height: Ctx::Height,
+        address: Ctx::Address,
+        evidence: (
+            SignedMessage<Ctx, <Ctx as Context>::Proposal>,
+            SignedMessage<Ctx, <Ctx as Context>::Proposal>,
```
romac, line 69:
```suggestion
            SignedProposal<Ctx>,
            SignedProposal<Ctx>,
```

code/crates/engine/src/util/events.rs
```diff
@@ -47,14 +49,34 @@ pub enum Event<Ctx: Context> {
...
+        ),
+    },
+    VoteEquivocationEvidence {
+        vote_height: Ctx::Height,
+        address: Ctx::Address,
+        evidence: (
+            SignedMessage<Ctx, <Ctx as Context>::Vote>,
+            SignedMessage<Ctx, <Ctx as Context>::Vote>,
```
romac, line 77:
```suggestion
            SignedVote<Ctx>,
            SignedVote<Ctx>,
```

code/crates/core-driver/src/driver.rs
romac: By the way, can you clarify why we need these two new variables when we already have evidence maps in the proposal and vote keepers?

romac:
I opened https://github.com/informalsystems/malachite/pull/1009 which addresses some of my comments above and simplifies the implementation by only checking for evidence on decide, rather than emit events at the time when equivocation happens.

bastienfaivre:
@ancazamfir @romac What should this branch implement, then? Based on your perspectives, #1009 is already in better shape than this one.

romac:
See my comment in #1009.

code/crates/metrics/src/metrics.rs
```diff
@@ -235,6 +243,18 @@ impl Metrics {
...
+
+            registry.register(
+                "equivocation_votes",
+                "Number of equivocation votes",
+                metrics.equivocation_votes.clone(),
+            );
+
+            registry.register(
+                "equivocation_proposals",
+                "Number of equivocation proposals",
```
romac, line 255:
```suggestion
                "Number of equivocating proposals",
```

code/crates/metrics/src/metrics.rs
```diff
@@ -235,6 +243,18 @@ impl Metrics {
                 "Time taken to verify a signature, in seconds",
                 metrics.signature_verification_time.clone(),
             );
+
+            registry.register(
+                "equivocation_votes",
+                "Number of equivocation votes",
```
romac, line 249:
```suggestion
                "Number of equivocating votes",
```

code/crates/test/tests/it/equivocation.rs
```diff
@@ -0,0 +1,61 @@
...
+use malachitebft_core_types::{Round, SignedProposal};
+use malachitebft_engine::{consensus::Msg, network::NetworkEvent};
+use malachitebft_peer::PeerId;
+use malachitebft_signing_ed25519::Signature;
+use malachitebft_test_framework::TestParams;
+
+use crate::TestBuilder;
+
+#[tokio::test]
+pub async fn equivocation_proposer() {
```
bastienfaivre, line 14: @romac, the inject feature is not enough/complete. Indeed, injecting a `ProposeValue` message does not work as the engine/driver will not consider two propose values; therefore, no conflicting proposals are sent. If we want to force receiving a proposal via a `NetworkEvent`, it also does not work since we do not have access to the peer id and cannot sign the proposal.
Do you have an idea on how to fix this?

romac: Ah bummer… Let me think about it and I'll come back to you

romac: Maybe we can pass an unsigned message to `inject` and use `Node::get_signing_provided` to sign the message in the test framework?

bastienfaivre: For that, we need to have access to the [app instance](https://github.com/informalsystems/malachite/blob/main/code/crates/test/app/src/node.rs#L55) (on which we can actually call `get_signing_provider`), but I don't see where we can access it in the framework.

ancazamfir: Maybe an overkill but can you run once and collect the proposals from that run? And then run again (with same keys) and inject in the second run the proposal from the first?

ancazamfir: The other question, does it make sense to do the tests in a separate PR and merge this one as is?

ancazamfir: > Do you mean hardcoding the messages? For example, for votes, retrieve data from such logs:

I thought you could intercept the proposal (or vote) in the first run, save it in some test state and then inject it in the second run.

> I would likely prefer the inject approach, which might also be useful for other tests.

You still do the inject but you need a properly signed (second) proposal

romac: Nifty idea, I'll give it a go next week!

romac: Nifty idea, I'll give it a go next week!

sergio-mena:
I'm closing this PR, as the work is continuing in #1418

Issue #954, opened 2025-04-02
code: Document value sync protocol and interactions with consensus and application


ancazamfir:
Write an ADR to document value sync protocol architecture and interactions with consensus and application.

PR #1005, opened 2025-05-06
chore(docs): Sync ADR for protocol details and interactions with app and consensus

ancazamfir:
Closes: #954

---

#### For all contributors


docs/architecture/adr-005-value-sync.md
```diff
@@ -0,0 +1,156 @@
...
+   - On `StartedHeight`: Updates `sync_height` and `tip_height`, initiates value request if a peer is ahead
+   - On `Decided`: Updates `tip_height` and removes pending requests
+   - On `Status`: Updates peer information and triggers sync if needed
+   - On `Tick`: If there are no pending requests and a peer is ahead, initiates a new value request
+
+### Protocol Flow
+
+The value sync protocol operates between two nodes: a syncing node that needs to catch up and a helping peer that provides the values. The flow consists of two main phases:
+
+**Normal Operation**: During normal operation, nodes periodically broadcast their status to inform peers about their current height. This helps peers detect when a node falls behind or when other nodes are ahead.
```
nenadmilosevic95, line 74: When you say broadcast, does it send the status message only to its direct peers or it broadcast it to the whole network?

docs/architecture/adr-005-value-sync.md
```diff
@@ -0,0 +1,156 @@
...
+    C1->>S1: StartedHeight(height)
+    S1->>S2: BroadcastStatus(height)
+
+    Note over A1,C1: Node Falls Behind
+    S2->>S1: Status(peer_height)
+    S1->>S2: GetValueRequest(height)
+    S2->>A2: GetDecidedValue(height)
+    A2->>S2: DecidedValue(value, certificate)
+    S2->>S1: ValueResponse(value, certificate)
+    S2->>C1: ValueResponse(value, certificate)
```
nenadmilosevic95, line 111: IMHO I think `S1` should talk with `C1` and not `S2` directly? What do you think?

docs/architecture/adr-005-value-sync.md
```diff
@@ -0,0 +1,156 @@
...
+
+    C1->>C1: ProcessCommitCertificate(certificate)
+    C1->>A1: ProcessValue(height, value)
+    A1->>C1: ProposedValue(height, value, validity)
```
nenadmilosevic95, line 115: Why do we need to check with application if the value is valid or not? This value is already decided

docs/architecture/adr-005-value-sync.md
```diff
@@ -0,0 +1,156 @@
...
+    A1->>C1: StartHeight(height)
+    C1->>S1: StartedHeight(height)
+    S1->>S2: BroadcastStatus(height)
+
+    Note over A1,C1: Node Falls Behind
+    S2->>S1: Status(peer_height)
+    S1->>S2: GetValueRequest(height)
+    S2->>A2: GetDecidedValue(height)
+    A2->>S2: DecidedValue(value, certificate)
+    S2->>S1: ValueResponse(value, certificate)
```
nenadmilosevic95, line 110: It is interesting that now `sync` is responsible for value propagation and not the app

docs/architecture/adr-005-value-sync.md
```diff
@@ -0,0 +1,156 @@
...
+
+**Request Tracking**
+   - `pending_decided_value_requests`: Map of height to peer for pending value requests
+   - Used to track ongoing sync requests and prevent duplicate requests
+   - The pending request is removed once a response is received, regardless of the value validity
+   - If consensus decided on that value, a `Decided` event will be seen later
+   - If no decision is made, sync will retry to get the value from a different peer
+
+**State Transitions**
+   - On `StartedHeight`: Updates `sync_height` and `tip_height`, initiates value request if a peer is ahead
```
nenadmilosevic95, line 65: Why do you initiate a value request on `StartedHeight`?

nenadmilosevic95 (commented):
@ancazamfir thanks for starting this ADR! I left some comments!

docs/architecture/adr-005-value-sync.md
ancazamfir: To the whole network. Will clarify.

docs/architecture/adr-005-value-sync.md
ancazamfir: The value response is sent by S2 to the "Network" then back to the syncing node. On the syncing node both Consensus and Sync are notified and receive the response. Consensus will validate certificate and send the app the value for validation, then decides (if all good) and we go to the next height. Sync will remove the pending request (regardless the response validity) and update metrics. 
I initially had "Network" in the diagram but removed it because was harder to read. I can add it back or clarify in the text.

docs/architecture/adr-005-value-sync.md
ancazamfir: Yes it's a bit strange. There is some boilerplate code in sync and network for request/ response so we need the app to pass through there.

docs/architecture/adr-005-value-sync.md
nenadmilosevic95: But `StartedHeight` is local event that states the this node is starting the new height? or no?

docs/architecture/adr-005-value-sync.md
nenadmilosevic95: Are you sure? @bastienfaivre  wrote here - https://github.com/informalsystems/malachite/issues/996 that we use [libp2p-scatter](https://github.com/romac/libp2p-scatter) that only send status message to its peers?

docs/architecture/adr-005-value-sync.md
nenadmilosevic95: I was just thinking that it is cleaner if communication goes through sync to the consensus, and not directly to consensus.

docs/architecture/adr-005-value-sync.md
ancazamfir: yes. So when consensus starts height `h` it tells sync which immediately asks for value at `h` if it has peers at height`>h`. In the meantime consensus also runs for `h` trying to reach a decision via consensus.

docs/architecture/adr-005-value-sync.md
ancazamfir: no, you're right, I got it wrong.

docs/architecture/adr-005-value-sync.md
ancazamfir: I see, yes we could make it that way. It was more out of convenience since network actor can notify multiple actors about the response so consensus gets the value right away. And there is no need of coordination between consensus and sync at the syncing node.

docs/architecture/adr-005-value-sync.md
ancazamfir: fixed in 125906c43694d7e4f21809d0f3b277c372cbbfc2

docs/architecture/adr-005-value-sync.md
ancazamfir: I will mark this solved as it shows the current implementation.

romac (approved):

PR #1009, opened 2025-05-08
refactor(code): Simplify implementation by only retrieving evidence on decide

romac:


bastienfaivre:
I just read the comment of @ancazamfir [here](https://github.com/informalsystems/malachite/pull/999#issuecomment-2862924649). Should we keep the `evidence` field in the `Decided` event?

romac:
I would suggest to remove it in this PR for now and add logging (via tracing) in the driver whenever we see a misbehavior.

romac:
Having events would be nice but we need to find a way to implement those nicely and I am not yet sure how to go about that. Having mutable state in the driver that we check and clear every time is brittle, hence why I attempted to remove it here, but maybe there is a better way. Perhaps by changing the EvidenceMap structs to store the latest evidence that was recorded so that we can check if it matches the last vote/proposal that was processed and then emit the event. What do you think? On my side I will think more on this and come back to you.

bastienfaivre:
I agree that adding a field `last_evidence` to the data structure would allow us to avoid clearing any value. Moreover, I don't see any other way to inform the engine that such an event occurred in the driver than adding a field somewhere, so this is an ideal solution at first glance.
I suggest the following:

1. We finish this PR by adding the logging required.
2. We merge it into main.
3. We merge main into #999 and implement this `last_evidence` approach.

What do you think?

romac:
Sounds good to me! @ancazamfir?

ancazamfir:
> Sounds good to me! @ancazamfir?

Can we keep still keep all evidence and only log the "last" one? Eventually we will send some information to the app if/ when we decide, based on the overall evidence. In Cosmos for example iirc we only send the list of validators that equivocated and not the vote details.

ancazamfir:
Can we also add evidence metrics. Not sure in which PR we do what but I am ok with either.

bastienfaivre:
> Can we still keep all evidence and only log the "last" one? 
That is precisely what we are planning to do. This `last_evidence` approach is a clean way to inform the engine if an equivocation occurred during the last driver process. In all cases, the `evidence_map` will keep all evidence that occurred.

code/crates/core-votekeeper/src/keeper.rs
```diff
@@ -234,6 +235,13 @@ where
             }) => {
                 // This is an equivocating vote
                 self.evidence.add(existing.clone(), conflicting);
+
+                debug!(
```
romac, line 239: Let's change this to `warn!` and show the full votes

PR #1032, opened 2025-05-16
feat(code): Implement Solution 2 for the RoundSync

ancazamfir:
Closes: #XXX

From the Round sync spec:
- upon entering the round `r` processes broadcast the certificate _EnterRoundCertificate_, reset _lastPrevote_ and _lastPrecommit_, and start _rebroadcastTimeout_ 
- if _rebroadcastTimeout_ expires and the process still did not receive a round certificate for a new round a process calls `REBROADCAST()` and starts the `rebroadcastTimeout` again
- `timeoutRebroadcast` is approximation on how long the whole round should take so it can be something like `timeoutPropose(r)+timeoutPrevote(r)+timeoutPrecommit(r)`

_REBROADCAST()_ - broadcasts the current non-nil values of _EnterRoundCertificate_, _lastPrevote_ and _lastPrecommit_ 

Summary of changes as compared to parent branch `anca/gossip`:
- set the timeout value for rebroadcast timer to `timeout_propose + timeout_prevote + timeout_precommit`
- start the rebroadcast timer when entering a new round

The rest of the requirements above are already in the parent branch

---

#### For all contributors

### DIFF
--- a/code/crates/core-consensus/src/handle/liveness.rs
+++ b/code/crates/core-consensus/src/handle/liveness.rs
@@ -98,7 +98,11 @@ pub async fn on_round_certificate<Ctx>(
 where
     Ctx: Context,
 {
-    info!(%certificate.height, %certificate.round, "Received round certificate");
+    info!(
+        %certificate.height,
+        %certificate.round,
+        "Received round certificate"
+    );
 
     if certificate.height != state.height() {
         warn!(
--- a/code/crates/core-consensus/src/handle/liveness.rs
+++ b/code/crates/core-consensus/src/handle/liveness.rs
@@ -143,5 +143,13 @@ where
         }
         apply_driver_input(co, state, metrics, DriverInput::Vote(vote)).await?;
     }
+
+    // Cancel rebroadcast timer
+    // TODO: Should do only if the round certificate is well formed, i.e. either PrecommitAny or SkipRound
+    perform!(
+        co,
+        Effect::CancelTimeout(Timeout::rebroadcast(certificate.round), Default::default())
+    );
+
     Ok(())
 }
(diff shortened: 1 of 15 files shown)


code/crates/core-consensus/src/handle/liveness.rs
```diff
@@ -98,7 +98,11 @@ pub async fn on_round_certificate<Ctx>(
 where
     Ctx: Context,
 {
-    info!(%certificate.height, %certificate.round, "Received round certificate");
+    info!(
```
nenadmilosevic95, line 101: I don't see that we cancel the timeout here?

nenadmilosevic95 (commented):
Looks good, I did the review of this together with #997 because my understanding is that it complements it and that we want to merge it there. I left some comments but two main things are:
- do we cancel the timeout when receiving the round cert?
- do we recalculate timeoutRebroadcast to account for changes/increasing in the values of other timeouts as the round increase?

code/crates/core-consensus/src/handle/liveness.rs
ancazamfir: doh! i forgot to do it here (done in option 1)
1afba3109e2d61dbe5fcda397c68ccbb04593e89

code/crates/test/app/config.toml
```diff
@@ -62,6 +62,10 @@ timeout_commit = "0s"
 # Override with MALACHITE__CONSENSUS__TIMEOUT_STEP env variable
 timeout_step = "2s"
 
+# How long we wait after entering a round before starting the rebroadcast liveness protocol
+# Override with MALACHITE__CONSENSUS__TIMEOUT_REBROADCAST env variable
+timeout_rebroadcast = "5s"
```
nenadmilosevic95, line 67: This is allowing you to override the default which is calculated based on other timeouts?

nenadmilosevic95 (approved):

PR #1044, opened 2025-05-20
test: Fix flaky test

romac:
Closes: #1038 

---

#### For all contributors


romac:
Looks like if a node crashes too early after having started it won't be able to reconnect to the other nodes, so to fix the test I changed it to crash the node at height 3 instead of 1. 

Would be good to figure out the root cause of this though, anyone has an idea?

romac:
Looks like waiting a bit more between restarting also solves it. Not sure which one's the best solution, so I'll just add both tests

Issue #1045, opened 2025-05-20
feat: keep nodes private from each other


faddat:
## solutions

    1. Tor Onion Services via Arti
       • Arti is Mozilla’s 100%-Rust Tor implementation (crates `arti-client`, `arti-socks`, etc.).
       • You can spin up an onion service in-process, advertise its `.onion` address in libp2p identify, and dial/listen over Tor circuits instead of raw TCP.
       • No JVM required, and it plugs cleanly into a custom libp2p `Transport`.
    2. Libp2p Circuit Relay
       • Use libp2p’s built-in relay protocol (`/libp2p/circuit/relay/0.1.0`).
       • Your node only connects to well-known “relay” nodes—peers see the relays’ IPs, not your own.
       • Not an anonymity network (relays see plaintext), but hides your endpoint.
    3. Rust Mix-nets (e.g. Nym)
       • The Nym mix-net client (Rust crate `nym-client`) gives stronger metadata privacy.
       • You’d run a local mix client, create a “gateway” service, then route your libp2p traffic through it.
       • It’s more complex, but fully Rust and purpose-built as a mix network.
    4. SOCKS5 or HTTP(S) Proxies
       • If you control a fleet of reverse-proxies (in the cloud or on VPNs), you can route all P2P traffic through them.
       • Rust crates like `tokio-socks` or `hyper` let you tunnel your libp2p transport over SOCKS/HTTPS proxies.
    5. WireGuard/VPN Overlays
       • Spin up a Rust‐based VPN (e.g. `boringtun` or `wireguard-rs`) between nodes.
       • Your real IP is hidden behind the VPN address space.
    6. I2P

note: I intend to send over a PR for this and flesh out this issue a good deal more.

PS: 

* #1046 
  * implements something that might keep nodes private from each other, but is likely to be closed or refactored.

* The next implementation is likely to be the libp2p circuit relay, because it should be relatively easy to implement and is interesting.  

PPS: This will help me to close this issue:

* https://github.com/faddat/aura/issues/37

malachite is a joy

romac:
Closing this as it is not on our roadmap at the moment, cf. https://github.com/informalsystems/malachite/pull/1046#issuecomment-2900253226

PR #1046, opened 2025-05-20
feat: add support for i2p

faddat:
Closes: #1045

NOTE: this is a very early draft PR.  It may need to be 100% reworked, but I figured it would make sense to show the work.

### Why should a consensus mechanism support anonymized traffic?

* Basically, it's not good if "the man" can scan and then nuke a given network.  Using i2p makes it harder for "the man" to nuke a network by preserving validator privacy.

### What's bad about this implementation?

* It requires the JDK to be installed, and is therefore feature-gated to ensure that it doesn't get in the way of normal day to day rust coding.

### Are there other options available if we want to make networks harder for the man to nuke and preserve validator privacy?

* Yes, but they require evaluation.  This is probably the first of a few PRs that look at ways to improve node operator privacy (thereby strengthening networks).  In the end we probably only need 1-3 really good techniques.  This PR isn't really good yet, so it's a draft :)

### if it isn't finished why's it here?

Because there are many ways to improve operator privacy, and we can find the best one by discussing it.  I'm not really sure this one is best.

---

#### For all contributors


romac:
Thank you for submitting this PR!

While this is a valuable feature concept, we need to be transparent about our current priorities and capacity. At this time, we don't have the bandwidth to incorporate and maintain this functionality within the Malachite codebase. As you may know, Malachite primarily serves the needs of Informal Systems' clients, and our development roadmap reflects those specific requirements.

That said, we believe that this idea has merit and could be developed as a sidecar implementation in a separate repository. Malachite is designed with flexibility in mind, which means your i2p feature could be built as a composable component that works alongside Malachite without requiring changes to the engine, for example as a new type of `Network` actor.

Let us know if you'd like to discuss this approach further. Thanks again for your contribution!

PR #1052, opened 2025-05-22
docs(github): Update contributing guidelines

romac:


adizere (approved):
Thanks Romain!

Nit - might be best to de-duplicate the "scope of contributions" section. For instance, keep only the text in "CONTRIBUTING_CODE.md" and link that section from "CONTRIBUTING.md"

romac:
I did this on purpose to make sure people would not miss it, but I can link to it from one of the two sections if you prefer?

Issue #1056, opened 2025-05-26
code: Discovery issues


bastienfaivre:
1. When receiving a connect request from an already upgraded peer, `handle_connect_request` does not check if the request uses the same connection as the one saved in the upgraded data. This leads to the requesting node thinking that its connection is upgraded, but the other nodes will close it later since it is an ephemeral connection.
2. The `should_dial` method does not detect a local address if the node was initialized with a multiaddr using `0.0.0.0` as an IP address.

PR #1060, opened 2025-05-27
fix(code/discovery): Connection upgrades and multiaddr issues

bastienfaivre:
Closes: #1056

---

#### For all contributors

### DIFF
--- a/code/crates/discovery/README.md
+++ b/code/crates/discovery/README.md
@@ -1,3 +1,7 @@
+# Discovery
+
+## Concept
+
 A node has two connection parameters: outbound and inbound peers. Outbound peers are those that the node actively connects to; inbound peers are those that connect to the node. A peer can be outbound, inbound, or ephemeral—temporary connections used for tasks like Kademlia requests.
 
 A node prioritizes fulfilling its outbound connections. To establish a persistent connection, it sends a connect request. The receiving node assigns the requester as outbound or inbound if there’s capacity; otherwise, it rejects the request.


bastienfaivre:
Discovery now supports all the listening addresses of a node, especially when listening on `0.0.0.0`. Moreover, the connection upgrade issue has been fixed by changing the connection philosophy: we now follow the same principle as in _libp2p_, where there can be multiple connections to a single peer. Therefore, the number of connections is now counted at the peer level, not the connection level. This introduces a new parameter `max_connections_per_peer` in the discovery config.

code/crates/discovery/src/dial.rs
```diff
@@ -0,0 +1,54 @@
...
+                Some(
+                    DialOpts::peer_id(peer_id)
+                        .addresses(self.listen_addrs.clone())
+                        .allocate_new_port()
+                        .build(),
+                )
+            } else {
+                Some(
+                    DialOpts::unknown_peer_id()
+                        .address(addr.clone())
```
romac, line 45: Shouldn't we use all listen addresses here as well?

code/crates/discovery/src/lib.rs
```diff
@@ -54,11 +45,11 @@ where
 
     selector: Box<dyn Selector<C>>,
 
-    bootstrap_nodes: Vec<(Option<PeerId>, Multiaddr)>,
+    bootstrap_nodes: Vec<(Option<PeerId>, Vec<Multiaddr>)>,
     discovered_peers: HashMap<PeerId, identify::Info>,
     active_connections: HashMap<PeerId, Vec<ConnectionId>>,
-    outbound_connections: HashMap<PeerId, OutboundConnection>,
-    inbound_connections: HashMap<PeerId, ConnectionId>,
+    outbound_peers: HashMap<PeerId, bool>, // true if the connection is persistent
```
romac: Let's replace that boolean with an enum to avoid _boolean blindness_ and make the code easier to read.

romac:
Looks great! I just find the change of nomenclature from in/outbound connections to in/bound peers a bit confusing. Can you perhaps shed some light on why you made that change? I understand that connections are now tracked by peer, but what if a peer has both a an inbound and outbound connection?

code/crates/discovery/src/dial.rs
bastienfaivre: There is no such option when the peer ID is unknown: https://github.com/libp2p/rust-libp2p/blob/00e0d29c9e1584655d97a0d96fa26cc336e34698/swarm/src/dial_opts.rs#L284

bastienfaivre:
I agree that the nomenclature is a bit confusing. Would it be ok to write a short README at the root of the crate to explain the system?

romac:
> Would it be ok to write a short README at the root of the crate to explain the system?

Good idea! But maybe start by posting the explanation here and then we can move it to a README

code/crates/discovery/src/dial.rs
romac: Ah I see, I guess it's because if each address advertises a different peer id, libp2p does not know what to do, and therefore only accepts a single address and associate it with the peer id that it advertises from then on.

bastienfaivre:
A node has two connection parameters: outbound and inbound peers. Outbound peers are those that the node actively connects to; inbound peers are those that connect to the node. A peer can be outbound, inbound, or ephemeral—temporary connections used for tasks like Kademlia requests.

A node prioritizes fulfilling its outbound connections. To establish a persistent connection, it sends a connect request. The receiving node assigns the requester as outbound or inbound if there’s capacity; otherwise, it rejects the request.

Once registered as an outbound or inbound peer, a node can open multiple connections, up to the limit set by the `max_connections_per_peer` parameter.

When an outbound connection is dropped, the node attempts to upgrade an existing inbound peer or find a new one (e.g., via Kademlia).

For more, please take a look at Chapter 3 of this [thesis](https://bastienfaivre.com/files/master-thesis.pdf).

@romac, what do you think? Should it be more detailed?

romac:
@bastienfaivre That's perfect, thanks!

PR #1066, opened 2025-06-02
feat(code): Add actor integrated application that uses proposal only

ancazamfir:
Closes: #XXX

---

#### For all contributors

### DIFF
--- a/code/examples/actor/proposal/Notes.md
+++ b/code/examples/actor/proposal/Notes.md
@@ -1,0 +1,572 @@
+# Review
+
+## Main.rs
+
+The `main.rs` file serves as the entry point for the actor application. It handles:
+
+1. Command-line argument parsing and configuration loading
+2. Initialization of the error handling system
+3. Execution of different commands:
+   - `start`: Launches the node with specified configuration
+   - `init`: Initializes a new node with default configuration
+   - `testnet`: Sets up a test network
+   - `distributed-testnet`: Creates a distributed test network
+   - `dump-wal`: Dumps the Write-Ahead Log (WAL) contents
+
+Integrates with the actor system through the `ActorNode` implementation. It supports both file-based and default configurations, and handles the setup of logging, metrics, and runtime environments.
+
+## Detailed Implementation
+
+### Configuration Management
+- Supports multiple configuration sources:
+  - File-based configuration
+  - Default configuration
+  - Value-based configuration
+- Handles home directory and config file paths
+- Manages genesis and private key files
+
+### Command Execution
+1. **Start Command**
+   - Loads node configuration
+   - Initializes logging with specified level and format
+   - Sets up metrics if enabled
+   - Builds and runs the runtime environment
+   - Launches the node with specified start height
+
+2. **Init Command**
+   - Creates initial node configuration
+   - Generates necessary files:
+     - Genesis file
+     - Private validator key file
+     - Configuration file
+
+3. **Testnet Commands**
+   - `testnet`: Creates a local test network
+   - `distributed-testnet`: Sets up a distributed test network across multiple machines
+   - Both commands initialize nodes with default configurations
+
+4. **Dump WAL Command**
+   - Uses ProtobufCodec for WAL operations
+   - Dumps the contents of the Write-Ahead Log
+
+### Runtime Environment
+- Configures logging levels and formats
+- Sets up metrics collection
+- Manages runtime parameters
+- Handles graceful shutdown
+
+## Spawn.rs
+
+The `spawn.rs` file implements the actor spawning and initialization.
+
+### Core Functionality
+
+#### Actor Spawning
+- Manages actor creation and initialization
+- Handles actor configuration
+- Controls actor lifecycle
+- Implements actor linking
+
+#### Actor Types
+1. **Mempool Actors**
+   - `MempoolNetwork`: Network communication for mempool
+   - `Mempool`: Core transaction pool management
+   - `MempoolLoad`: Transaction load generation
+
+2. **Network Actors**
+   - `Network`: P2P networking for consensus
+
+3. **Host Actor**
+   - `Host`: Block creation and proposal handling
+
+4. **Consensus Actors**
+   - `Consensus`: Consensus protocol implementation
+
+5. **Other Actors**
+   - `Wal`: Write-Ahead Log management
+   - `Sync`: Block/ Value synchronization
+
+#### Actor Configuration
+1. **Network Configuration**
+   - Peer discovery settings
+   - Connection parameters
+   - Message handling options
+   - Network topology
+
+2. **Mempool Configuration**
+   - Pool size limits
+   - Transaction batching
+   - Load generation
+   - Network integration
+
+3. **Host Configuration**
+   - Block size limits
+   - Proposal parameters
+   - Storage settings
+   - State management
+
+4. **Consensus Configuration**
+   - Validator set
+   - Initial height
+   - Timeout settings
+   - Threshold
+   - Value payload mode
+   - ..
+
+#### Actor Linking
+- Establishes communication channels
+- Sets up message routing
+- Manages actor dependencies
+- Controls actor hierarchy
+
+## Node.rs
+
+The `node.rs` file implements the core node functionality for the actor-based application. It provides the infrastructure for running and managing nodes in the network.
+
+### Core Components
+
+#### ActorNode Structure
+- Manages node configuration and state
+- Handles home directory and configuration sources
+- Supports custom start heights
+- Implements the Node trait for standard node operations
+
+#### Configuration Management
+- Supports multiple configuration sources:
+  - File-based configuration
+  - Default configuration
+  - Value-based configuration
+- Manages file paths for:
+  - Genesis configuration
+  - Private validator keys
+  - Node configuration
+
+#### Node Implementation
+1. **Configuration Loading**
+   - Loads configuration from specified sources
+   - Handles default configurations
+   - Manages file-based configurations
+
+2. **Key Management**
+   - Handles private key generation
+   - Manages public key derivation
+   - Supports keypair creation
+   - Implements key file operations
+
+3. **Genesis Management**
+   - Loads genesis configuration
+   - Manages validator sets
+   - Handles initial state setup
+
+#### Node Operations
+1. **Start Operation**
+   - Initializes node configuration
+   - Sets up logging and metrics
+   - Creates necessary directories
+   - Spawns node actors
+   - Manages node lifecycle
+
+2. **Node Handle**
+   - Provides event subscription
+   - Manages node lifecycle
+   - Handles graceful shutdown
+   - Controls node state
+
+#### Test Network Support
+1. **Local Testnet**
+   - Creates single-machine test networks
+   - Configures local node instances
+   - Sets up test environment
+
+2. **Distributed Testnet**
+   - Supports multi-machine test networks
+   - Manages distributed configurations
+   - Handles cross-machine communication
+
+#### Actor System
+The node implementation spawns several actors that work together to provide the complete node functionality:
+
+1. **Mempool Actors**
+   - `MempoolNetwork`: Handles P2P communication for mempool
+     - Manages peer connections
+     - Handles message broadcasting
+     - Implements gossip protocol
+     - Dedicated network layer for transaction propagation
+   - `Mempool`: Core mempool functionality
+     - Manages transaction pool
+     - Handles transaction batching
+     - Controls transaction limits
+   - `MempoolLoad`: Simulates transaction load
+     - Generates test transactions
+     - Controls load patterns
+     - Manages transaction flow
+
+2. **Network Actors**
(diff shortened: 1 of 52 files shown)


code/examples/actor/proposal/proto/sync.proto
```diff
@@ -0,0 +1,59 @@
...
+    ValueId value_id = 3;
+    repeated CommitSignature signatures = 4;
+}
+
+message SyncRequest {
+  ValueRequest value_request = 1;
+}
+
+message SyncResponse {
+  ValueResponse value_response = 1;
```
romac, line 48: This field should probably be explicitly optional and handled as such, cf. https://github.com/informalsystems/malachite/issues/1074#issuecomment-2939030159

code/examples/actor/proposal/proto/consensus.proto
```diff
@@ -0,0 +1,57 @@
...
+    Value value = 3;
+    optional uint32 pol_round = 4;
+    Address validator_address = 5;
+}
+
+message ProposalPart {
+    oneof part {
+        ProposalData data = 2;
+    }
+}
```
nenadmilosevic95: I don't understand why we use oneof here?

code/examples/actor/proposal/build.rs
```diff
@@ -0,0 +1,23 @@
...
+    for proto in protos {
+        println!("cargo:rerun-if-changed={proto}");
+    }
+
+    let fds = protox::compile(protos, ["./proto"])?;
+
+    let mut config = prost_build::Config::new();
+    config.bytes(["."]);
+    config.enable_type_names();
+    config.default_package_filename("test");
```
nenadmilosevic95, line 19: Will this line be ignored and anyway all code will be inside `p2p.rs` file?

code/examples/actor/proposal/src/types/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+pub use crate::genesis::*;
+pub use crate::height::*;
+pub use crate::hash::*;
+pub use crate::proposal::*;
+pub use crate::proposal_part::*;
+pub use crate::signing::*;
+pub use crate::validator_set::*;
+pub use crate::value::*;
+pub use crate::vote::*;
+pub use crate::proto::*;
```
nenadmilosevic95: Should we use here `self` instead of `crate` ?

code/examples/actor/proposal/src/types/height.rs
```diff
@@ -0,0 +1,72 @@
...
+        self.0
+    }
+
+    pub fn increment(&self) -> Self {
+        Self(self.0 + 1)
+    }
+
+    pub fn decrement(&self) -> Option<Self> {
+        self.0.checked_sub(1).map(Self)
+    }
```
nenadmilosevic95, line 24: Should we put this inside `impl malachitebft_core_types::Height for Height ` block ?

code/examples/actor/proposal/src/types/value.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+impl Value {
+    pub fn new(value: Block) -> Self {
+        Self { value }
+    }
+
+    pub fn id(&self) -> ValueId {
+        ValueId::new(self.value.block_hash.clone())
+    }
```
nenadmilosevic95, line 76: Why do we define this method two times?

code/examples/actor/proposal/src/types/validator_set.rs
```diff
@@ -0,0 +1,181 @@
...
+        // Sort the validators according to the current Tendermint requirements
+        //
+        // use core::cmp::Reverse;
+        //
+        // (first by validator power, descending, then by address, ascending)
+        // vals.sort_unstable_by(|v1, v2| {
+        //     let a = (Reverse(v1.voting_power), &v1.address);
+        //     let b = (Reverse(v2.voting_power), &v2.address);
+        //     a.cmp(&b)
+        // });
```
nenadmilosevic95, line 133: Why is this code commented?

code/examples/actor/proposal/src/node.rs
```diff
@@ -0,0 +1,489 @@
...
+    address::Address,
+    height::Height,
+    signing::{PrivateKey, PublicKey},
+    validator_set::{Validator, ValidatorSet},
+};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Genesis {
+    pub validator_set: ValidatorSet,
+}
```
nenadmilosevic95, line 30: Isn't this already defined in types/genesis.rs?

code/examples/actor/proposal/src/mempool.rs
```diff
@@ -0,0 +1,267 @@
...
+                    } else {
+                        // No more transactions to reap
+                        break;
+                    }
+                }
+
+                reply.send(txes)?;
+            }
+
+            Msg::Update { tx_hashes } => {
```
nenadmilosevic95: Why is this called update, better to call it prune or something like that?

code/examples/actor/proposal/src/actor.rs
```diff
@@ -0,0 +1,595 @@
...
+    let latest_block_height = state
+        .block_store
+        .max_decided_value_height()
+        .await
+        .unwrap_or_default();
+    let start_height = latest_block_height.increment();
+
+    state.consensus = Some(consensus.clone());
+
+    tokio::time::sleep(Duration::from_millis(200)).await;
```
nenadmilosevic95: why this sleep here?

code/examples/actor/proposal/src/actor.rs
```diff
@@ -0,0 +1,595 @@
...
+    role: Role,
+) -> Result<(), ActorProcessingErr> {
+    state.height = height;
+    state.round = round;
+    state.proposer = Some(proposer);
+    state.role = role;
+
+    // If we have already built or seen one or more values for this height and round,
+    // feed them back to consensus. This may happen when we are restarting after a crash.
+    replay_undecided_values(state, height, round).await?;
```
nenadmilosevic95: What are we replaying here?

code/examples/actor/proposal/src/actor.rs
```diff
@@ -0,0 +1,595 @@
...
+        .await?;
+
+    let proposed_value = values
+        .into_iter()
+        .find(|v| v.proposer == state.host.address);
+
+    Ok(proposed_value)
+}
+
+async fn on_process_synced_value(
```
nenadmilosevic95, line 320: I am confused what is this function used for?

romac:
Moved the application to https://github.com/informalsystems/malachite-apps/pull/2

romac:
Re-instated the app code so that Nenad can post his pending comments and push a few commits

code/examples/actor/proposal/src/metrics.rs
```diff
@@ -0,0 +1,81 @@
...
+            finalized_txes: Counter::default(),
+            block_tx_count: Histogram::new(linear_buckets(0.0, 32.0, 128)),
+            block_size_bytes: Histogram::new(linear_buckets(0.0, 64.0 * 1024.0, 128)),
+        }))
+    }
+
+    pub fn register(registry: &SharedRegistry) -> Self {
+        let metrics = Self::new();
+
+        registry.with_prefix("starknet_app", |registry| {
```
nenadmilosevic95: Should we rename this?

nenadmilosevic95 (commented):
Some minor comments added

code/examples/actor/proposal/src/metrics/db_metrics.rs
adizere, line 1: This file & all references can be deleted - seems to not be used.

code/examples/actor/proposal/src/metrics.rs
adizere: Agree!

adizere: It should be something different, eg `app_actor_proposal`.

code/examples/actor/proposal/Notes.md
adizere, line 1: Not sure in this form these notes are useful. It's a good braindump, but it will become outdated fast, and it seems quite painful to maintain.

code/examples/actor/proposal/src/actor.rs
adizere: This is the context for calling this method:

```
            // It may happen that our node is lagging behind its peers. In that case,
            // a synchronization mechanism will automatically kick to try and catch up to
            // our peers. When that happens, some of these peers will send us decided values
            // for the heights in between the one we are currently at (included) and the one
            // that they are at. When the engine receives such a value, it will forward to the application
            // to decode it from its wire format and send back the decoded value to consensus.
```

ref: from [here](https://github.com/informalsystems/malachite/blob/535d3e420cab5b59afee6a5f39c038a6216a419b/code/crates/test/app/src/app.rs#L249-L254)

adizere: So the method `on_process_synced_value` is basically a way for consensus to tell the application: "hey decode this previously decided value for me"

code/examples/actor/proposal/src/metrics/db_metrics.rs
adizere: deleted in f5dd2ec1

code/examples/actor/proposal/src/metrics.rs
adizere: fixed in 65ac08e4

code/examples/actor/proposal/src/types/lib.rs
adizere: was fixed

adizere:
To discuss sync:

- Most of Nenad's comments
- Can we reuse `malachitebft_test::Height` in this example & more generally reuse other test types so that we deduplicate them
- We should mark the distributed-testnet command as `unimplemented` (as we do in the channel-based app example)
- What do we do with the Notes.md and more generally documentation for this example -- related to being moved to the apps repo

I'll schedule a sync soon.

code/examples/actor/proposal/src/actor.rs
```diff
@@ -0,0 +1,558 @@
...
+                on_get_validator_set(state, height, reply_to).await
+            }
+
+            HostMsg::Decided {
+                certificate,
+                consensus,
+                ..
+            } => on_decided(state, &consensus, certificate, &self.metrics).await,
+
+            HostMsg::GetDecidedValue { height, reply_to } => {
```
adizere, line 160: cc @kirdatatjana @insumity

code/examples/actor/proposal/proto/consensus.proto
ancazamfir: ah, it was a left over when i was doing some experiments (trying to stuff the consensus message in the part so we don't need a new host msg)...removed now in 399b5a1a78bdedfb48bf88a3b0a16eaf804ff66b

code/examples/actor/proposal/Notes.md
ancazamfir: Oh we can remove it, came from a copy of the example app we gave to LeftCurve and we used these notes in a few meetings with them to discuss the integration points.

ancazamfir: 806021a1659f851c458ad95fb29e87b62b95d04e

code/examples/actor/proposal/src/node.rs
ancazamfir: removed in 53bb59ebd3c870272ff4ac86e8db383299fff1ce

code/examples/actor/proposal/src/mempool.rs
ancazamfir: 05953661d1a7db2438bbb2b3dc24ab353dfdbfab

code/examples/actor/proposal/src/actor.rs
ancazamfir: is see it's removed. We do this sleep in all apps, @romac mentioned something around connection errors. We should probably investigate and clean the others if not required.

code/examples/actor/proposal/src/types/validator_set.rs
ancazamfir: Not really required for now, we keep the genesis order and the proposer selection ignores the voting power. But we can uncomment also.

code/examples/actor/proposal/build.rs
ancazamfir: i see you changed already.

code/examples/actor/proposal/src/types/height.rs
ancazamfir: Yes we could. Maybe we open an issue to discuss and solve across all apps. If we move all methods out of the concrete implementation then the app code will have to do calls like:
```
use malachitebft_core_types::CommitCertificate, Height as HeightTrait;

let start_height = HeightTrait::increment(&latest_block_height);
```

btw, the Height trait implements `increment()` and `decrement()` already.

ancazamfir: I made some changes in db158edf76c7ec79ce8c2e26adcd1682916470cf so you see how it would look.

PR #1071, opened 2025-06-03
feat(code/sync): Add peer scoring system 

romac:
Closes: #1072

This PR adds a peer scoring system to the `sync` module with a customizable scoring strategy.

Strategies must implement the `ScoringStrategy` trait:

```rust
/// Strategy for scoring peers based on sync results
pub trait ScoringStrategy: Send + Sync {
    /// Initial score for the given peer
    fn initial_score(&self, peer_id: PeerId) -> Score;

    /// Update the score based on previous score and sync result
    fn update_score(&mut self, previous_score: Score, result: SyncResult) -> Score;
}
```

where `SyncResult` is defined as:

```rust
/// Result of a sync request to a peer
#[derive(Copy, Clone, Debug)]
pub enum SyncResult {
    /// Successful response with given response time
    Success(Duration),

    /// Timeout response
    Timeout,

    /// Failed response
    Failure,
}
```

The default (and currently only) strategy implements an _exponential moving average with variable learning rates_ that rewards fast responses to requests, penalize slow responses a little, and strongly penalize timeouts and failures (i.e. empty responses).

```rust
/// Exponential Moving Average scoring strategy
#[derive(Copy, Clone, Debug)]
pub struct ExponentialMovingAverage {
    /// Learning rate for successful responses
    pub alpha_success: f64,

    /// Learning rate for timeouts
    pub alpha_timeout: f64,

    /// Learning rate for failures
    pub alpha_failure: f64,

    /// Threshold for slow responses.
    ///
    /// This should typically be based on the block time,
    /// as we do not want responses that are slower than the
    /// expected block time to be considered successful otherwise
    /// a node might not be able to keep up with the network.
    pub slow_threshold: Duration,
}
```

These parameters can be tweaked. Their default values are currently:

| Parameter | Default value |
|--------|--------|
| `alpha_success` | 0.20 |
| `alpha_timeout` | 0.10 |
| `alpha_failure` | 0.15 |
| `slow_threshold` | 1 second | 

The slow threshold should be chosen to be lower than the expected block time.

`sync::State::new` now takes a `ScoringStrategy` which allows users of the library to customize the scoring strategy according to their needs, as well as a parameter to enable pruning peers scores after they have not been active for the given duration. This allows to reset the score of peers that have not been picked in a long time.

```rust
impl<Ctx> State<Ctx>
where
    Ctx: Context,
{
    pub fn new(
        // Random number generator for selecting peers
        rng: Box<dyn rand::RngCore + Send>,
        // Strategy for scoring peers based on their performance
        scoring_strategy: impl ScoringStrategy + 'static,
        // Threshold for considering a peer inactive, and their score reset to the initial value
        inactive_threshold: Option<Duration>,
    ) -> Self
```

---

#### For all contributors


code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,618 @@
...
+    }
+
+    /// Reset the score of peers that haven't been updated for the specified duration
+    pub fn prune_inactive_peers(&mut self, inactive_threshold: Duration) {
+        let now = Instant::now();
+
+        self.scores
+            .retain(|_, score| now.duration_since(score.last_update) < inactive_threshold);
+    }
+}
```
nenadmilosevic95: This was a bit confusing for me because the method is `prune` and the comment is `reset` and I see that both are correct because in the code we actually prune but in reality the effect is that the score will be reset to the initial value. IDK if we should address this by renaming somehow

code/crates/sync/src/handle.rs
```diff
@@ -132,6 +133,13 @@ where
 
     perform!(co, Effect::BroadcastStatus(state.tip_height));
 
+    if let Some(inactive_threshold) = state.inactive_threshold {
+        // If we are at or above the inactive threshold, we can prune inactive peers.
```
nenadmilosevic95, line 106: Do we prune them or just reset their scores? It is a bit ambiguous here

code/crates/sync/src/handle.rs
```diff
@@ -238,17 +246,30 @@ pub async fn on_value_response<Ctx>(
...
+    debug!(%response.height, %request_id, %peer_id, "Received response");
 
     state.remove_pending_decided_value_request(response.height);
 
-    metrics.decided_value_response_received(response.height.as_u64());
+    let response_time = metrics.value_response_received(response.height.as_u64());
+
+    if let Some(response_time) = response_time {
+        let sync_result = if response.value.is_none() {
+            SyncResult::Failure
```
nenadmilosevic95, line 237: This failure means that the request want this response does not exist? At this point we already checked that the response is valid (well-formed etc.)?

code/crates/sync/src/handle.rs
```diff
@@ -306,8 +327,11 @@ where
         Request::ValueRequest(value_request) => {
             let height = value_request.height;
             warn!(%peer_id, %height, "Value request timed out");
+
             state.remove_pending_decided_value_request(height);
-            metrics.decided_value_request_timed_out(height.as_u64());
+            metrics.value_request_timed_out(height.as_u64());
+
+            state.peer_scorer.update_score(peer_id, SyncResult::Timeout);
```
nenadmilosevic95, line 331: In case the response was slower than the timeout but it arrives, are we going to update_score for this peer two times:
1. when timeout expire we will call `update_score(..., SyncResult::Timeout)`
2. when it arrives we will call `update_score(..., SyncResult::Failure)` 
?

code/crates/sync/src/metrics.rs
```diff
@@ -30,8 +28,8 @@ pub struct Inner {
     server_latency: Histogram,
     request_timeouts: Counter,
 
-    instant_request_sent: Arc<DashMap<(u64, i64), Instant>>,
-    instant_request_received: Arc<DashMap<(u64, i64), Instant>>,
+    instant_request_sent: Arc<DashMap<u64, Instant>>,
+    instant_request_received: Arc<DashMap<u64, Instant>>,
```
nenadmilosevic95, line 32: Just to check my understanding here, these two are needed to calculate `client_latency` and `server_latency` and since we calculate client_latency one request at a time we don't need a tuple and also for `server_latency` since we process one request at a time we don't need a tuple? 

Apart from this you just did some renaming in this file?

code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,618 @@
...
+
+pub type Score = f64;
+
+/// Strategy for scoring peers based on sync results
+pub trait ScoringStrategy: Send + Sync {
+    /// Initial score for new peers
+    fn initial_score(&self, peer_id: PeerId) -> Score;
+
+    /// Update the score based on previous score and sync result
+    fn updated_score(&self, previous_score: Score, result: SyncResult) -> Score;
```
nenadmilosevic95: I am in doubt should we call this update_score or updated_score

code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,618 @@
...
+        peer_score.score = new_score;
+        peer_score.last_update = Instant::now();
+    }
+
+    /// Get the current score for a peer
+    pub fn get_score(&self, peer_id: &PeerId) -> Score {
+        self.scores
+            .get(peer_id)
+            .map(|p| p.score)
+            .unwrap_or(self.strategy.initial_score(*peer_id))
```
nenadmilosevic95, line 117: Ok this is ensuring that even if remove the peer from peer_scorer field we will return initial score for it?

code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,618 @@
...
+    }
+}
+
+/// Tracks peer scores using a scoring strategy
+pub struct PeerScorer {
+    scores: HashMap<PeerId, PeerScore>,
+    strategy: Box<dyn ScoringStrategy>,
+}
+
+impl PeerScorer {
```
nenadmilosevic95, line 67: To sum up my understanding of peer scorer:
- we keep a score for each peer
- we add an entry for a peer first time we call update for this peer
- we remove an entry for a peer in prune method if it is not active for a predefined time
- if we don't have a score for a peer yet we will consider that this peer has `initial_score`, and if the current score is negative we will set it to 0

code/crates/sync/src/handle.rs
nenadmilosevic95: In general, I am bit in doubt about this `inactive_threshold` mechanism because a node updates peer's activity when the peer responds to its request but what if a peer don't send request to this peer for some time, does this mean that the peer is inactive? I guess no..

code/crates/sync/src/scoring/ema.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    fn updated_score(&self, previous_score: Score, result: SyncResult) -> Score {
+        match result {
+            SyncResult::Success(response_time) => {
+                // Calculate quality score between 0-1 based on response time
+                // We use an exponential decay function to penalize slow responses more heavily
+                let quality = if response_time < self.slow_threshold {
+                    // Fast responses get a high quality score
+                    1.0
```
nenadmilosevic95, line 85: Ok, so this part we need to change if we want to award faster peers. IDK but for me personally maybe the best thing would be to not use slow_threshold here but rather just rank them based on the how fast the response was because anyway you will not always pick the best one?

code/crates/sync/src/scoring/ema.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Learning rate for failures
+    pub alpha_failure: f64,
+
+    /// Threshold for slow responses.
+    ///
+    /// This should typically be based on the block time,
+    /// as we do not want responses that are slower than the
+    /// expected block time to be considered successful otherwise
+    /// a node might not be able to keep up with the network.
+    pub slow_threshold: Duration,
```
nenadmilosevic95, line 26: It is interesting for me the relation between the timeout we use in sync to say that we don't wait for a response any longer and this one. I guess the one in sync is much larger than this one?

code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,618 @@
...
+    /// Select a peer using weighted probabilistic selection
+    pub fn select_peer<R: Rng>(&self, peers: &[PeerId], rng: &mut R) -> Option<PeerId> {
+        if peers.is_empty() {
+            return None;
+        }
+
+        let scores = peers.iter().map(|id| self.get_score(id).max(0.0));
+
+        // Sample from peers using a weighted distribution based on their scores
+        let distr = WeightedIndex::new(scores).ok()?;
```
nenadmilosevic95, line 134: We are creating a WeighedIndex every time we want to select a peer, I guess this means that there is a higher probability we will pick the peer with the highest score more freq, but I see why it is like this because the scores are changing between two calls of select_peer. IDK if we can do something about this

nenadmilosevic95 (commented):
@romac I did a full pass and left comments. 
The parts I didn't mange to cover today:
- tests in 'scoring.rs' file
- I didn't go deep in understanding the logic of updating scores in 'ema.rs'
I can do these two tomorrow!

code/crates/sync/src/scoring.rs
romac: Yeah I wasn't sure about that one. I think pruning is the better name as it also alludes to the fact that the we remove these peers from the storage, which may be important if there are lots of peers and we want to keep memory usage low. How about I update the comment to? _Prune peers whose scores have not been updated for the specified duration, effectively resetting their score to the initial score. A peer might be inactive because they were not picked for a long time due to their score being very low. Resetting their score gives them a chance to participate again._

code/crates/sync/src/handle.rs
romac: > if a peer don't send request to this peer for some time, does this mean that the peer is inactive?

Yes that's the definition of inactivity I am using here, cf. my comment above. Does that make sense?

code/crates/sync/src/handle.rs
romac: Yeah this means that the peer did not have the value for the requested height even though it advertised differently. But yes, I need to handle better the case where the response is not well-formed, see https://github.com/informalsystems/malachite/issues/1074#issuecomment-2940681110

code/crates/sync/src/handle.rs
romac: When it arrives it will depend on whether it contains an empty value (`Failure`) or the requested value (`Success` but deemed slow). But you are right that we are updating the score twice here. 

Not sure what to do when the response still arrives after the timeout:
(a) We just drop it and do not update the score a second time
(b) We still process it but do not update the score a second time
(c) We still process it and we update the score a second time

code/crates/sync/src/metrics.rs
romac: > Apart from this you just did some renaming in this file?

Yes just renaming

> since we process one request at a time we don't need a tuple?

Yes, not sure why we even had a tuple in the first place, its second component was always set to `-1`

romac: But perhaps it would be more resilient to track requests with their request id

code/crates/sync/src/scoring.rs
romac: `update` sounds imperative to me, like we are mutating some state here which is not the case, while `updated` sounds more functional. But I agree that `update_score` reads better. 

Now that I think about it, some strategies may need to track some state within the strategy, so `update_score` might be better regardless. I will also change the self type to `&mut` to allow strategies to keep and mutate internal state.

code/crates/sync/src/scoring.rs
romac: Yes

code/crates/sync/src/scoring.rs
romac: Exactly. We have to enforce non-negative scores because the weighted sampling requires non-negative weights.

romac: But maybe we should just renormalize the score into a weight in the range `0.0..=1.0` instead of clamping it.

code/crates/sync/src/scoring/ema.rs
romac: For now we agreed with Farcaster that we would reward all _fast enough_ peers the same.

code/crates/sync/src/scoring/ema.rs
romac: Yes that's the expectation, will clarify in the doc comment.

code/crates/sync/src/scoring.rs
romac: We could try keeping the distribution around and use [`WeightedIndex::update_weights`](https://docs.rs/rand/latest/rand/distr/weighted/struct.WeightedIndex.html#method.update_weights) but that would require more bookkeeping and it does not work if the number of weights (so peers) changes in between two calls. So I think it's better to stick to this for now.

code/crates/sync/src/scoring.rs
nenadmilosevic95: Yes that is good. Just note that by resetting the score in theory we can also reduce the score of a peer.

code/crates/sync/src/handle.rs
nenadmilosevic95: So with this mechanism we want to help peers with low score (< initial score) and give them second chance? Do we want to do this also with peers that are faulty? 

Does it make sense that we only update the scores of peers that has score lower than initial score? This way we add a safeguard to not lower the score of peers that were responsive but just were not picked in a while (btw I am not sure if in practice this can happen since responsive peers are awarded the same)

code/crates/sync/src/handle.rs
nenadmilosevic95: So let's see we have three possibilities:
1) timeout + failure 
2) timeout + success 
3) timeout + the message never arrives
In 1) and 2) we will basically punish the node twice while in 3) we will punish the node only once. So it seems that non responding node will be better off compared to 1) and 2) which is not fair with respect to situation 2). 

So one alternative would be to consider all 3 cases the same, namely explicitly say that timeout should be very conservative and then when it expires and a node did not receive a response we treat it as failure, and if we receive a response after that we just ignore.  So I guess this is the option (a) or (b)? Maybe (b) makes sense if the answer is correct, however, if we do (a) then we can immediately send a new request when the timeout expires, as we don't do this now?

code/crates/sync/src/metrics.rs
nenadmilosevic95: Maybe we will need this in the future when if we allow multiple parallel requests

code/crates/sync/src/scoring.rs
nenadmilosevic95: I guess both are fine

code/crates/sync/src/scoring.rs
romac: Right, will add that to the comment as well

code/crates/sync/src/metrics.rs
romac: Right, but I think even with batching we do not want to have concurrent requests, is that right @bastienfaivre @hvanz?

code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,625 @@
...
+        f: impl Fn(&mut Unstructured) -> Result<T>,
+        size: RangeInclusive<usize>,
+    ) -> Result<Vec<T>> {
+        let size = u.int_in_range(size)?;
+        (0..size).map(|_| f(u)).collect::<Result<Vec<T>>>()
+    }
+
+    // Property: Scores are bounded between 0.0 and 1.0
+    #[test]
+    fn scores_are_bounded() {
```
nenadmilosevic95, line 244: Maybe to make sure that this indeed checked we should just apply Successes to try to go over 1 and just apply Failures to go below 0?

code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,625 @@
...
+
+            let initial_score = scorer.get_score(&peer_id);
+            let update_score = scorer
+                .strategy
+                .update_score(initial_score, SyncResult::Success(response_time));
+
+            // Success should never decrease score below a reasonable threshold
+            // (allowing for very slow responses that might slightly decrease score)
+            assert!(
+                update_score >= initial_score * 0.9,
```
nenadmilosevic95: I don't understand this, why would SyncResult::Success(response_time) decrease the score?

code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,625 @@
...
+            )?;
+
+            let bad_results = arb_vec(
+                u,
+                |u| u.choose_iter([SyncResult::Timeout, SyncResult::Failure]),
+                5..=15,
+            )?;
+
+            let good_peer = PeerId::random();
+            let bad_peer = PeerId::random();
```
nenadmilosevic95, line 464: Should we assert they are not the same?

code/crates/sync/src/handle.rs
romac: > Does it make sense that we only update the scores of peers that has score lower than initial score?  This way we add a safeguard to not lower the score of peers that were responsive but just were not picked in a while

I think there is a way to get into a situation where there is a peer that has score > 0.5 but still lower than all other peers (maybe because it was slow at first and then sped up a bit). In this case we would not reset that peer's score and therefore won't help it be picked again now that it's fast.

So I think we should leave it at that for now, what do you think?

code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,625 @@
...
+
+            // Score for peer2 should remain unchanged
+            let final_score_peer2 = scorer.get_score(&peer2);
+            assert_eq!(initial_score_peer2, final_score_peer2);
+
+            Ok(())
+        });
+    }
+
+    // Property: Fast responses help a peer recover more quickly than timeouts penalize it
```
nenadmilosevic95, line 673: I am not sure if this test is checking this, it seems to me that it just checks that Success increases the score?

code/crates/sync/src/handle.rs
nenadmilosevic95: But I don't see how we will help this peer since we will reset its score to 0.5 and before it was > 0.5?

code/crates/sync/src/handle.rs
romac: Opened an issue + PR for this: https://github.com/informalsystems/malachite/pull/1078

code/crates/sync/src/scoring.rs
romac: 1b4a2eb854fab22db10202a7ab2940ba4d3864fa

romac: Good idea!

code/crates/sync/src/scoring.rs
romac: Ah that's from before when all responses were graded based on their response time, not just slow ones. Will adapt the test, good catch!

romac: c732a82a

romac:
## TODO

- [x] Address Nenad's remaining comments
- [x] Add metrics for peer scoring

code/crates/sync/src/scoring.rs
romac: You are right! Fixed in 2a44cbef26dcb7d2211cf550dc49b7ae113c4530

code/crates/sync/src/metrics.rs
bastienfaivre: Indeed, we don't have concurrent requests with the first design. We will need to investigate later if this can bring any benefits.

code/crates/sync/src/handle.rs
romac: Duh, you are right, that won't help at all.

code/crates/sync/src/scoring/ema.rs
romac: That said, it is worth noting that a response time just under the threshold gets a quality of `1.0`, while a time exactly at the threshold gets `exp(-1) ≈ 0.367` which creates a discontinuity in the scoring function.

code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,731 @@
...
+
+    /// Failed response
+    Failure,
+}
+
+pub type Score = f64;
+
+/// Strategy for scoring peers based on sync results
+pub trait ScoringStrategy: Send + Sync {
+    /// Initial score for new peers
```
romac:
```suggestion
    /// Initial score for new peers.
    ///
    /// ## Important
    /// The initial score MUST be in the `0.0..=1.0` range.
```

code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,731 @@
...
+    fn initial_score(&self, peer_id: PeerId) -> Score;
+
+    /// Update the peer score based on previous score and sync result
```
romac, line 39:
```suggestion
    /// Update the peer score based on previous score and sync result
    ///
    /// ## Important
    /// The updated score must be in the `0.0..=1.0` range.
```

ancazamfir:
I did a high level review of the PR, looks good. Should we have a way to disable scoring and fall back to the old behavior? Maybe useful if something goes wrong, bugs in scoring (can't see any), maybe a peer ends up being hammered with requests, etc. I am not sure myself, just a thought.

code/crates/sync/src/scoring/ema.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn update_score(&mut self, previous_score: Score, result: SyncResult) -> Score {
+        match result {
+            SyncResult::Success(response_time) => {
+                // Calculate quality score between 0-1 based on response time
+                // We use an exponential decay function to penalize slow responses more heavily
+                let quality = if response_time < self.slow_threshold {
+                    // Fast responses get a high quality score
+                    1.0
+                } else {
+                    // Slow responses get a low quality score based on slow they were
```
hvanz:
```suggestion
                    // Slow responses get a low quality score based on how slow they were
```

code/crates/sync/src/scoring/ema.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    /// Learning rate for timeouts
+    pub alpha_timeout: f64,
+
+    /// Learning rate for failures
+    pub alpha_failure: f64,
+
+    /// Threshold for slow responses.
+    ///
+    /// This should typically be smaller than both the expected
```
hvanz, line 21: Should we validate this value, so we could at least to emit a warning? I'm not sure though where is best to check for that, because this condition is specific to this strategy, then we would need to add block time and sync request timeout to the strategy constructor.

code/crates/sync/src/scoring.rs
```diff
@@ -0,0 +1,731 @@
...
+
+    /// Prune peers whose scores have not been updated for the specified duration,
+    /// effectively resetting their score to the initial score.
+    ///
+    /// A peer might be inactive because they were not picked for a long time
+    /// due to their score being very low. Resetting their score gives them a chance to participate again.
+    ///
+    /// Note that by resetting the score we can also reduce the score of a peer,
+    /// if the peer had a high score but was inactive for a long time.
+    pub fn prune_inactive_peers(&mut self, inactive_threshold: Duration) {
```
hvanz: The word prune makes me think of cutting the network connection. How about `reset_inactive_peers`?

code/crates/sync/src/scoring/ema.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub slow_threshold: Duration,
+}
+
+impl Default for ExponentialMovingAverage {
+    fn default() -> Self {
+        Self::new(
+            0.2,                    // Success
+            0.1,                    // Timeout
+            0.15,                   // Failure
+            Duration::from_secs(1), // Slow threshold
```
hvanz, line 35: Isn't one second too little to reset the scores? We would be giving bad peers another chance too fast I think. 

Of course this will also depend on the application. For instance when values are too big, they will take more time to transmit, so the app developer should take this into account (though then the block time and sync timeout will also be bigger).

A slightly different strategy would be to reset scores only when there are not enough good peers available. (No need to address this in this PR.)

code/crates/sync/src/handle.rs
hvanz: I think it's ok to punish the node twice on case 1, for being late and for giving a bad result.

hvanz (approved):
Great work! 🚀 Eventually we may need to tweak the strategy or the default values but it's a very good first implementation. I suggest to also test it on scale (~100 nodes) with perturbations (restarting nodes) while emulating slow connections.

code/crates/sync/src/scoring/ema.rs
romac: This is not the inactivity threshold but the threshold for a response being deemed slow.

code/crates/sync/src/scoring.rs
romac: Good point!

code/crates/sync/src/scoring/ema.rs
romac: I'd rather not mix too many concerns into the scoring strategy, and instead leave to users of the library to read the docs and configure this properly. We can't prevent every type of mistake anyway since they might e.g. give us an incorrect value for the block time.

romac:
> Should we have a way to disable scoring and fall back to the old behavior? Maybe useful if something goes wrong, bugs in scoring (can't see any), maybe a peer ends up being hammered with requests, etc. I am not sure myself, just a thought.

Good idea, I will add a config setting for it.

PR #1078, opened 2025-06-05
fix(code/sync): Process sync responses even when they fail to decode

romac:
Closes: https://github.com/informalsystems/malachite/issues/1079

This way we do not have to wait until it times out to issue another request for that height.

This PR is best reviewed commit by commit.

---

#### For all contributors

### DIFF
--- a/code/crates/starknet/host/src/actor.rs
+++ b/code/crates/starknet/host/src/actor.rs
@@ -198,7 +198,7 @@ impl Host {
             } => on_decided(state, &consensus, &self.mempool, certificate, &self.metrics).await,
 
             HostMsg::GetDecidedValue { height, reply_to } => {
-                on_get_decided_block(height, state, reply_to).await
+                on_get_decided_value(height, state, reply_to).await
             }
 
             HostMsg::ProcessSyncedValue {
@@ -507,7 +507,7 @@ fn on_process_synced_value(
     Ok(())
 }
 
-async fn on_get_decided_block(
+async fn on_get_decided_value(
     height: Height,
     state: &mut HostState,
     reply_to: RpcReplyPort<Option<RawDecidedValue<MockContext>>>,


code/crates/sync/src/handle.rs
```diff
@@ -379,7 +379,6 @@ where
         state.store_pending_decided_value_request(height, request_id);
     } else {
         warn!(height.sync = %height, %peer, "Failed to send value request to peer");
-        request_value_from_peer_except(co, state, metrics, height, peer).await?;
```
ancazamfir: Why do we wait for next status?

romac: Good question, I should have clarified in a comment. My reasoning was that, if our node can't send a request at some point, it might stay that way for a while eg. because the node lost network for a while. Then if we try again straight away, we'll end up looping over and over again trying to send requests, failing, trying again, etc. I figured since it's probably pretty rare that an otherwise functioning node is unable to send requests, we might as well wait until we get another update to try again. Alternatively we could wait for the next tick instead, or add delay before retrying, or add a counter after which we stop retrying until the next update. What do you think?

ancazamfir: I see, thanks for the explanation. I thought this can fail only if something is wrong with the peer or the connsection/channel to the peer. Let's keep it the way it is for now.

ancazamfir (approved):
🚀

code/crates/sync/src/effect.rs
```diff
@@ -0,0 +1,102 @@
...
+use malachitebft_core_types::Context;
+use malachitebft_peer::PeerId;
+use thiserror::Error;
+
+use crate::{InboundRequestId, OutboundRequestId, ValueRequest, ValueResponse};
+
+/// Provides a way to construct the appropriate [`Resume`] value to
+/// resume execution after handling an [`Effect`].
+///
+/// Eeach `Effect` embeds a value that implements [`Resumable`]
```
hvanz:
```suggestion
/// Each `Effect` embeds a value that implements [`Resumable`]
```

hvanz (approved):
Looks good to me

code/crates/sync/src/state.rs
```diff
@@ -82,15 +86,47 @@ where
...
+
+            // If there are no more requests for this height, remove the entry
+            if request_ids.is_empty() {
+                self.pending_decided_value_requests.remove(&height);
+            }
+        }
     }
 
     pub fn has_pending_decided_value_request(&self, height: &Ctx::Height) -> bool {
-        self.pending_decided_value_requests.contains_key(height)
```
nenadmilosevic95, line 94: Minor thing, would also this old check work here because we anyway remove the entry if there is no ids for this height?

code/crates/sync/src/handle.rs
```diff
@@ -246,14 +218,31 @@ where
...
+    _metrics: &Metrics,
+    request_id: OutboundRequestId,
+    peer: PeerId,
+) -> Result<(), Error<Ctx>>
+where
+    Ctx: Context,
+{
+    debug!(%request_id, %peer, "Received empty response");
+
+    state.remove_pending_decided_value_request_by_id(&request_id);
```
nenadmilosevic95, line 240: Why don't we send request here again the same way we do when we receive invalid certificate?

code/crates/sync/src/handle.rs
```diff
@@ -306,14 +299,34 @@ where
...
+    certificate: CommitCertificate<Ctx>,
+    error: CertificateError<Ctx>,
+) -> Result<(), Error<Ctx>>
+where
+    Ctx: Context,
+{
+    error!(%error, %certificate.height, %certificate.round, "Received invalid certificate");
+    trace!("Certificate: {certificate:#?}");
+
+    state.remove_pending_decided_value_request_by_height(&certificate.height);
```
nenadmilosevic95, line 325: why do we remove here by_height and in on_empty_request we remove by_id?

nenadmilosevic95 (commented):
LGTM! I left few comments, I haven't realized that the PR is already merged :D

code/crates/sync/src/handle.rs
romac: Will fix it but realized we should be careful about the rate at which we try to send another requests for the case where none of our peers have the value and we don't want to spam our whole network.

code/crates/sync/src/handle.rs
romac: Because when the response is empty we don't know what height it was for and have to gather that info by request id instead.

Issue #1070, opened 2025-06-03
feat: Batch value sync requests


hvanz:
Currently syncing a value happens one block each time.

bastienfaivre:
```mermaid
sequenceDiagram
    box Syncing Node
        participant A1 as Application
        participant C1 as Consensus
        participant S1 as Sync
    end

    box Helping Peer
        participant S2 as Sync
        participant C2 as Consensus
        participant A2 as Application
    end

    Note over A1,C1: Normal Operation
    A1->>C1: StartHeight(height)
    C1->>S1: StartedHeight(height)
    S1->>S2: BroadcastStatus(height)

    Note over A1,C1: Node Falls Behind
    S2->>S1: Status(peer_height)
    S1->>S2: GetValuesRequest(from_height, to_height, max_size)
    S2->>A2: GetDecidedValues(from_height, to_height)
    A2->>S2: DecidedValues(ord_map(height -> (value, certificate)))
    S2->>S1: ValuesResponse(ord_map(height -> (value, certificate)))
    S2->>C1: ValuesResponse(ord_map(height -> (value, certificate)))

    Note over A1,S1: For each height's value, do:

    C1->>C1: ProcessCommitCertificate(certificate)
    C1->>A1: ProcessValue(height, value)
    A1->>C1: ProposedValue(height, value, validity)
    C1->>A1: Decided(height, certificate)
    C1->>S1: Decided(height)

    Note over A1,S1: End for
```

Issues:

- Suppose we request values for heights 1-20 from another peer, and in the response, all values are correct, except the one for height 1 (or any technically any height <20). This prevents the consensus actor from making progress. So, we would need to request this value again (potentially from another peer). The problem is that the sync actor has no way of knowing whether the remaining values (2-20) are correct. It would be highly inefficient to re-request all these values again. Moreover, are these other values cached by the consensus actor (replay values)?
- The last point is crucial since batch processing might accentuate the unintuitive behaviour observed with Snapchain nodes: what should the sync actor do while waiting for a sync response to be fully processed?

adizere:
> Suppose we request values for heights 1-20 from another peer, and in the response, all values are correct, except the one for height 1 (or any technically any height <20). This prevents the consensus actor from making progress. So, we would need to request this value again (potentially from another peer). The problem is that the sync actor has no way of knowing whether the remaining values (2-20) are correct. It would be highly inefficient to re-request all these values again. Moreover, are these other values cached by the consensus actor (replay values)?

[...]

PR #1082, opened 2025-06-10
feat(code/sync): Batch sync

bastienfaivre:
Closes: #1070 

---

#### For all contributors


bastienfaivre:
@hvanz I added a few `TODO(SYNC)` where most of the work should be done. Moreover, please double-check the `test/proto/sync.proto`, `test/src/codec/json/raw.rs`, and `test/src/codec/proto/mod.rs` files, as I am unsure about the changes I made.

code/crates/engine/src/host.rs
```diff
@@ -108,6 +108,12 @@ pub enum HostMsg<Ctx: Context> {
         reply_to: RpcReplyPort<Option<RawDecidedValue<Ctx>>>,
     },
 
+    // Retrieve decided values in a batch from the block store
+    GetDecidedValues {
+        range: RangeInclusive<Ctx::Height>,
+        reply_to: RpcReplyPort<BTreeMap<Ctx::Height, Option<RawDecidedValue<Ctx>>>>,
```
ancazamfir: Why not `Vec`? Also in `AppMsg::GetDecidedValues`?

code/crates/starknet/host/src/actor.rs
```diff
@@ -542,6 +548,42 @@ async fn on_get_decided_value(
...
+    loop {
+        match state.block_store.get(height).await {
+            Ok(Some(block)) => {
+                let block = RawDecidedValue {
+                    value_bytes: block.block.to_bytes().unwrap(),
+                    certificate: block.certificate,
+                };
+                values.insert(height, Some(block));
+            }
+            Ok(None) => continue,
```
ancazamfir: Shouldn't we `break` here? There is no point in sending values with some missing, the requester will/ should discard.

ancazamfir:
Browsed through the changes quickly, I realize you may have discussed this already but here are some comments and questions:
- why do we need two protocol versions? Why can't we add optional range values in request/ response. We could add an optional `end_height` in the request, if present the requester (syncing node) wants a range, otherwise a single value. Same for response, add an optional set of values, if not present the helper node is old version or doesn't have the values.
- we should code the helper node to only return contiguous values, i.e. no missing blocks.
- we should add check in helper node to not go over max payload for the response (`max_response_size` not used currently).
- the syncing node should drop if missing values
- if the syncing node asks for blocks `[n, N]` it should accept responses with `[n, K]` with `K<=N` as long as all `K-n+1` are present

ancazamfir:
Another thought, use the existing request/ response structures and mechanism but send also request for `sync_height + 1`. This way, when we move to a new `sync_height` the request for it is pending (and likely already in the consensus buffered messages) and we just send a request for `sync_height + 1` to have it ready for next height. Can also increase the number of requests to have `N` ahead of sync height.

ancazamfir:
When processes sync responses (either individual heights from a batch [here](https://github.com/informalsystems/malachite/blob/f3e9896ac9f0bd22d510cf74462c73232cd0857e/code/crates/engine/src/consensus.rs#L520) or future heights if we go with the sync ahead approach) we need to buffer the responses. With current implementations of [push](https://github.com/informalsystems/malachite/blob/621af93c35cdb88ce472dd43616ef456a8bd6a2f/code/crates/core-consensus/src/util/max_queue.rs#L62) they are very likely to be discarded because we always have messages from the height at which the synced validators are.

code/crates/engine/src/host.rs
ancazamfir: Also, can we just use the existing `GetDecidedValue` multiple times? This way the app doesn't need to change. There is a performance impact but should be minimal. At the helper node side we should bound the range of the response as a syncing (maybe byzantine) peer may ask for a big range.

bastienfaivre: The use of `RangeInclusive` was suggested by @romac to avoid having two variables: `from` and `to`.
For the `BTreeMap`, it is used to efficiently consider and read the values, as we will likely do so based on the height order.
Regarding using `GetDecidedValue` multiple times, it is possible. I believe the performance impact will be primarily at the application level, depending on the actions required to retrieve a value (such as a database query or similar). I agree that not having to change the application is a desired approach, but I don't like the idea of generating lots of events for something that could be achieved in a single shot. Moreover, the application would then also be able to optimize the retrieval of these values.
And so, continuing the reasoning, I would potentially consider having a single `GetDecidedValues` app message that asks the application for a range (or vector?) of values, even if it is used to retrieve a single value sometimes.

@romac, what do you think?

> peer may ask for a big range

There are quite a lot of security concerns to think about here, indeed.

bastienfaivre:
> why do we need two protocol versions?

C.f. Slack, for backward compatibility.

> we should code the helper node to only return contiguous values, i.e. no missing blocks.

Agree.

> we should add check in helper node to not go over max payload for the response (max_response_size not used currently).

Agree.

> the syncing node should drop if missing values

Agree.

> if the syncing node asks for blocks `[n, N]` it should accept responses with `[n, K]` with `K<=N` as long as all `K-n+1` are present

Agree. However, we need to be cautious about how we track which heights were requested. If we receive a smaller response than expected, we also need to unmark from the in-flight requests the values we did not receive.

code/crates/test/proto/sync.proto
```diff
@@ -24,11 +24,28 @@ message ValueResponse {
     SyncedValue value = 2;
 }
 
+message BatchRequest {
+    uint64 from_height = 1;
+    uint64 to_height = 2;
+    uint64 max_response_size = 3;
```
hvanz: I see only one use of `max_response_size`, which is for node operators to limit the batch size when they don't have enough bandwidth. Then we would need to check that the response from the peer is within the requested bound. And if the response is bigger than `max_response_size` then we would need to penalize the peer for not following the request, either by reducing its score or discarding the peer completely. In any case, this field is unused right now.

bastienfaivre:
When calling `request_values`, we don't request value (or values) from a peer if there is a pending request to the `sync_height`. If there is no such request, we select the batch size to use.

> How is the batch size chosen? For sure, it should be 1 if we are only connected to v1 nodes. But if we have access to at least one v2 node, should the batch size be an arbitrary value or the minimum between this arbitrary value and the actual height's gap between the peers?

We then record the heights and request ID associated with the request.

When receiving a value or batch request, we retrieve these actual values.

> What should the response of the batch request look like if we are missing a few requested values?

When receiving a value or batch request, the consensus tries to apply the values. The sync clears the pending values

> WARNING! We should track the height associated with a given request. Now, it is based on the response, which can't be trusted theoretically.

hvanz:
Superseded by #1130

PR #1084, opened 2025-06-12
feat(code/sync): Re-request sync from another peer on failed responses

romac:
---

#### For all contributors


code/crates/sync/src/handle.rs
```diff
@@ -222,13 +222,20 @@ where
 
     metrics.decided_value_response_received(response.height.as_u64());
 
+    if response.value.is_none() {
```
nenadmilosevic95, line 249: What is the scenario that lead a node to enter this if instead of calling `on_empty_value_response`?

romac: This is the case where the node has replied with a well-formed response that does not contain a value, likely because the node does not actually have that value.

The other case is for when the node sends back a response that cannot be decoded.

Maybe we can clarify this with comments and better names.

nenadmilosevic95: I was guessing this is the case, in that case I think the on_empty_value_response should be renamed

PR #1090, opened 2025-06-17
feat(sync): Simplify proto messages for batching

hvanz:
Simplify proto messages by 
1. merging `BatchRequest` into `ValueRequest` and 
2. using an array of `SyncedValue` in `BatchResponse` instead of `BatchSyncValue`, and removing the `to_height` field.

This results in the following proto messages:
```
message ValueRequest {
    uint64 height = 1;
    optional uint64 end_height = 2;
}

message ValueResponse {
    uint64 height = 1;
    SyncedValue value = 2;
}

message BatchResponse {
    uint64 height = 1;
    repeated SyncedValue values = 2;
}
```

Using `SyncedValue` instead of `BatchSyncValue` results in replacing all `BTreeMap` with `Vec` for decided values in messaging and state retrieval

Also:
- Updated the data structure used for handling decided values across multiple components, including the application, engine, and sync modules. 
  - This change simplifies the data handling by using a Vec instead of a BTreeMap, improving performance and reducing complexity in the codebase. 
- Adjusted related functions and message formats accordingly.

### DIFF
--- a/code/crates/discovery/src/handlers/close.rs
+++ b/code/crates/discovery/src/handlers/close.rs
@@ -20,9 +20,7 @@ where
             || self
                 .active_connections
                 .get(&peer_id)
-                .map_or(true, |connection_ids| {
-                    !connection_ids.contains(&connection_id)
-                })
+                .is_none_or(|connection_ids| !connection_ids.contains(&connection_id))
     }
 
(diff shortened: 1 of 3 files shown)


bastienfaivre (commented):
Looks good for me!

PR #1091, opened 2025-06-17
feat(code/core-consensus): Buffer all inputs for the next `N` heights

romac:
Closes: #XXX

---

#### For all contributors


code/crates/core-consensus/src/state.rs
```diff
@@ -51,7 +51,7 @@ where
             ctx,
             driver,
             params,
-            input_queue: Default::default(),
+            input_queue: BoundedQueue::new(100), // FIXME
```
bastienfaivre: Do you plan to make it a config parameter?

bastienfaivre (approved):
LGTM, thanks!

code/crates/core-consensus/src/state.rs
romac: Ah right, forgot about that. Thanks!

romac: https://github.com/informalsystems/malachite/pull/1091/commits/fb0b9171f1ea1d8a23199e09a116748ece1064c8

code/crates/core-consensus/src/util/bounded_queue.rs
```diff
@@ -0,0 +1,392 @@
...
+        if let Some(values) = self.queue.get_mut(&index) {
+            values.push(value);
+            return true;
+        }
+
+        // If the index does not exist, check if we can add a new entry.
+        if self.queue.len() < self.capacity {
+            self.queue.insert(index, vec![value]);
+            return true;
+        }
```
ancazamfir, line 41: Not sure this is enough. We should handle the out of order case. Consensus messages at tip height may be added resulting in sync responses < the `sync_height` + `capacity` window to be dropped. No issue in main but with multiple syncs it will cause issues. For example this test should pass but it fails:
```
    #[test]
    fn push_out_of_order_to_full_queue_fails_for_new_index() {
        let mut queue = BoundedQueue::new(2);
        queue.push(10, "a");
        queue.push(30, "b");

        // Assert queue is full
        assert_eq!(queue.queue.len(), queue.capacity);

        // Try to push a new index to a full queue
        let result = queue.push(20, "c");

        assert!(result, "Push should not when capacity is reached");
        assert_eq!(queue.queue.len(), 2, "Queue length should not change");
        assert!(
            queue.queue.contains_key(&30),
            "The new element should be added"
        );
        assert!(
            !queue.queue.contains_key(&30),
            "The high value element should be remmoved"
        );
    }
```

ancazamfir:
`parse_default_config_file()` tests are failing.

code/crates/core-consensus/src/util/bounded_queue.rs
romac: Good catch! Fixed here: https://github.com/informalsystems/malachite/pull/1091/commits/b8be296563b8f30e3f1733f4f454949fb8b5be51

ancazamfir (approved):
👍

PR #1093, opened 2025-06-18
feat(code/sync): Re-request value when we get an invalid value

romac:
Closes: #XXX

---

#### For all contributors

### DIFF
--- a/code/crates/core-types/src/value.rs
+++ b/code/crates/core-types/src/value.rs
@@ -1,5 +1,7 @@
 use core::fmt::{Debug, Display};
 
+use malachitebft_peer::PeerId;
+
 /// Represents either `Nil` or a value of type `Value`.
 ///
 /// This type is isomorphic to `Option<Value>` but is more explicit about its intent.
@@ -130,8 +132,20 @@ impl ValuePayload {
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum ValueOrigin {
     /// Synchronization protocol
-    Sync,
+    Sync(PeerId),
 
     /// Consensus protocol
     Consensus,
+}
+
+impl ValueOrigin {
+    /// Value was received from the synchronization protocol.
+    pub fn is_sync(&self) -> bool {
+        matches!(self, Self::Sync(_))
+    }
+
+    /// Value was received from the consensus protocol.
+    pub fn is_consensus(&self) -> bool {
+        matches!(self, Self::Consensus)
+    }
 }
--- a/code/crates/engine/src/consensus.rs
+++ b/code/crates/engine/src/consensus.rs
@@ -18,7 +18,7 @@ use malachitebft_core_consensus::{
 };
 use malachitebft_core_types::{
     Context, Proposal, Round, SigningProvider, SigningProviderExt, Timeout, TimeoutKind,
-    ValidatorSet, ValueId, ValueOrigin, Vote,
+    ValidatorSet, Validity, ValueId, ValueOrigin, Vote,
 };
 use malachitebft_metrics::Metrics;
 use malachitebft_sync::{self as sync, ValueResponse};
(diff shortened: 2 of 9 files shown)


bastienfaivre (approved):
Good for me, thank you!

code/crates/core-types/src/value.rs
```diff
@@ -130,8 +132,20 @@ impl ValuePayload {
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum ValueOrigin {
     /// Synchronization protocol
-    Sync,
+    Sync(PeerId),
```
ancazamfir, line 135: We should maybe add at some point the peer to `height_per_request_id` and `pending_value_requests` in sync so we wouldn't need to keep it here.

code/crates/engine/src/sync.rs
```diff
@@ -97,6 +97,9 @@ pub enum Msg<Ctx: Context> {
 
     /// We received an invalid [`CommitCertificate`] from a peer
     InvalidCommitCertificate(PeerId, CommitCertificate<Ctx>, CertificateError<Ctx>),
+
+    /// We received an invalid value from a peer
+    InvalidValue(PeerId, ProposedValue<Ctx>),
```
ancazamfir: Could we merge `InvalidCommitCertificate` and `InvalidValue`? Sync doesn't care why it is invalid, and handlers are identical for these two cases. Only needs the `height` (not the full certificate or value) and `peer`.
Also have doubts about the `peer`, maybe sync should keep track of it. But we can discuss this later.

code/crates/engine/src/consensus.rs
```diff
@@ -611,6 +613,14 @@ where
...
                     .send(|| Event::ReceivedProposedValue(value.clone(), origin));
 
+                if let (Some(sync), ValueOrigin::Sync(from), Validity::Invalid) =
+                    (self.sync.as_ref(), origin, value.validity)
+                {
+                    if let Err(e) = sync.cast(SyncMsg::InvalidValue(from, value.clone())) {
+                        error!("Error when notifying sync of received proposed value: {e}");
+                    }
+                }
+
```
ancazamfir: Can we move this check in `NetworkEvent::SyncResponse()`?

code/crates/engine/src/consensus.rs
```diff
@@ -503,8 +503,8 @@ where
                                 reply_to,
                             },
                             &myself,
-                            |proposed| {
-                                Msg::<Ctx>::ReceivedProposedValue(proposed, ValueOrigin::Sync)
+                            move |proposed| {
```
ancazamfir, line 514: Could we maybe add here something like this? So we handle also the case the `value.id()` doesn't match the certificate:
```
     move |proposed| {
         if proposed.validity == Validity::Invalid || proposed.value.id() != value.certificate.value_id {
             if let Some(sync) = sync_ref.as_ref() {
                 if let Err(e) = sync.cast(SyncMsg::InvalidValue(peer, proposed.clone())) {
                        error!("Error when notifying sync of received proposed value: {e}");
                 }
              }
          }
          Msg::<Ctx>::ReceivedProposedValue(proposed, ValueOrigin::Sync(peer))
```

code/crates/core-types/src/value.rs
romac: Yes definitely a better of doing this

code/crates/engine/src/consensus.rs
romac: We do not know the validity at that point. The raw synced value first needs to be sent to the host for this.

This whole setup is quite cumbersome, we should eventually refactor this so that the sync actor can talk directly to the host instead of going through consensus.

code/crates/engine/src/sync.rs
romac: 2f2293089832ba6a32a9a436b2b8fcecb8038f30

code/crates/engine/src/consensus.rs
romac: Nevermind, I thought you meant in the sync that.

code/crates/engine/src/consensus.rs
romac: 681cd809239d4231e55be04751c68ba97688f9d6

code/crates/engine/src/consensus.rs
romac: 681cd809239d4231e55be04751c68ba97688f9d6

ancazamfir (approved):
🚀

code/crates/core-types/src/value.rs
romac: Will do this in a follow-up PR.

PR #1094, opened 2025-06-18
feat(sync/batching): simplify app interface

hvanz:
Main changes:
- app interface: in `HostMsg` remove `GetDecidedValue` and keep only `GetDecidedValues`
- proto definitions: remove `BatchResponse` message and consolidate all responses in `ValueResponse`

### DIFF
--- a/code/crates/sync/src/handle.rs
+++ b/code/crates/sync/src/handle.rs
@@ -568,10 +568,9 @@ where
     let mut end_height = height;
     let request_id = if batch_size > 1 {
         end_height = end_height.increment_by(batch_size);
-        let max_response_size = 10 * 1024 * 1024; // 10 MiB
         perform!(
             co,
-            Effect::SendBatchRequest(peer, BatchRequest::new(RangeInclusive::new(height, end_height), max_response_size), Default::default()),
+            Effect::SendBatchRequest(peer, BatchRequest::new(RangeInclusive::new(height, end_height)), Default::default()),
             Resume::ValueRequestId(id) => id,
         )
     } else {
--- a/code/crates/sync/src/handle.rs
+++ b/code/crates/sync/src/handle.rs
@@ -1,4 +1,3 @@
-use std::collections::BTreeMap;
 use std::ops::RangeInclusive;
 
 use derive_where::derive_where;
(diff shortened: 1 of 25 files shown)


code/crates/sync/src/state.rs
```diff
@@ -99,18 +99,29 @@ where
...
+        height: Ctx::Height,
+        except: Option<PeerId>,
+    ) -> Option<PeerId> {
         let peers_at_later_height = self
             .peers
             .iter()
             .filter(|(_, detail)| {
                 (detail.status.history_min_height..=detail.status.tip_height).contains(&height)
             })
+            .filter(|(&peer, _)| except.is_some_and(|x| x == peer))
```
ancazamfir, line 124:
```suggestion
            .filter(|(&peer, _)| except.is_none_or(|x| x != peer))
```

code/crates/sync/src/handle.rs
```diff
@@ -280,67 +236,43 @@ pub async fn on_value_response<Ctx>(
...
 {
-    debug!(%response.height, %request_id, %peer_id, "Received response");
+    let start = response.start_height;
+    debug!(start_height = %start, peer = %peer_id, "Received batch response with {} values", response.values.len());
 
-    state.remove_pending_value_request_by_height(&response.height);
+    // Remove pending requests for the values received
+    if !response.values.is_empty() {
+        let last_value_height = start.increment_by(response.values.len() as u64 - 1);
+        state.remove_pending_value_request_by_height_range(&(start..=last_value_height));
```
ancazamfir, line 245: We probably should not remove all range here. Instead remove heights when decided. The problem with this is that when we start a height, in `on_started_height` we will request again. I think you will see something like this:
- start height=1, request 1..100
- height=1 is decided, the range is removed
- start height=2:
  - consensus will have the 2..100 synced values buffered (after Romain's fix)
  - sync runs `on_started_height(2)` and will request 2..101
so we request again, consensus will end up with duplicate sync values

ancazamfir:
> - app interface: in HostMsg remove GetDecidedValue and keep only GetDecidedValues

Why don't we call `GetDecidedValue` multiple times instead? This way we don't change the app API. And in the sync we control how many values to actual get, in order, up to some approx size, etc.
We can optimize later.

hvanz:
Superseded by #1130

Issue #1092, opened 2025-06-18
feat/sync: Send requests in parallel


hvanz:
The current sync protocol implementation sends one request at a time.

With this optimization a node can send concurrently up to `N` requests for heights `h` to `h+N`, to the same or different peers. As soon as the request for the lowest height `h` is available, the consensus module can execute it, and the sync actor can send a request for `h+N+1`. The consensus module will need to buffer inputs (see #1091).

Requirements:
- Changes by this feature should not affect the proto sync messages or the application interface.

PR #1095, opened 2025-06-19
feat(code/sync): Parallel requests

bastienfaivre:
Closes: #1092 

---

#### For all contributors


bastienfaivre:
WARNING! This PR is based on #1091 and #1093.

code/crates/sync/src/handle.rs
```diff
@@ -158,35 +165,45 @@ pub async fn on_started_height<Ctx>(
 where
     Ctx: Context,
 {
-    let tip_height = height.decrement().unwrap_or(height);
+    if state.tip_height >= height && !restart {
```
bastienfaivre: @romac @ancazamfir

Upon reviewing the logs of several tests, I noticed that with certificate buffering, we always receive (when syncing) the `Decided` event before the corresponding `StartHeight` event for the same height. I tried moving the cast of `StartedHeight` (c.f. https://github.com/informalsystems/malachite/blob/main/code/crates/engine/src/consensus.rs#L393) before the process of starting the height replay queues messages, but I still get the behaviour sometimes. Moreover, since we cannot assume such an order theoretically, I needed to add this condition here, which only resets the tip and sync height when the height is restarted. Otherwise, we don't do anything. Without this condition, the sync would re-request the value of the corresponding height, as the pending validation is cleared by the `Decided` event that occurred before.

To optimize, the `on_decided` method now updates directly sync height and triggers potential requests. I know it is not optimal in the sense that we start syncing for a height that we have not even received confirmation from the consensus that it has started that height. But I assume that the consensus is running infinitely, as we can make progress that way without issues.

ancazamfir: > I tried moving the cast of StartedHeight (c.f. https://github.com/informalsystems/malachite/blob/main/code/crates/engine/src/consensus.rs#L393) before the process of starting the height replay queues messages,

We should do this change

> but I still get the behaviour sometimes.

do you see this in a test? which one?

> since we cannot assume such an order theoretically..

You mean `StartedHeight(h), Decided(h)` order? This should be guaranteed if consensus sends `StartedHeight(h)` before it processes the buffered messages (which would result in sending `Decided(h)` in case of parallel sync). They are both sent from consensus to sync.

imo, we should not remove the request when we get decided for a height `h`, just mark it decided and remove it when we get `startedHeight` for `h+1`.

Also, we should do a proper state machine for the requests. We can combine `pending_requests` with `pending_validation` in one and keep track of state.

ancazamfir: I made the changes suggested above in `anca/try_request_fsm` branch, off your PR's branch. Let me know the test where you saw issues with the change where consensus buffer replay is done after the StrtedHeight is sent to sync.

romac: > imo, we should not remove the request when we get decided for a height h, just mark it decided and remove it when we get startedHeight for h+1.

curious what difference that would make?

romac: > Also, we should do a proper state machine for the requests. We can combine pending_requests with pending_validation in one and keep track of state.

agreed

bastienfaivre: I added a comment [here](https://github.com/informalsystems/malachite/pull/1104#issuecomment-2996027524).

ancazamfir: > > imo, we should not remove the request when we get decided for a height h, just mark it decided and remove it when we get startedHeight for h+1.
> 
> curious what difference that would make?

I was thinking of the window between `Decided(h)` and `StartedHeight(h+1)` where we could do a new request for `h` if we remove the old one (that's just been decided). 

Also doesn't feel right that we would update the sync height in `on_decided()` and then in `on_started_height()`. Same for tip height.

romac (commented):
Let's add an entry in `RELEASE_NOTES` and then we can merge!

RELEASE_NOTES.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-> Nothing yet
+- Add parallel requests for the sync module ([#1092](https://github.com/informalsystems/malachite/issues/1092))
```
hvanz, line 6: add changes to configuration?

romac: Ah good catch! Let's add those here and in the BREAKING_CHANGES doc. Let's not mention the TOML config as that is app-specific but rather the changes in the `malachitebft-config` and `malachitebft-sync` structs.

romac:
I will update the crates version on main to 0.4.0-pre so silence the semver warning.

romac (approved):

PR #1097, opened 2025-06-20
chore(code/sync): Collapse log when we have multiple pending requests

romac:


code/crates/sync/src/handle.rs
```diff
@@ -412,23 +412,31 @@ where
...
-                break;
-            };
-
-            request_value_from_peer(&co, state, metrics, height, peer).await?;
-        } else {
-            debug!(height.sync = %height, "Already have a pending request or validation for this height");
+    // Find out the first height for which we do not have a pending request or validation.
+    while state.has_pending_value_request(&height) || state.has_pending_value_validation(&height) {
+        height = height.increment();
+        if height >= limit {
```
bastienfaivre, line 418: If we reach this case (i.e., all heights in the current range have a pending request or validation), the `height` variable is equal to the `limit`, which is not in the range. But the following code will request a value for this height, as the limit is checked at the end of the `loop`.

code/crates/sync/src/handle.rs
```diff
@@ -412,23 +412,31 @@ where
...
+        let Some(peer) = state.random_peer_with_tip_at_or_above(height) else {
+            debug!(height.sync = %height, "No peer to request sync from");
+            // No peer reached this height yet, we can stop here.
+            break;
+        };
+
+        request_value_from_peer(&co, state, metrics, height, peer).await?;
+
+        height = height.increment();
         if height >= limit {
```
bastienfaivre: Move this condition to the beginning of the `loop` to avoid the issue described above.

PR #1104, opened 2025-06-23
feat(code/sync): Requests FSM

bastienfaivre:
Closes: #XXX

---

#### For all contributors

### DIFF
--- a/code/crates/sync/src/handle.rs
+++ b/code/crates/sync/src/handle.rs
@@ -248,15 +248,23 @@ where
     // We do not update the peer score if we do not know the response time.
     // This should never happen, but we need to handle it gracefully just in case.
 
+    if let Some(height) = state.get_height_for_request_id(&request_id) {
+        if height != response.height {
+            warn!(%response.height, %request_id, "Received response for wrong height");
+
+            // TODO: penalize peer?
+        }
+    }
+
     if response.value.is_none() {
         warn!(%response.height, %request_id, "Received invalid value response");
 
-        if let Some(height) = state.remove_pending_value_request_by_id(&request_id) {
-            // It is possible that this height has been already validated via consensus messages.
-            // Therefore, we ignore the response status.
-            if !state.is_pending_value_request_validated(&height) {
-                request_value_from_peer_except(co, state, metrics, height, peer_id).await?;
-            }
+        // It is possible that this height has been already validated via consensus messages.
+        // Therefore, we ignore the response status.
+        if !state.is_pending_value_request_validated_by_height(&response.height) {
+            state.remove_pending_value_request_by_id(&request_id);
+
+            request_value_from_peer_except(co, state, metrics, response.height, peer_id).await?;
         }
     } else {
         state.response_received(request_id, response.height);
@@ -277,10 +285,10 @@ where
 {
     debug!(%request_id, %peer, "Received invalid response");
 
-    if let Some(height) = state.remove_pending_value_request_by_id(&request_id) {
-        // It is possible that this height has been already validated via consensus messages.
-        // Therefore, we ignore the response status.
-        if !state.is_pending_value_request_validated(&height) {
+    // It is possible that this height has been already validated via consensus messages.
+    // Therefore, we ignore the response status.
+    if !state.is_pending_value_request_validated_by_id(&request_id) {
+        if let Some(height) = state.remove_pending_value_request_by_id(&request_id) {
             request_value_from_peer_except(co, state, metrics, height, peer).await?;
         }
     }
--- a/code/crates/sync/src/handle.rs
+++ b/code/crates/sync/src/handle.rs
@@ -191,7 +191,7 @@ where
 {
     debug!(height.tip = %height, "Updating request state");
 
-    state.decided_received(height);
+    state.validate_response(height);
 
     Ok(())
 }
--- a/code/crates/sync/src/handle.rs
+++ b/code/crates/sync/src/handle.rs
@@ -251,11 +251,12 @@ where
     if response.value.is_none() {
         warn!(%response.height, %request_id, "Received invalid value response");
 
-        // If we received an empty response, we will try to request the value from another peer.
-        request_value_from_peer_except(co, state, metrics, response.height, peer_id).await?;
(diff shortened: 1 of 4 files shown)


bastienfaivre:
@ancazamfir I added a test that, despite running successfully, showcases the issue I mentioned [here](https://github.com/informalsystems/malachite/pull/1095#discussion_r2159517726). Indeed, we observe in the logs that we receive the `Decided` event before the `StartedHeight` one:
```
2025-06-23T10:29:05.954929Z DEBUG test{id=1}:node{id=3}:node{moniker=node-3}:sync{height.tip=2 height.sync=3}: Updating request state height.tip=4
2025-06-23T10:29:05.954949Z DEBUG test{id=1}:node{id=3}:node{moniker=node-3}:sync{height.tip=2 height.sync=3}: Starting new height height.tip=3 height.sync=4 restart=false
```
And so, this incorrect order leads to an incorrect tip and sync height, that are one height late.

Moreover, note that the queued messages are replayed [here](https://github.com/informalsystems/malachite/blob/72c4fcc8b4a7df04492ae37f918e9d41f88f89b3/code/crates/engine/src/consensus.rs#L371) (to be more precise, [here](https://github.com/informalsystems/malachite/blob/72c4fcc8b4a7df04492ae37f918e9d41f88f89b3/code/crates/core-consensus/src/handle/start_height.rs#L67)), and not with `self.process_buffered_msgs(&myself, state).await;`, which is different:
```rust
    /// A buffer of messages that were received while
    /// consensus was `Unstarted` or in the `Recovering` phase
    msg_buffer: MessageBuffer<Ctx>,
```
I did the correct change in [ad94f06](https://github.com/informalsystems/malachite/pull/1104/commits/ad94f06214cadbea0d3ec5f6aee2c18a77a215bf), and now the events are correctly ordered.

But, will we always get the events in the correct order?

hvanz:
> But, will we always get the events in the correct order?

It feels like the sync protocol is getting so complex that it deserves a formal spec and MBT.

ancazamfir:
> Moreover, note that the queued messages are replayed [here](https://github.com/informalsystems/malachite/blob/72c4fcc8b4a7df04492ae37f918e9d41f88f89b3/code/crates/engine/src/consensus.rs#L371) (to be more precise, [here](https://github.com/informalsystems/malachite/blob/72c4fcc8b4a7df04492ae37f918e9d41f88f89b3/code/crates/core-consensus/src/handle/start_height.rs#L67)), and not with `self.process_buffered_msgs(&myself, state).await;`, which is different:

ah right, you are correct, my bad! too many queues and buffers :) Thanks for the test and the fix.

ancazamfir:
> But, will we always get the events in the correct order?

If we look at messages from consensus to sync actor, the order should be preserved. 
The one that is not guaranteed is the (request_response, decided) as the first comes from the network and the second from consensus. So sync should be able to handle the case where it sends the request then sees decided from consensus and then response.

code/crates/sync/src/handle.rs
```diff
@@ -243,8 +232,6 @@ where
 {
     debug!(%response.height, %request_id, %peer_id, "Received response");
 
-    state.remove_pending_value_request_by_height(&response.height);
-
     let response_time = metrics.value_response_received(response.height.as_u64());
```
hvanz, line 238: >  So sync should be able to handle the case where it sends the request then sees decided from consensus and then response.

Then shouldn't we check here that `pending_value_requests` is different than `RequestState::Validated`? If consensus already decided on a value and changed the status in `on_decided`, then `on_value_response` should not call `request_value_from_peer_except`.

bastienfaivre: Handled in [d1d5853](https://github.com/informalsystems/malachite/pull/1104/commits/d1d5853df4686a7530be1b73561e064422d8f664).

bastienfaivre:
The sync actor now handles the event order problem and the wrong height response.

code/crates/discovery/src/handlers/dial.rs
```diff
@@ -116,8 +120,24 @@ where
             .dial_add_peer_id_to_dial_data(connection_id, peer_id);
     }
 
-    pub fn handle_failed_connection(&mut self, swarm: &mut Swarm<C>, connection_id: ConnectionId) {
+    pub fn handle_failed_connection(
```
romac: Let's merge `main` in [bastien/parallel-sync](https://github.com/informalsystems/malachite/tree/bastien/parallel-sync) to clean up the diff

bastienfaivre: Done.

code/crates/sync/src/handle.rs
```diff
@@ -178,31 +173,25 @@ where
...
 
-    state.sync_height = height.increment();
-    state.tip_height = height;
-
-    state.remove_pending_value_validation(&height);
-    state.remove_pending_value_request_by_height(&height);
-
-    // Trigger potential requests if possible.
-    request_values(co, state, metrics).await?;
+    state.decided_received(height);
```
romac, line 194:
```suggestion
    state.decision_made(height);
```

code/crates/sync/src/state.rs
```diff
@@ -121,17 +133,42 @@ where
...
+                *state = RequestState::WaitingValidation;
+            }
+        }
+    }
+
+    /// Mark that a decided value has been validated for a height.
+    ///
+    /// State transition: WaitingValidation -> Validated
+    /// It is also possible to have the following transition: WaitingResponse -> Validated.
+    pub fn decided_received(&mut self, height: Ctx::Height) {
```
romac:
```suggestion
    pub fn validate_response(&mut self, height: Ctx::Height) {
```

code/crates/test/tests/it/value_sync.rs
```diff
@@ -199,6 +199,42 @@ pub async fn start_late() {
         .await
 }
 
+#[tokio::test]
+pub async fn start_late_parallel_requests() {
+    const HEIGHT: u64 = 5;
```
romac:
```suggestion
    const HEIGHT: u64 = 10;
```

PR #1109, opened 2025-06-26
feat(code)!: Decouple `Host` messages from `Consensus` actor

romac:
Closes: #XXX

---

#### For all contributors

### DIFF
--- a/code/crates/engine/src/consensus.rs
+++ b/code/crates/engine/src/consensus.rs
@@ -1145,7 +1145,7 @@ where
                 Ok(r.resume_with(()))
             }
 
-            Effect::RebroadcastRoundCertificate(certificate, r) => {
+            Effect::RepublishRoundCertificate(certificate, r) => {
                 // Notify any subscribers that we are about to rebroadcast a round certificate
                 self.tx_event
                     .send(|| Event::RebroadcastRoundCertificate(certificate.clone()));
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -18,7 +18,7 @@
 
 ### `malachitebft-core-consensus`
 
-- Renamed `Effect::RebroadcastVote` to `Effect::RepublishVote` ([#1011](https://github.com/informalsystems/malachite/issues/1011))
+- Rename `Effect::RebroadcastVote` to `Effect::RepublishVote` and `Effect::RebroadcastRoundCertificate` to `Effect::RepublishRoundCertificate` ([#1011](https://github.com/informalsystems/malachite/issues/1011))
 
 ## 0.4.0
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Update libp2p to v0.56.x ([#1124](https://github.com/informalsystems/malachite/pull/1124))
-- Rename `Effect::RebroadcastVote` to `Effect::RepublishVote` ([#1011](https://github.com/informalsystems/malachite/issues/1011))
+- Rename `Effect::RebroadcastVote` to `Effect::RepublishVote` and `Effect::RebroadcastRoundCertificate` to `Effect::RepublishRoundCertificate` ([#1011](https://github.com/informalsystems/malachite/issues/1011))
 - Decouple `Host` messages from the `Consensus` actor ([#1109](https://github.com/informalsystems/malachite/pull/1109))
 
 ## 0.4.0
--- a/code/crates/core-consensus/src/effect.rs
+++ b/code/crates/core-consensus/src/effect.rs
@@ -89,15 +89,15 @@ where
     /// Resume with: [`resume::Continue`]
     PublishLivenessMsg(LivenessMsg<Ctx>, resume::Continue),
 
-    /// Rebroadcast a vote to peers
+    /// Re-publish a vote to peers
     ///
     /// Resume with: [`resume::Continue`]
     RepublishVote(SignedVote<Ctx>, resume::Continue),
 
-    /// Rebroadcast a round certificate to peers
+    /// Re-publish a round certificate to peers
     ///
     /// Resume with: [`resume::Continue`]
-    RebroadcastRoundCertificate(RoundCertificate<Ctx>, resume::Continue),
+    RepublishRoundCertificate(RoundCertificate<Ctx>, resume::Continue),
 
     /// Requests the application to build a value for consensus to run on.
     ///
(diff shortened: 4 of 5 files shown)


code/crates/app-channel/src/connector.rs
```diff
@@ -94,28 +84,11 @@ where
...
-                                    ValueOrigin::Consensus,
-                                );
-                                if let Err(e) = consensus.cast(msg) {
-                                    error!("Failed to send back undecided value to consensus: {e}");
-                                }
-                            }
-                        }
-                    }
-                });
+                let values = rx_values.await?;
```
sergio-mena: Don't fully understand this part of the code, but, comparing left an right, I see we are now blocking (on the `await` whereas before we where explicitly avoiding it.

code/crates/engine/src/consensus.rs
```diff
@@ -963,12 +971,18 @@ where
...
+                    ractor::call!(self.host, |reply_to| HostMsg::StartedRound {
+                        height,
+                        round,
+                        proposer: proposer.clone(),
+                        role,
+                        reply_to,
+                    })?;
+
+                for value in undecided_values {
+                    let _ = myself.cast(Msg::ReceivedProposedValue(value, ValueOrigin::Consensus));
```
sergio-mena, line 984: Just for my own instruction: this is the message you said is not emitted yet in `proposalOnly` mode, isn't it?

code/crates/engine/src/host.rs
```diff
@@ -8,33 +8,64 @@ use malachitebft_core_consensus::{Role, VoteExtensionError};
...
 use crate::util::streaming::StreamMessage;
 
 pub use malachitebft_core_consensus::{LocallyProposedValue, ProposedValue};
 
 /// A reference to the host actor.
 pub type HostRef<Ctx> = ActorRef<HostMsg<Ctx>>;
 
+/// What to do next after a decision.
+#[derive_where(Debug)]
+pub enum Next<Ctx: Context> {
```
sergio-mena, line 20: Maybe `NextHeight` is clearer?

code/crates/engine/src/host.rs
```diff
@@ -8,33 +8,64 @@ use malachitebft_core_consensus::{Role, VoteExtensionError};
...
 pub enum HostMsg<Ctx: Context> {
-    /// Consensus is ready
-    ConsensusReady(ConsensusRef<Ctx>),
+    /// Notifies the application that consensus is ready.
+    ///
+    /// The application MUST reply with a message to instruct
+    /// consensus to start at a given height.
+    ConsensusReady {
+        /// Use this reply port to instruct consensus to start the first height.
+        reply_to: RpcReplyPort<(Ctx::Height, Ctx::ValidatorSet)>,
```
sergio-mena, line 37: Would the interface be more homogeneous if we used here `Next::Start`?
Although I think we can't do that without opening the door to "accepting" `Next::Restart` at build time, even if we add a sanity check at runtime.
Maybe something to consider later on if we manage to remove the `Restart` variant in the future.

code/crates/starknet/host/src/actor.rs
```diff
@@ -12,9 +12,8 @@ use tokio::time::Instant;
 use tracing::{debug, error, info, trace, warn};
 
 use malachitebft_core_consensus::{PeerId, Role, VoteExtensionError};
-use malachitebft_core_types::{CommitCertificate, Round, Validity, ValueId, ValueOrigin};
-use malachitebft_engine::consensus::{ConsensusMsg, ConsensusRef};
```
sergio-mena, line 16: So cool to get rid of `ConsensusMsg` here! 🚀

code/crates/starknet/host/src/actor.rs
```diff
@@ -204,53 +204,24 @@ impl Host {
...
-        .await?;
-
-    let consensus = state.consensus.as_ref().unwrap();
-
-    for value in undecided_values {
-        info!(%height, %round, hash = %value.value, "Replaying already known proposed value");
-
-        consensus.cast(ConsensusMsg::ReceivedProposedValue(
-            value,
-            ValueOrigin::Consensus,
```
sergio-mena, line 251: Are we still passing this info?
Although it seems it's always `ValueOrigin::Consensus`

sergio-mena (approved):
I like these changes. They will simplify the readability of `TestConsensus` in Muppet 🚀

code/crates/engine/src/consensus.rs
romac: No, it's the one in this PR: https://github.com/informalsystems/malachite/pull/1110

code/crates/engine/src/host.rs
romac: Well it's not really the next height, but what to do next after a decision, it could very well be a restart of the same height if that one failed to commit in the app for some reason. Not sure what's the best name for it, agreed that `Next` is perhaps too barebones.

code/crates/engine/src/host.rs
romac: I'd rather not enforce this via runtime checks if it can be done statically.

> Maybe something to consider later on if we manage to remove the Restart variant in the future.

Agreed!

code/crates/starknet/host/src/actor.rs
romac: This is the missing piece in this PR and the reason why it's still draft. Need to find a nice way of doing this. Happy to hear suggestions if you have any :)

code/crates/engine/src/host.rs
sergio-mena: OK, it was just a suggestion. Up to you folks

code/crates/app-channel/src/connector.rs
romac: Fixed in f7c4428df3a2dd368d57835dfb942a0e3f4f721

code/crates/starknet/host/src/actor.rs
romac: The origin is now supplied here when processing the reply sent via the `RpcReplyPort` in `on_started_round`:

https://github.com/informalsystems/malachite/blob/4ee029e93a890e2c579296fbd2b4a5a96a1047eb/code/crates/engine/src/consensus.rs#L984

Those values are always considered to have been originated via consensus, hence we can hardcode it in the Consensus actor instead of leaving it up to the app.

romac:
@ancazamfir @hvanz Any concerns with merging this?

ancazamfir:
> @ancazamfir @hvanz Any concerns with merging this?

Looks good!

ancazamfir (approved):

PR #1130, opened 2025-07-11
feat(code/sync): Batch values in sync messages

hvanz:
Closes: #1070

### DIFF
--- a/code/crates/app-channel/src/connector.rs
+++ b/code/crates/app-channel/src/connector.rs
@@ -252,11 +252,11 @@ where
                 consensus.cast(rx.await?.into())?;
             }
 
-            HostMsg::GetDecidedValue { height, reply_to } => {
+            HostMsg::GetDecidedValues { range, reply_to } => {
                 let (reply, rx) = oneshot::channel();
 
                 self.sender
-                    .send(AppMsg::GetDecidedValue { height, reply })
+                    .send(AppMsg::GetDecidedValues { range, reply })
                     .await?;
 
                 reply_to.send(rx.await?)?;
--- a/code/crates/config/src/lib.rs
+++ b/code/crates/config/src/lib.rs
@@ -445,6 +445,9 @@ pub struct ValueSyncConfig {
     /// Threshold for considering a peer inactive
     #[serde(with = "humantime_serde")]
     pub inactive_threshold: Duration,
+
+    /// Maximum number of decided values to request in a single batch
+    pub batch_size: usize,
 }
 
 impl Default for ValueSyncConfig {
@@ -458,6 +461,7 @@ impl Default for ValueSyncConfig {
             parallel_requests: 5,
             scoring_strategy: ScoringStrategy::default(),
             inactive_threshold: Duration::from_secs(60),
+            batch_size: 1000,
         }
     }
 }
--- a/code/crates/engine/src/network.rs
+++ b/code/crates/engine/src/network.rs
@@ -1,11 +1,11 @@
-use std::collections::{BTreeSet, HashMap};
+use std::collections::{BTreeMap, HashMap};
 use std::marker::PhantomData;
 
 use async_trait::async_trait;
 use derive_where::derive_where;
 use eyre::eyre;
 use libp2p::identity::Keypair;
-use libp2p::request_response;
+use libp2p::{request_response, StreamProtocol};
 use ractor::{Actor, ActorProcessingErr, ActorRef, RpcReplyPort};
 use tokio::task::JoinHandle;
 use tracing::{error, trace};
@@ -100,7 +100,7 @@ pub struct Args {
 pub enum NetworkEvent<Ctx: Context> {
     Listening(Multiaddr),
 
-    PeerConnected(PeerId),
+    PeerConnected(PeerId, Vec<StreamProtocol>),
     PeerDisconnected(PeerId),
 
     Vote(PeerId, SignedVote<Ctx>),
@@ -122,7 +122,7 @@ pub enum State<Ctx: Context> {
     Stopped,
     Running {
         listen_addrs: Vec<Multiaddr>,
-        peers: BTreeSet<PeerId>,
+        peers: BTreeMap<PeerId, Vec<StreamProtocol>>,
         output_port: OutputPort<NetworkEvent<Ctx>>,
         ctrl_handle: Box<CtrlHandle>,
         recv_task: JoinHandle<()>,
@@ -212,7 +212,7 @@ where
 
         Ok(State::Running {
             listen_addrs: Vec::new(),
-            peers: BTreeSet::new(),
+            peers: BTreeMap::new(),
             output_port: OutputPort::with_capacity(128),
             ctrl_handle: Box::new(ctrl_handle),
             recv_task,
@@ -253,8 +253,8 @@ where
                     subscriber.send(NetworkEvent::Listening(addr.clone()));
                 }
 
-                for peer in peers.iter() {
-                    subscriber.send(NetworkEvent::PeerConnected(*peer));
+                for (peer_id, protocols) in peers.iter() {
+                    subscriber.send(NetworkEvent::PeerConnected(*peer_id, protocols.clone()));
                 }
 
                 subscriber.subscribe_to_port(output_port);
@@ -333,9 +333,9 @@ where
                 output_port.send(NetworkEvent::Listening(addr));
             }
 
-            Msg::NewEvent(Event::PeerConnected(peer_id)) => {
-                peers.insert(peer_id);
-                output_port.send(NetworkEvent::PeerConnected(peer_id));
+            Msg::NewEvent(Event::PeerConnected(peer_id, protocols)) => {
+                peers.insert(peer_id, protocols.clone());
+                output_port.send(NetworkEvent::PeerConnected(peer_id, protocols));
             }
 
             Msg::NewEvent(Event::PeerDisconnected(peer_id)) => {
--- a/code/crates/engine/src/sync.rs
+++ b/code/crates/engine/src/sync.rs
@@ -424,6 +424,7 @@ where
                                 &myself,
                                 state,
                                 sync::Input::SyncRequestTimedOut(
+                                    request_id,
                                     inflight.peer_id,
                                     inflight.request,
                                 ),
--- a/code/crates/engine/src/sync.rs
+++ b/code/crates/engine/src/sync.rs
@@ -303,6 +303,12 @@ where
                     .await?;
             }
 
+            Msg::NetworkEvent(NetworkEvent::PeerConnected(peer_id, protocols)) => {
+                info!(%peer_id, "Connected to peer");
+
+                state.sync.add_peer(peer_id, protocols);
+            }
+
             Msg::NetworkEvent(NetworkEvent::PeerDisconnected(peer_id)) => {
                 info!(%peer_id, "Disconnected from peer");
 
--- a/code/crates/engine/src/sync.rs
+++ b/code/crates/engine/src/sync.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::ops::RangeInclusive;
 use std::time::Duration;
 
 use async_trait::async_trait;
@@ -89,7 +90,11 @@ pub enum Msg<Ctx: Context> {
     StartedHeight(Ctx::Height, bool),
 
     /// Host has a response for the blocks request
-    GotDecidedValue(InboundRequestId, Ctx::Height, Option<RawDecidedValue<Ctx>>),
+    GotDecidedValues(
+        InboundRequestId,
+        RangeInclusive<Ctx::Height>,
+        Vec<RawDecidedValue<Ctx>>,
+    ),
 
     /// A timeout has elapsed
     TimeoutElapsed(TimeoutElapsed<Timeout>),
@@ -276,13 +281,12 @@ where
                 Ok(r.resume_with(()))
             }
 
-            Effect::GetDecidedValue(request_id, height, r) => {
+            Effect::GetDecidedValues(request_id, range, r) => {
+                let range_cloned = range.clone();
                 self.host.call_and_forward(
-                    |reply_to| HostMsg::GetDecidedValue { height, reply_to },
+                    |reply_to| HostMsg::GetDecidedValues { range, reply_to },
                     myself,
-                    move |synced_value| {
-                        Msg::<Ctx>::GotDecidedValue(request_id, height, synced_value)
-                    },
+                    move |values| Msg::<Ctx>::GotDecidedValues(request_id, range_cloned, values),
                     None,
                 )?;
 
@@ -382,11 +386,11 @@ where
                     .await?;
             }
 
-            Msg::GotDecidedValue(request_id, height, block) => {
+            Msg::GotDecidedValues(request_id, range, block) => {
                 self.process_input(
                     &myself,
                     state,
-                    sync::Input::GotDecidedValue(request_id, height, block),
+                    sync::Input::GotDecidedValues(request_id, range, block),
                 )
                 .await?;
             }
--- a/code/crates/sync/src/behaviour.rs
+++ b/code/crates/sync/src/behaviour.rs
@@ -20,16 +20,26 @@ pub struct Behaviour {
 pub type Event = rpc::Event<RawRequest, RawResponse>;
 
 impl Behaviour {
-    pub const PROTOCOL: [(StreamProtocol, ProtocolSupport); 1] = [(
+    pub const SYNC_V2_PROTOCOL: (StreamProtocol, ProtocolSupport) = (
+        StreamProtocol::new("/malachitebft-sync/v2beta1"),
+        ProtocolSupport::Full,
+    );
+
+    pub const SYNC_V1_PROTOCOL: (StreamProtocol, ProtocolSupport) = (
         StreamProtocol::new("/malachitebft-sync/v1beta1"),
         ProtocolSupport::Full,
-    )];
+    );
+
+    pub const PROTOCOLS: [(StreamProtocol, ProtocolSupport); 2] = [
+        Self::SYNC_V2_PROTOCOL, // Sync v2 is preferred
+        Self::SYNC_V1_PROTOCOL, // Sync v1 is for backward compatibility
+    ];
 
     pub fn new(config: Config) -> Self {
         let rpc_config = rpc::Config::default().with_request_timeout(config.request_timeout);
 
         Self {
-            rpc: rpc::Behaviour::with_codec(Codec::new(config), Self::PROTOCOL, rpc_config),
+            rpc: rpc::Behaviour::with_codec(Codec::new(config), Self::PROTOCOLS, rpc_config),
             // metrics: None,
         }
     }
@@ -38,7 +48,7 @@ impl Behaviour {
         let rpc_config = rpc::Config::default().with_request_timeout(config.request_timeout);
 
         Self {
-            rpc: rpc::Behaviour::with_codec(Codec::new(config), Self::PROTOCOL, rpc_config),
+            rpc: rpc::Behaviour::with_codec(Codec::new(config), Self::PROTOCOLS, rpc_config),
             // metrics: Some(Metrics::new(registry)),
         }
     }
--- a/code/crates/sync/src/config.rs
+++ b/code/crates/sync/src/config.rs
@@ -3,6 +3,7 @@ use std::time::Duration;
 use crate::scoring::Strategy;
 
 const DEFAULT_PARALLEL_REQUESTS: u64 = 5;
+const DEFAULT_BATCH_SIZE: usize = 1000;
 
 #[derive(Copy, Clone, Debug)]
 pub struct Config {
@@ -13,6 +14,7 @@ pub struct Config {
     pub parallel_requests: u64,
     pub scoring_strategy: Strategy,
     pub inactive_threshold: Option<Duration>,
+    pub batch_size: usize,
 }
 
 impl Config {
@@ -52,6 +54,11 @@ impl Config {
         self.inactive_threshold = inactive_threshold;
         self
     }
+
+    pub fn with_batch_size(mut self, batch_size: usize) -> Self {
+        self.batch_size = batch_size;
+        self
+    }
 }
 
 impl Default for Config {
@@ -64,6 +71,7 @@ impl Default for Config {
             parallel_requests: DEFAULT_PARALLEL_REQUESTS,
             scoring_strategy: Strategy::default(),
             inactive_threshold: None,
+            batch_size: DEFAULT_BATCH_SIZE,
         }
     }
 }
--- a/code/crates/sync/src/handle.rs
+++ b/code/crates/sync/src/handle.rs
@@ -1,3 +1,4 @@
+use std::cmp::max;
 use std::ops::RangeInclusive;
 
 use derive_where::derive_where;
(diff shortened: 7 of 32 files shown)


code/crates/test/proto/sync.proto
```diff
@@ -17,11 +17,12 @@ message Status {
 
 message ValueRequest {
     uint64 height = 1;
+    optional uint64 end_height = 2;
 }
 
 message ValueResponse {
-    uint64 height = 1;
-    SyncedValue value = 2;
+    uint64 start_height = 1;
```
hvanz, line 24: We can safely rename the fields because the binary format is compatible as long as we keep the same tag number.

code/crates/test/proto/sync.proto
```diff
@@ -17,11 +17,12 @@ message Status {
...
+    repeated SyncedValue values = 2;
```
hvanz, line 25: We can safely add `repeated` and backward compatibility because an old node expecting a single value will parse the last occurrence of the field and ignore the previous ones.

code/crates/app-channel/src/connector.rs
```diff
@@ -252,11 +252,11 @@ where
                 consensus.cast(rx.await?.into())?;
             }
 
-            HostMsg::GetDecidedValue { height, reply_to } => {
+            HostMsg::GetDecidedValues { range, reply_to } => {
```
ancazamfir: IMO, it would be simpler in a first phase for consensus to ask for individual values from the app such that we don't break the APIs. This optimization could be added later.

code/crates/sync/src/behaviour.rs
```diff
@@ -20,16 +20,26 @@ pub struct Behaviour {
 pub type Event = rpc::Event<RawRequest, RawResponse>;
 
 impl Behaviour {
-    pub const PROTOCOL: [(StreamProtocol, ProtocolSupport); 1] = [(
+    pub const SYNC_V2_PROTOCOL: (StreamProtocol, ProtocolSupport) = (
```
ancazamfir: Do we really need new sync protocol versions for this PR if V2 is backwards compatible with V1? I guess the code would need to change such that if a V2 node asks for `h,..,H` range and it receives only response for `h` (from a V1 node) it will accept it and ask for `h+1,..,H`.

code/crates/test/tests/it/value_sync.rs
```diff
@@ -235,6 +235,43 @@ pub async fn start_late_parallel_requests() {
...
+
+    test.add_node()
+        .with_voting_power(5)
+        .start_after(1, Duration::from_secs(10))
+        .wait_until(HEIGHT)
+        .success();
+
+    test.build()
+        .run_with_params(
+            Duration::from_secs(30),
```
ancazamfir, line 268: probably needs a higher duration.

code/crates/test/tests/unit/sync.rs
```diff
@@ -0,0 +1,117 @@
...
+                peer_id,
+                tip_height: Height::new(max),
+                history_min_height: Height::new(min),
+            },
+        },
+    );
+}
+
+#[test]
+fn filtering_peers() {
```
ancazamfir: The tests here are a bit hard to parse and understand what is covered. Ideally there should be a list of test entries, with test name (short description), input, expected output after running `filter_peers_by_range`, etc.
I can make the change if you want.

code/crates/sync/src/handle.rs
```diff
@@ -294,71 +303,79 @@ pub async fn on_invalid_value_response<Ctx>(
...
+        height = height.increment();
+    }
 
+    debug!(%request_id, range = %DisplayRange::<Ctx>(&range), "Sending response to peer");
     perform!(
         co,
         Effect::SendValueResponse(
             request_id,
-            ValueResponse::new(height, response),
+            ValueResponse::new(*start, values),
```
ancazamfir, line 386: Ideally we should trim to be under rpc response max size. Also to prevent request for a very high range.

code/crates/test/tests/it/value_sync.rs
```diff
@@ -235,6 +235,43 @@ pub async fn start_late_parallel_requests() {
...
+        .success();
+
+    test.add_node()
+        .with_voting_power(10)
+        .start()
+        .wait_until(HEIGHT * 2)
+        .success();
+
+    test.add_node()
+        .with_voting_power(5)
```
ancazamfir: setting power to 0 will make consensus progress faster.

code/crates/sync/src/behaviour.rs
hvanz: Good point! I removed the v2 version; it's not needed. This code comes from the first implementation of batching. I manually tested syncing again in mixed networks and it works as expected.

code/crates/test/tests/it/value_sync.rs
hvanz: Yep, spot on. I find the problem with the test the hard way, before seeing your comment.

code/crates/test/tests/unit/sync.rs
hvanz: Agree. I did split the tests. Also now the function under test is simpler, which reduces the test scenarios.

code/crates/test/tests/it/value_sync.rs
hvanz: Right. With voting power 0, on this test I can reduce the timeout from 60 to 30 seconds and it still passes. That doesn't work in the Starknet test, even when I decrease `block_size` from 1Mib to 1kib (it takes 30.7 seconds in my laptop).

code/crates/sync/src/handle.rs
hvanz: True. Also when values are big, responses could become huge and the node expecting them will time out. 

The problem with capping the response is that, in the current design, a response is valid if it contains all requested heights or a single one (for nodes that don't support batching). We'll have to change how to handle partial responses.

code/crates/app-channel/src/connector.rs
hvanz: In principle I agree but I'm not sure what's the best way to collect all values returned by `HostMsg::GetDecidedValue`, so then we can build the sync response with those values in the correct order.  The `GotDecidedValues` messages may arrive out of order and we wouldn't know which one is the last one.

With a single call to the app, it's easy. The app fetches the decided values in the correct order and returns the array ready to be used in the response.

ancazamfir: Not sure I understand. If you get a request for `h..H` you can have a loop and `HostMsg::GetDecidedValue` for `h`, `h+1`,...,`H` in order. The caller also has the `request_id`. Same can happen for parallel requests.

code/crates/test/tests/unit/sync.rs
ancazamfir: I meant something like this: 3c4740d9c4e59913c9db2488804ce09925b0eb34. Had these changes locally (off your PR branch) since last week, didn't look at the latest but just to give an idea.

code/crates/sync/src/handle.rs
ancazamfir: But with a batch request being received by a node only supporting single (old version) the requester should already be able to generalize to this case. Will check the changes again, maybe I misunderstood the compatibility impl.

code/crates/test/tests/it/value_sync.rs
ancazamfir: yeah there are probably some default params that could be set (like exec time per Tx) but we can leave this for now.

code/crates/test/tests/unit/sync.rs
hvanz: Thanks for the code. I adapted the tests to your style.

code/crates/sync/src/handle.rs
hvanz: The latest code now supports handling partial responses (that is, a prefix of the requested range of heights), so this is not an problem for trimming responses. Now I see two issues:
1. How to properly calculate the number of bytes of the RPC response. Before performing `Effect::SendValueResponse`, we want to measure each value that will go into the response to keep it under under `cfg.sync.max_response_size`.
2. `cfg.sync.max_response_size` default value is 512MiB, which may be too high. In practice, this value is overwritten by `cfg.p2p.rpc_max_size`, with 10MiB by default. But with big blocks (as in the starknet app) I've seen that nodes were timing out waiting for sync responses (in a local testnet without latency). Probably this should be left for a separate PR.

code/crates/app-channel/src/connector.rs
hvanz: In the end I reverted this API-breaking change. As suggested, values are now fetched one by one from the app with `HostMsg::GetDecidedValue` when handling `Effect::GetDecidedValues` in the engine.

code/crates/sync/src/config.rs
```diff
@@ -3,6 +3,7 @@ use std::time::Duration;
 use crate::scoring::Strategy;
 
 const DEFAULT_PARALLEL_REQUESTS: u64 = 5;
+const DEFAULT_BATCH_SIZE: usize = 1000;
```
ancazamfir: This is a very large default imo. I would make it O(10). With defaults for starknet we cannot sync, requests timeout.

code/crates/sync/src/handle.rs
ancazamfir: Could you open an issue and hopefully we can fix this quickly. With variable block sizes it is hard to have an optimal batch size. Having the response size capped would help a lot, also with UX. 
In the meantime let's make the default for `max_response_size` and `DEFAULT_BATCH_SIZE` lower.

ancazamfir: https://github.com/informalsystems/malachite/issues/1164

code/crates/sync/src/config.rs
ancazamfir: #1164

code/crates/config/src/lib.rs
```diff
@@ -457,11 +457,11 @@ impl Default for ValueSyncConfig {
             status_update_interval: Duration::from_secs(10),
             request_timeout: Duration::from_secs(10),
             max_request_size: ByteSize::mib(1),
-            max_response_size: ByteSize::mib(512),
+            max_response_size: ByteSize::mib(10),
             parallel_requests: 5,
```
romac, line 461: Do we want to keep parallel requests enabled by default or should we rely solely on batching?

ancazamfir: I think it's fine to keep both enabled by default.

code/crates/sync/src/state.rs
```diff
@@ -82,128 +67,96 @@ where
...
-    pub fn has_pending_value_request(&self, height: &Ctx::Height) -> bool {
-        self.pending_value_requests.contains_key(height)
+        range: &RangeInclusive<Ctx::Height>,
+    ) -> RangeInclusive<Ctx::Height> {
+        let start = max(self.tip_height.increment(), *range.start());
+        start..=*range.end()
     }
 
-    /// Check if a pending decided value request for a given height is in the `Validated` state.
-    pub fn is_pending_value_request_validated_by_height(&self, height: &Ctx::Height) -> bool {
```
ancazamfir, line 203: Just started to look at main merge (relevant changes coming from #1153) and noticed that in this PR all consensus validation checks are gone, i.e. `is_pending_value_request_validated_by..` are removed. Will need to understand the corner cases these were covering and what happens without them.

ancazamfir: If we restart a height when there are `max_parallel_requests` pending requests the node fails to sync. See test in 7975d296551387a35eb1c390b2891f87b43e3a4b. It passes with `batch_size: 1,`.
The simple solution here, given that this is a rare case we don't want to optimize for, may be to remove all pending requests on restart height.

In general we should have more tests with `batch_size > 1`, current default is `1`. Same for `parallel_requests`.

code/crates/config/src/lib.rs
```diff
@@ -454,10 +457,11 @@ impl Default for ValueSyncConfig {
             status_update_interval: Duration::from_secs(10),
             request_timeout: Duration::from_secs(10),
             max_request_size: ByteSize::mib(1),
-            max_response_size: ByteSize::mib(512),
+            max_response_size: ByteSize::mib(10),
```
nenadmilosevic95, line 460: What is happening if the response needs to be higher?

code/crates/sync/src/handle.rs
```diff
@@ -165,16 +209,23 @@ pub async fn on_started_height<Ctx>(
...
-    state.remove_pending_request_by_height(height_to_remove);
+    // The tip is the last decided value.
+    state.tip_height = height.decrement().unwrap_or_default();
+
+    // Garbage collect fully-validated requests.
+    state.remove_fully_validated_requests();
+
+    if restart {
+        // Consensus is retrying the height, so we should sync starting from it.
+        state.sync_height = height;
```
nenadmilosevic95, line 224: This is interesting, we essentially reset `sync_height` to possibly lower value? What is happening with already sent requests for heights > height?

code/crates/sync/src/handle.rs
```diff
@@ -190,9 +241,17 @@ pub async fn on_decided<Ctx>(
...
     Ctx: Context,
 {
-    debug!(height.tip = %height, "Updating request state");
+    debug!(%height, "Consensus decided on new value");
+
+    state.tip_height = height;
 
-    state.validate_response(height);
+    // Garbage collect fully-validated requests.
+    state.remove_fully_validated_requests();
```
nenadmilosevic95, line 251: We do this both here and in started_height?

code/crates/sync/src/handle.rs
```diff
@@ -190,9 +241,17 @@ pub async fn on_decided<Ctx>(
...
+
+    // The next height to sync should always be higher than the tip.
+    if state.sync_height == state.tip_height {
```
nenadmilosevic95, line 254: sync_height can't be < tip_height at any point in the execution?

code/crates/sync/src/handle.rs
```diff
@@ -202,24 +261,26 @@ pub async fn on_value_request<Ctx>(
...
     perform!(
         co,
-        Effect::GetDecidedValue(request_id, request.height, Default::default())
+        Effect::GetDecidedValues(request_id, request.range, Default::default())
     );
 
     Ok(())
 }
 
+/// Assumes that the range of values in the response matches the requested range
```
nenadmilosevic95: Or it is a prefix ?

code/crates/sync/src/handle.rs
```diff
@@ -295,78 +325,79 @@ pub async fn on_invalid_value_response<Ctx>(
...
+        error!(
+            "Received {} values from host, expected {batch_size}",
+            values.len()
+        )
+    }
+
+    // Validate the height of each received value
+    let mut height = *start;
+    for value in values.clone() {
+        if value.certificate.height != height {
```
nenadmilosevic95, line 372: it is assumed that host always returns values in the height order?

code/crates/sync/src/handle.rs
```diff
@@ -375,67 +406,75 @@ where
...
 where
     Ctx: Context,
 {
-    error!(%peer, height.sync = %height, "Error while processing value");
-
-    state.remove_pending_request_by_height(&height);
+    error!(%peer_id, %height, "Error while processing value");
 
     // NOTE: We do not update the peer score here, as this is an internal error
     //       and not a failure from the peer's side.
```
nenadmilosevic95, line 467: when this error happens?

code/crates/sync/src/handle.rs
```diff
@@ -444,101 +483,130 @@ async fn request_values<Ctx>(
...
+        warn!(%request_id, "All values in range {} have been validated, skipping re-request", DisplayRange::<Ctx>(&range));
+        return Ok(());
     }
 
+    let except_peer_id = match except_peer_id {
+        Some(peer_id) => {
+            if stored_peer_id == peer_id {
+                Some(peer_id)
+            } else {
+                warn!(%request_id, peer.actual = %peer_id, peer.expected = %stored_peer_id, "Received response from different peer than expected");
```
nenadmilosevic95: the only reason for this `if` is to log this warning?

code/crates/sync/src/state.rs
```diff
@@ -82,138 +67,96 @@ where
...
-            .collect::<Vec<_>>();
+            .map(|(peer, _)| (*peer, range.clone()))
+            .collect::<HashMap<_, _>>();
 
-        self.peer_scorer.select_peer(&peers, &mut self.rng)
+        // Prefer peers that have the whole range of values in their history.
+        if !peers_with_whole_range.is_empty() {
+            peers_with_whole_range
+        } else {
+            // Otherwise, just get the peers that can provide a prefix of the range.
```
nenadmilosevic95, line 96: I guess this can happen only if peer with a full range is in "except" peer?

code/crates/sync/src/state.rs
```diff
@@ -82,138 +67,96 @@ where
...
+        self.pending_requests
+            .iter()
+            .find(|(_, (range, _))| range.contains(&height))
+            .map(|(request_id, (_, stored_peer_id))| (request_id.clone(), *stored_peer_id))
     }
 
-    /// Remove a pending decided value request by its ID and return the height and peer it was associated with.
-    pub fn remove_pending_value_request_by_id(
+    /// Return a new range of heights, trimming from the beginning any height
+    /// that is not validated by consensus.
```
nenadmilosevic95:
```suggestion
    /// that is validated by consensus.
```

code/crates/engine/src/sync.rs
```diff
@@ -276,15 +281,30 @@ where
...
+                            None,
+                        )
+                        .await?
+                        .success_or(eyre!("Failed to get decided value for height {height}"))?;
+
+                    if let Some(value) = value {
+                        values.push(value);
+                    } else {
+                        warn!("Decided value not found for height {height}");
+                        break;
```
nenadmilosevic95, line 302: What is happening in this case?

code/crates/config/src/lib.rs
ancazamfir: @insumity is working on a separate PR to trim the response to fit into `max_response_size`

code/crates/sync/src/handle.rs
ancazamfir: currently they stay there...with this issue that they might prevent syncing the restart height.

code/crates/sync/src/handle.rs
ancazamfir: yes, in this PR. In main we mark the height as "decided" and in this way we will not request again between decided and start next height.
Here we move the heights to ensure that.

code/crates/sync/src/handle.rs
ancazamfir: Prefix is not valid, the response must keep the range and may reply with fewer values.

code/crates/sync/src/handle.rs
ancazamfir: yes. Info like this should be in the sync docs. Not sure it is.

code/crates/sync/src/handle.rs
ancazamfir: This is for internal errors. For now there is only one case.

When consensus processes the commit certificate one of the following errors may happen:
```
    /// The validator set was not found at the given height.
    #[error("Validator set not found at height {0}")]
    ValidatorSetNotFound(Ctx::Height),

    /// The certificate is invalid.
    #[error("Invalid certificate: {1}")]
    InvalidCommitCertificate(CommitCertificate<Ctx>, CertificateError<Ctx>),
```
The first one is an "internal" error and it will propagate to`on_value_processing_error` handler 
The second one is an "external" error, peer's assumed fault, so we process with `on_invalid_value`, penalize the peer, etc.

code/crates/sync/src/state.rs
ancazamfir: or there are no peers with full range at all, e.g. peers that have pruned some or all of the range heights.

code/crates/engine/src/sync.rs
ancazamfir: I think the only time this can happen if the app has pruned the height. It should be on the first height of the range unless, in between host requests there are decisions and some of these heights are pruned.

One thing to check here is if we should send empty response or not response at all.

code/crates/sync/src/handle.rs
nenadmilosevic95: What do you mean by prefix is not valid? I meant that here we assume that responses match the requested range, or it is a prefix of a range

code/crates/sync/src/handle.rs
nenadmilosevic95: I see Romain fixed this in the new commit

code/crates/sync/src/state.rs
nenadmilosevic95: Hmm, but if they pruned, they will prune the earlier heights so I guess they will not have the prefix of the range?

code/crates/sync/src/handle.rs
ancazamfir: Sorry which commit?

code/crates/sync/src/handle.rs
ancazamfir: I was wrong, yes it can be a prefix. The comment is wrong and should be fixed, clarified.

code/crates/sync/src/handle.rs
nenadmilosevic95: This one [e9fd8fd5f6](https://github.com/informalsystems/malachite/pull/1130/commits/e9fd8fd5f6157f3b64cb93b5363f32a0e749966a)

code/crates/sync/src/config.rs
```diff
@@ -52,18 +54,24 @@ impl Config {
...
 
 impl Default for Config {
     fn default() -> Self {
         Self {
             enabled: true,
             request_timeout: Duration::from_secs(10),
-            max_request_size: 1024 * 1024,        // 1 MiB
-            max_response_size: 512 * 1024 * 1024, // 512 MiB
+            max_request_size: 1024 * 1024,       // 1 MiB
+            max_response_size: 10 * 1024 * 1024, // 512 MiB
```
hvanz:
```suggestion
            max_response_size: 10 * 1024 * 1024, // 10 MiB
```

code/crates/sync/src/handle.rs
hvanz: Yes, the comment is wrong. I did not updated it when the definition of valid value changed. I fixed it.

code/crates/sync/src/state.rs
ancazamfir: update after syncing with @nenadmilosevic95. He will take care of this failure:
- clean the sync pending requests
- empty `state.msg_buffer`
- empty `state.consensus.input_queue`

nenadmilosevic95: Push these changes now the test is passing

ancazamfir: looks good, thanks!

code/crates/sync/src/handle.rs
ancazamfir: thanks @hvanz for the fixes here

ancazamfir (approved):
Looks good! Thanks all for the effort here.
We can ignore the CI failures, the issues are also in main, surfaced by recent CI toolchain version.

code/crates/engine/src/network.rs
```diff
@@ -334,7 +343,7 @@ where
             }
 
             Msg::NewEvent(Event::PeerConnected(peer_id)) => {
-                peers.insert(peer_id);
+                peers.insert(peer_id, Status::default());
```
romac: I can't find a place where we update a peer's status. If I didn't miss it, we could definitely fix it, but do we really need to track their status in the Network module in the first place?

hvanz: Good catch! This code actually comes from the first implementation and now it's not used anymore. I restored it to its original state.

romac (approved):

Issue #1136, opened 2025-07-17
chore(code): Parallel sync issues


ancazamfir:
cc @hvanz 

 Assume a syncing node starts at `h` and the network is at `H`. 
- The node will ask for decided values `h, h+1, ...`. The consensus engine sends the value to the app and the certificate to the consensus core. 
- In the same time the node receives proposal (parts) for `H, H+1,...`. 
- The app/ host stores the undecided values for `H, H+1,...`. Here our example apps are not doing great as they send proposed values to consensus (always `Valid`)
- When the sync value for `H` is eventually received, we send the value (block) to the app and we attempt to process the certificate. With parallel requests this is likely buffered. The starknet app doesn't store the value but it signals consensus that it received and accepted the value as `Valid` (again buffered). The other apps do store the value but otherwise do the same thing. 
  - A real app at height `h` cannot determine the validity of a block at `H` (issue 1). So we need to redesign some stuff here: either we buffer in consensus the full sync value, meaning the value and the certificate, and replay the same way we do certificates. Or we put some burden on the app to store synced blocks as they come and validate and notify consensus when possible.

- When the app moves to `H`, consensus will find the synced value in the buffered input and replay it. The problem is that when we send the sync value to the app we set the proposer as our own address and not the proposer for the height and round (issue 2)  
- Then, on start new round `r`, the app will replay the (undecided) value for `H` and `r` that it got from the parts. 
- So the driver gets two proposed values for `H` (same value/ hash) but from two different proposers and it panics (#1135)

Next steps proposal:
- [x] (#1141)Fix the proposer setting in the engine: this line https://github.com/informalsystems/malachite/blob/b55cfe9dfd97ab937f7e4219a55b0350b6d2e159/code/crates/engine/src/consensus.rs#L523 should probably read:
	```validator_address: state.consensus.get_proposer(certificate_height, certificate_round).clone(),``` (@romac)
  - Note: cannot yet reproduce this with instructions from #1135 to tell for sure this is the fix.
- [x] Fix our example apps for the cases where a value (either from parts or sync) is received early and cannot be processed yet. No `ProposedValue` should be send to consensus. 
   - alternatively buffer future app values and parts. The problem with parts is that we currently don't have a `height()` method.  (@romac)
   - In general we need to show code in the app that suggests a good implementation.
   - We decided to do the following:
      - [x] (#1144) Buffer parts for future heights in the app for now
      - [x] (#1149) Buffer the sync value in consensus
[...]

PR #1143, opened 2025-07-25
fix(code/sync): Derive the consensus queue capacity from the `sync.parallel_requests` setting

romac:
Part of: #1136 

This is required in order to ensure that consensus inputs yielded by the sync module cannot overflow the queue capacity and therefore dropped

---

#### For all contributors

### DIFF
--- a/code/crates/starknet/host/src/node.rs
+++ b/code/crates/starknet/host/src/node.rs
@@ -251,7 +251,7 @@ fn make_config(index: usize, total: usize, settings: MakeConfigSettings) -> Conf
         moniker: format!("starknet-{index}"),
         consensus: ConsensusConfig {
             value_payload: ValuePayload::PartsOnly,
-            queue_capacity: 100,
+            queue_capacity: 100, // Deprecated, derived from `sync.parallel_requests`
             timeouts: TimeoutConfig::default(),
             p2p: P2pConfig {
                 protocol: PubSubProtocol::default(),
@@ -347,7 +347,7 @@ fn make_distributed_config(
     Config {
         moniker: format!("starknet-{index}"),
         consensus: ConsensusConfig {
-            queue_capacity: 100,
+            queue_capacity: 100, // Deprecated, derived from `sync.parallel_requests`
             value_payload: ValuePayload::PartsOnly,
             timeouts: TimeoutConfig::default(),
             p2p: P2pConfig {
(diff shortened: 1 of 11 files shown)


code/crates/starknet/host/src/node.rs
```diff
@@ -251,7 +251,7 @@ fn make_config(index: usize, total: usize, settings: MakeConfigSettings) -> Conf
         moniker: format!("starknet-{index}"),
         consensus: ConsensusConfig {
             value_payload: ValuePayload::PartsOnly,
-            queue_capacity: 100,
+            queue_capacity: 100, // Deprecated, derived from `sync.parallel_requests`
```
ancazamfir, line 254: Can't we just remove it for our test apps?

romac: This field is part of the `ConsensusConfig` defined in the `config` crate, that is passed down to the `Consensus` actor, hence we cannot remove it altogether. What we could do is:
1. Copy the `ConsensusConfig` type into the `engine` crate (with the `queue_capacity` field)
2. Remove the `queue_capacity` from the `config::ConsensusConfig`
3. Fill in `engine::ConsensusConfig::queue_capacity` when we convert between `config::ConsensusConfig` and `engine::ConsensusConfig`

Or we can leave it as is, which I agree is not the cleanest.

What do you think?

ancazamfir: > ...hence we cannot remove it altogether.

Not sure I understand why. Is it for backwards compatibility? Not sure if anyone has deployed yet with `queue_capacity` other than snapchain upgrade attempt which is halted.
If it can't be removed, yes, let's do what you propose above.

romac: It's just that if we remove it then we need to pass it alongside the config to the Consensus actor instead of as part of the config, which makes yet one more parameter to the supply to the actor. It would be nicer to keep it in the config struct.

ancazamfir: I see. But when we say "This setting is deprecated and will be removed in the future." what is the final solution? The 1..3 steps you describe above?

ancazamfir: Ok, given the time constraints, I propose we keep things the way they are for now and figure out in the next PR. We can release with this for now if that doesn't make it in time.

ancazamfir (approved):

code/crates/starknet/host/src/node.rs
ancazamfir: #1150

PR #1144, opened 2025-07-25
feat(app): Buffer proposals for future heights

romac:
Closes: Partial #1136

---

#### For all contributors

### DIFF
--- a/code/crates/sync/src/handle.rs
+++ b/code/crates/sync/src/handle.rs
@@ -173,7 +173,8 @@ where
     state.sync_height = height;
     state.tip_height = tip_height;
 
-    state.remove_pending_request_by_height(&tip_height);
+    let height_to_remove = if restart { &height } else { &tip_height };
+    state.remove_pending_request_by_height(height_to_remove);
 
     // Trigger potential requests if possible.
     request_values(co, state, metrics).await?;
--- a/code/examples/channel/src/state.rs
+++ b/code/examples/channel/src/state.rs
@@ -134,7 +134,7 @@ impl State {
                 part.height = %parts.height,
                 part.round = %parts.round,
                 part.sequence = %sequence,
-                "Received outdated proposal part, ignoring"
+                "Received outdated proposal, ignoring"
             );
 
             return Ok(None);
@@ -167,15 +167,18 @@ impl State {
             }
         }
 
+        let proposal_height = parts.height;
+
         // Re-assemble the proposal from its parts
         let value = assemble_value_from_parts(parts)?;
 
-        info!(
-            "Storing undecided proposal {} {}",
-            value.height, value.round
-        );
-
-        self.store.store_undecided_proposal(value.clone()).await?;
+        if proposal_height == self.current_height {
+            info!(%value.height, %value.round, "Storing undecided proposal");
+            self.store.store_undecided_proposal(value.clone()).await?;
+        } else {
+            info!(%value.height, %value.round, "Storing pending proposal");
+            self.store.store_pending_proposal(value.clone()).await?;
+        }
 
         Ok(Some(value))
     }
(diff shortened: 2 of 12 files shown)


ancazamfir:
I looked at the failures, there are a couple of issues I see so far. Let's say the current height is `h` and we receive parts and votes for `H` which are buffered.
- we store `H` in pending
- node reaches `H`
- app replays `H` value, i.e. sends `ProposedValue` for `H`
- consensus replays `H` messages, decides on `H`, sends `Decided` to the app
- app fails the commit because the value is not in undecided. This **commit error** is seen:
```
Commit failed: Trying to commit a value with value id 12561 at height 3 and round 1 for which there is no proposal
```
- app restarts height `H`
- sync has a pending request for `H` and never asks for `H` again
- **node is stuck**

There are two problems/ fixes:
- **node is stuck**: when a node restarts a height `h`, sync removes pending for `h-1`. It should remove the pending request for `h` and not `h-1`. With this fix the tests should pass as with commit error the undecided value will come via sync. (8eb6972ccbd391f63563f83d6ca165ecb440afc4)
- **commit error**: when the replays from pending it should also move it to undecided so when the value is decided, it will find it and commit is successful (16371facaf2f436edb4927b2d7b351bc488f562d)

ancazamfir:
One thought, we could add future values to undecided as Invalid, but not notify consensus, when `Decided` is received do not prune higher heights from undecided, and on next height revalidate the undecided for that next height and send to consensus (regardless of validity). This way a pending proposal table is not needed.

romac:
> One thought, we could add future values to undecided as Invalid, but not notify consensus, when Decided is received do not prune higher heights from undecided, and on next height revalidate the undecided for that next height and send to consensus (regardless of validity). This way a pending proposal table is not needed.

Instead of using `Invalid`, we can also add a flag to the undecided table to indicate whether or not the proposal is pending, effectively storing `(ProposedValue, ValueType)` with `enum ValueType { Pending, Current }`. What do you think?

ancazamfir:
> > One thought, we could add future values to undecided as Invalid, but not notify consensus, when Decided is received do not prune higher heights from undecided, and on next height revalidate the undecided for that next height and send to consensus (regardless of validity). This way a pending proposal table is not needed.
> 
> Instead of using `Invalid`, we can also add a flag to the undecided table to indicate whether or not the proposal is pending, effectively storing `(ProposedValue, ValueType)` with `enum ValueType { Pending, Current }`. What do you think?

Yes that would be clean! Should we do this in a separate PR?

code/examples/channel/src/store/keys.rs
```diff
@@ -4,6 +4,7 @@ use malachitebft_app_channel::app::types::core::Round;
 use malachitebft_test::{Height, ValueId};
 
 pub type UndecidedValueKey = (HeightKey, RoundKey, ValueIdKey);
+pub type PendingValueKey = (HeightKey, RoundKey, ValueIdKey);
```
nenadmilosevic95, line 7: Can we just rename this to `ValueKey` and keep just one?

code/examples/channel/src/state.rs
```diff
@@ -167,15 +167,18 @@ impl State {
...
-        );
-
-        self.store.store_undecided_proposal(value.clone()).await?;
+        if proposal_height == self.current_height {
+            info!(%value.height, %value.round, "Storing undecided proposal");
+            self.store.store_undecided_proposal(value.clone()).await?;
+        } else {
+            info!(%value.height, %value.round, "Storing pending proposal");
+            self.store.store_pending_proposal(value.clone()).await?;
+        }
```
nenadmilosevic95, line 181: Minor: `undecided` and `pending` are bit ambiguous for me, I would rather use something like `current` and `future`

code/examples/channel/src/state.rs
```diff
@@ -200,11 +203,17 @@ impl State {
...
+        let mut proposal = self
             .store
             .get_undecided_proposal_by_value_id(value_id)
-            .await
-        else {
+            .await;
+
+        if !matches!(proposal, Ok(Some(_))) {
+            // If we didn't find an undecided proposal, try to get a pending one
+            proposal = self.store.get_pending_proposal_by_value_id(value_id).await
```
nenadmilosevic95, line 213: What is the scenario when this is going to happen?

code/crates/sync/src/handle.rs
```diff
@@ -173,7 +173,8 @@ where
     state.sync_height = height;
     state.tip_height = tip_height;
 
-    state.remove_pending_request_by_height(&tip_height);
+    let height_to_remove = if restart { &height } else { &tip_height };
+    state.remove_pending_request_by_height(height_to_remove);
```
nenadmilosevic95, line 177: Here we check if it's a restart. If so, we can't remove the previous height (tip height) since we don't have it? However, we do remove the pending request for the current height. But will there actually be a pending request in this case, given that it's a restart?

nenadmilosevic95 (commented):
LGTM, left few comments

Issue #865, opened 2025-02-19
Vote collection trait constraint and potential optimization


adizere:
## Problem statement

The [`PerRound`](https://github.com/informalsystems/malachite/blob/a8be71757b35cfe878f28076ddcfe5eb7480f106/code/crates/core-votekeeper/src/keeper.rs#L41) struct in the vote keeper seems to use a `BTreeSet` to collect multiple `SignedVote`s within a height+round.

It seems like the operations with this set are the constraint that require a [`Vote` to satisfy the `Ord` trait](https://github.com/informalsystems/malachite/blob/a8be71757b35cfe878f28076ddcfe5eb7480f106/code/crates/core-types/src/vote.rs#L21). 

The operations on this set are as follows, as of a8be71757b35cfe878f28076ddcfe5eb7480f106 after brief skimming through the codebase:

1. [insertion](https://github.com/informalsystems/malachite/blob/a8be71757b35cfe878f28076ddcfe5eb7480f106/code/crates/core-votekeeper/src/keeper.rs#L107) of some vote
    - frequency: happens ~O(N) times per height and round assuming N validators - i.e., very frequently
2. [traversal of the whole set of collected votes](https://github.com/informalsystems/malachite/blob/a8be71757b35cfe878f28076ddcfe5eb7480f106/code/crates/core-votekeeper/src/keeper.rs#L120) to find a vote, where duplicates are identified by `vote_type` and `validator_address`
    - frequency: same as (1), for every received vote, ~O(N) times within a height and round
3. [traversal and cloning of the whole set of collected votes](https://github.com/informalsystems/malachite/blob/a8be71757b35cfe878f28076ddcfe5eb7480f106/code/crates/core-consensus/src/state.rs#L144) to be able to send the set of votes to a peer
    - frequency: this should happen very rarely, whenever a peer requests `VoteSetRequest`

This data structure is probably far from being a bottleneck, and unclear how much an optimization would help with performance. However -- it would be good to avoid doing naive things such as it happens at (2). I might miss something and maybe that pattern is necessary, but it should be investigated, because we impose the trait constraint `Ord` on `Vote`s but it doesn't seem like we exploit this property to easily find a Vote in a set.

### Definition of done for this issue

- [ ] confirm the above investigation - is indeed `Ord` required just so we can `insert` each `Vote` in a `BTreeSet` and then couldn't we employ `contains`?
- [ ] remove the trait constraint `Ord` if the core code is not making use of it, or 
[...]

PR #1148, opened 2025-07-29
fix(code/core): Remove `Ord` constraint from `Vote` trait

insumity:
Closes: #865 

---

**Description**
Issue #865 states that although we use the `Ord` constraint for `Vote`s and we store `received_votes` in a `BTreeSet`, we do not actually utilize of the vote ordering property for anything. For example, when calling `get_vote` we still traverse through the whole set of votes to find a specific vote. Indeed we do not use the vote ordering for anything.
Moreover, after the [removal of the vote-set-based synchronization](https://github.com/informalsystems/malachite/pull/1008), some methods like `restore_votes` are not used anymore.

**Solution**

1. We removed the `Ord` constraints from `Vote` and use `Vec` (instead of `BTreeSet`) for `received_votes`.

2. We performed some cleaning up and removed the unused `restore_votes`, and `round_range_inclusive` methods.

This PR should not contain any breaking changes because we remove (and do not introduce) a constraint from `Vote`.

#### For all contributors


code/crates/core-consensus/src/handle/vote.rs
```diff
@@ -76,17 +76,14 @@ where
 
     debug_assert_eq!(consensus_height, vote_height);
 
-    // Only append to WAL and store the non-nil precommit if we have not yet seen this vote.
-    if !state.driver.votes().has_vote(&signed_vote) {
```
romac, line 80: Let's not remove this condition here, we want to avoid storing duplicate votes in the WAL for performance reasons: updating the WAL is expensive because it will issue a `fsync` syscall on the WAL file before consensus can resume and those are notoriously slow and expensive.

romac (approved):

PR #1154, opened 2025-08-04
fix(qa): Mount source directory as read-write

romac:
Without this we get such failures when running QA:

```#14 [builder 3/4] WORKDIR /mnt
#15 [builder 4/4] RUN --mount=type=bind,from=code,target=/mnt cargo build --release --target-dir /tmp
#15 0.273     Updating crates.io index
#15 2.883 error: failed to write /mnt/Cargo.lock
#15 2.883 
#15 2.883 Caused by:
#15 2.883   failed to open: /mnt/Cargo.lock
#15 2.883 
#15 2.883 Caused by:
#15 2.883   Read-only file system (os error 30)
```

https://github.com/informalsystems/malachite/actions/runs/16723245984/job/47332329264

### DIFF
--- a/qa/docker/Dockerfile
+++ b/qa/docker/Dockerfile
@@ -3,7 +3,7 @@
 FROM rust AS builder
 RUN apt-get update && apt-get install -y protobuf-compiler
 WORKDIR /mnt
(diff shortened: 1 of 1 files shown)


romac:
Works now: https://github.com/informalsystems/malachite/actions/runs/16726045274/job/47341834904

Issue #1164, opened 2025-08-06
feat(code): Limit the number of values in sync response to `cfg.sync.max_response_size`


ancazamfir:
This is an issue remaining after #1070 

The syncing node may request a value range for which the response exceeds the `cfg.p2p.rpc_max_size`.
`cfg.p2p.rpc_max_size` is sometimes used to overwrite `cfg.sync.max_response_size`.
The number of values per batch can be configured with `sync_cfg.batch_size`. 

This issue is to address the following:
- Ensure that the response payload does not exceed:
  - the number of values required: i.e. if `n` values are requested do not respond with `N>n` values, AND
  - the `min(cfg.p2p.rpc_max_size, cfg.p2p.rpc_max_size)` in byte size

- Make sure that the syncing node does not penalize peers that respond with less than `n` values

See also these [comments](https://github.com/informalsystems/malachite/pull/1130#discussion_r2214400980)

PR #1171, opened 2025-08-12
feat(code): limit the number of values in a sync response based on upper limit

insumity:
Closes: #1164 

We introduce a check so that a peer only sends values of at most `cfg.sync.max_response_size`.
The retrieval of the right amount of values is performed when handling the `Effect::GetDecidedValues` effect in the engine and not in `sync` in order to avoid sending a huge amount of values to `sync` and then have `sync` do the retrieval.

We manually tested these changes by running the tests from the `value_sync.rs` file and looking at the logs. With testing, we identified an issue where in case a syncing peer got a partial response (i.e., some of the values) back, it immediately re-requests the exact same range - because the partial values are not yet processed by the consensus engine - flooding the network. We fixed this issue in this PR.

---

#### For all contributors

### DIFF
--- a/code/crates/sync/src/handle.rs
+++ b/code/crates/sync/src/handle.rs
@@ -338,9 +338,9 @@ where
                 state.update_request(request_id, peer_id, updated_range);
             }
 
-            // issue a new request to the same peer for the remaining values
+            // Issue a new request to any peer, not necessarily the same one, for the remaining values
             let new_range = new_start..=end;
-            request_values_from_peer(&co, state, metrics, new_range, peer_id).await?;
+            request_values_range(co, state, metrics, new_range).await?;
         }
     }
 
@@ -538,6 +538,41 @@ where
 
         request_values_from_peer(&co, state, metrics, range, peer).await?;
     }
+
+    Ok(())
+}
+
+/// Request multiple batches of values in parallel.
+async fn request_values_range<Ctx>(
+    co: Co<Ctx>,
+    state: &mut State<Ctx>,
+    metrics: &Metrics,
+    range: RangeInclusive<Ctx::Height>,
+) -> Result<(), Error<Ctx>>
+where
+    Ctx: Context,
+{
+    let max_parallel_requests = state.max_parallel_requests();
+
+    if state.pending_requests.len() as u64 >= max_parallel_requests {
+        info!(
+            %max_parallel_requests,
+            pending_requests = %state.pending_requests.len(),
+            range = %DisplayRange::<Ctx>(&range),
+            "Maximum number of parallel requests reached, skipping request for values"
+        );
+
+        return Ok(());
+    };
+
+    // Get a random peer that can provide the values in the range.
+    let Some((peer, range)) = state.random_peer_with(&range) else {
+        // No connected peer reached this height yet, we can stop syncing here.
+        debug!(range = %DisplayRange::<Ctx>(&range), "No peer to request sync from");
+        return Ok(());
+    };
+
+    request_values_from_peer(&co, state, metrics, range, peer).await?;
 
     Ok(())
 }


kirdatatjana (approved):

code/crates/sync/src/handle.rs
```diff
@@ -311,9 +311,35 @@ where
...
+            let end = *requested_range.end();
+
+            if response.values.is_empty() {
+                error!(%request_id, %peer_id, "Received response contains no values");
+            } else {
+                // the response of this request only provided `response.values.len()` values, so update the pending request accordingly
+                let updated_range = *requested_range.start()..=new_start.decrement().unwrap();
+                state
+                    .pending_requests
+                    .insert(request_id.clone(), (updated_range, peer_id));
```
hvanz: suggestion: put all this logic in a state function `state.update_request(request_id, ...)` or similar

code/crates/sync/src/handle.rs
```diff
@@ -311,9 +311,35 @@ where
...
+            }
+
+            // issue a new request to the same peer for the remaining values
+            let new_range = new_start..=end;
+            request_values_from_peer(&co, state, metrics, new_range, peer_id).await?;
```
hvanz: This works but we could request the remaining values from any random peer that has them, not necessarily the same peer.

code/crates/engine/src/sync.rs
```diff
@@ -283,8 +283,21 @@ where
             }
 
             Effect::GetDecidedValues(request_id, range, r) => {
+                // For simplicity, to avoid asking the application for the exact size of addresses, signatures, etc.
+                // we assume some maximum byte sizes for an address and a signature in order to calculate
+                // an **approximate** total size per value, so that we only send at most `self.sync_config.max_response_size`
+                // to another peer.
+                const MAX_BYTES_PER_ADDRESS: usize = 32;
+                const MAX_BYTES_PER_SIGNATURE: usize = 100;
```
hvanz, line 291: I would declare these variables close to the declarations of the address and signature types. Also, each app will have its own definition of address and signature.

code/crates/sync/src/handle.rs
insumity: Yes, we can request from any peer. There are currently 3 methods that can be used to request values from another peer: `request_values_from_peer`, `request_values`, and `re_request_values_from_peer_except`. I use `request_from_values_from_peer` here because it is the only method out of the 3 that has a `range` parameter.

code/crates/engine/src/sync.rs
insumity: Not sure I fully follow your comment: Are you suggesting contacting the app to receive the sizes of addresses and signatures? If not, where exactly would you declare those constants?

romac: We could add them as static const to the `Address` and `SigningScheme` traits.

code/crates/sync/src/handle.rs
```diff
@@ -311,9 +311,35 @@ where
                 "Received response from different peer than expected"
```
romac, line 311: We should probably use this PR to abort processing the response in the case where we got a response from another peer than the one we expected.

code/crates/sync/src/handle.rs
romac: I am on it.

code/crates/engine/src/sync.rs
romac: I am on it.

romac: But that's pretty brittle as well, and could break things if the upper bound is not tight enough, eg. some application uses a signature scheme with variable-size signatures and defines `Signature::MAX_SIZE` to a large value just in case.

romac: The best way would probably be to encode the commit certificate (or even the full sync response) using the `(Network)Codec` and use the actual computed size. Not great for performance, we should probably either measure the impact upfront, or go with it and only optimize this if we see too big a perf impact. What do you think?

romac:
It would be good to add an integration test that exercises the new [code path](https://app.codecov.io/github/circlefin/malachite/pull/1171#561861bcefa993d7210649dc45a64627-R326).

code/crates/engine/src/sync.rs
insumity: >The best way would probably be to encode the commit certificate (or even the full sync response) using the (Network)Codec and use the actual computed size.

I agree this is the right away to go about it. I'll rewrite the PR to take this into account.

insumity:
Thanks you very much for the updates and your comments @romac.

> It would be good to add an integration test that exercises the new [code path](https://app.codecov.io/github/circlefin/malachite/pull/1171#561861bcefa993d7210649dc45a64627-R326).

I'll do that, and will also look into the network codec to get the actual size as you suggest. Just a note that I’ll probably only be able to look into this tomorrow afternoon and early next week (doubt I can finish it by tomorrow).

romac:
>  Just a note that I’ll probably only be able to look into this tomorrow afternoon and early next week (doubt I can finish it by tomorrow).

No problem, thank you!

romac:
Superseded by #1184

PR #1174, opened 2025-08-13
feat(code): allow nodes to run in sync-only mode without participating in consensus

ancazamfir:
Closes: #XXX
Adds new Sync-Only mode for Full Nodes

Behavior:

- Subscribes only to sync network topics
- Downloads and validates blocks
- Does not process consensus messages (proposals, votes, certificates)
- Does not participate in voting or proposal validation

Use Cases: 
- Block explorers, read-only applications, archival nodes
- Validators that are far behind
   - The UX is not great: starts with `consensus.enable=false`, when close to the tip, the validator needs to be manually restarted with updated config: `consensus.enable=true`
   - For this use case we should have config reload or even better rejoining consensus when getting close to the tip

---

#### For all contributors

### DIFF
--- a/code/crates/test/framework/src/node.rs
+++ b/code/crates/test/framework/src/node.rs
@@ -12,6 +12,7 @@ use malachitebft_test::middleware::{DefaultMiddleware, Middleware};
 use crate::Expected;
 
 pub type NodeId = usize;
+pub type ConfigCustomizer<Config> = Arc<dyn Fn(&mut Config) + Send + Sync>;
 
 pub enum Step<Ctx, S>
 where
@@ -48,6 +49,7 @@ where
     pub steps: Vec<Step<Ctx, State>>,
     pub state: State,
     pub middleware: Arc<dyn Middleware>,
+    pub config_customizer: Option<ConfigCustomizer<malachitebft_test_app::config::Config>>,
 }
 
 impl<Ctx, State> TestNode<Ctx, State>
@@ -70,6 +72,7 @@ where
             steps: vec![],
             state,
             middleware: Arc::new(DefaultMiddleware),
+            config_customizer: None,
         }
     }
 
@@ -308,6 +311,13 @@ where
 
     pub fn full_node(&mut self) -> &mut Self {
         self.voting_power = 0;
+        self
+    }
+
+    pub fn with_consensus_disabled(&mut self) -> &mut Self {
+        self.config_customizer = Some(Arc::new(|config| {
+            config.consensus.enabled = false;
+        }));
         self
     }
 
--- a/code/crates/test/tests/it/value_sync.rs
+++ b/code/crates/test/tests/it/value_sync.rs
@@ -343,6 +343,47 @@ pub async fn start_late_rotate_epoch_validator_set() {
         .await
 }
 
+#[tokio::test]
+pub async fn sync_only_fullnode_without_consensus() {
+    const HEIGHT: u64 = 8;
+
+    let mut test = TestBuilder::<()>::new();
+
+    // First two nodes are normal validators that will drive consensus
+    test.add_node()
+        .with_voting_power(10)
+        .start()
+        .wait_until(HEIGHT)
+        .success();
+
+    test.add_node()
+        .with_voting_power(10)
+        .start()
+        .wait_until(HEIGHT)
+        .success();
+
+    // Third node is a sync-only full node (0 voting power, consensus disabled)
+    // It should be able to sync values but not participate in consensus
+    test.add_node()
+        .with_voting_power(0) // Non-validator
+        .with_consensus_disabled() // Use the new per-node consensus disable
+        .start_after(1, Duration::from_secs(5)) // Start late to force syncing
+        .wait_until(HEIGHT)
+        .success();
+
+    test.build()
+        .run_with_params(
+            Duration::from_secs(45),
+            TestParams {
+                enable_value_sync: true,
+                // consensus_enabled defaults to true for other nodes
+                parallel_requests: 3,
+                ..Default::default()
+            },
+        )
+        .await
+}
+
 #[derive(Debug)]
 struct ResetHeight {
     reset_height: u64,
--- a/code/crates/app/src/spawn.rs
+++ b/code/crates/app/src/spawn.rs
@@ -226,6 +226,7 @@ fn make_gossip_config(cfg: &ConsensusConfig) -> NetworkConfig {
         channel_names: ChannelNames::default(),
         rpc_max_size: cfg.p2p.rpc_max_size.as_u64() as usize,
         pubsub_max_size: cfg.p2p.pubsub_max_size.as_u64() as usize,
+        enable_consensus: cfg.enabled,
         enable_sync: true,
     }
 }
--- a/code/crates/config/src/lib.rs
+++ b/code/crates/config/src/lib.rs
@@ -495,6 +495,13 @@ impl FromStr for ScoringStrategy {
 /// Consensus configuration options
 #[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
 pub struct ConsensusConfig {
+    /// Enable consensus protocol participation
+    ///
+    /// When disabled, the node only runs the synchronization protocol
+    /// and does not subscribe to consensus-related topics
+    #[serde(default)]
+    pub enabled: bool,
+
     /// Timeouts
     #[serde(flatten)]
     pub timeouts: TimeoutConfig,
--- a/code/crates/engine/src/util/output_port.rs
+++ b/code/crates/engine/src/util/output_port.rs
@@ -178,7 +178,7 @@ impl OutputPortSubscription {
 /// `From<T>` to convert the published message type to its own message format.
 ///
 /// # Example
-/// ```
+/// ```ignore
 /// // First, define the publisher's message types, including a variant for
 /// // subscribing `OutputPortSubscriber`s and another for publishing messages:
 /// use ractor::{
@@ -192,7 +192,6 @@ impl OutputPortSubscription {
 ///     Subscribe(OutputPortSubscriber<u8>), // Message type for subscribing an actor to the output port
 /// }
 ///
-/// #[cfg(feature = "cluster")]
 /// impl Message for PublisherMessage {
 ///     fn serializable() -> bool {
 ///         false
@@ -251,7 +250,6 @@ impl OutputPortSubscription {
 ///     Handle(String), // Subscriber's intent for message handling
 /// }
 ///
-/// #[cfg(feature = "cluster")]
 /// impl Message for SubscriberMessage {
 ///     fn serializable() -> bool {
 ///         false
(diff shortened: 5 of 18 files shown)


code/crates/test/tests/it/value_sync.rs
```diff
@@ -343,6 +343,47 @@ pub async fn start_late_rotate_epoch_validator_set() {
...
+    test.add_node()
+        .with_voting_power(10)
+        .start()
+        .wait_until(HEIGHT)
+        .success();
+
+    // Third node is a sync-only full node (0 voting power, consensus disabled)
+    // It should be able to sync values but not participate in consensus
+    test.add_node()
+        .with_voting_power(0) // Non-validator
```
romac: We can use the `full_node` method to make this more explicit:

```suggestion
        .full_node()
```

code/crates/test/framework/src/node.rs
```diff
@@ -311,6 +314,13 @@ where
         self
     }
 
+    pub fn with_consensus_disabled(&mut self) -> &mut Self {
+        self.config_customizer = Some(Arc::new(|config| {
+            config.consensus.enabled = false;
+        }));
```
romac: Just a note that if we ever add more config customizations, we'll need to either change `config_customizer` to `Vec<ConfigCustomizer<Config>>` or be careful to compose any additional customization function with the existing one, if any.

code/crates/test/framework/src/node.rs
```diff
@@ -48,6 +49,7 @@ where
     pub steps: Vec<Step<Ctx, State>>,
     pub state: State,
     pub middleware: Arc<dyn Middleware>,
+    pub config_customizer: Option<ConfigCustomizer<malachitebft_test_app::config::Config>>,
```
romac: We can maybe get rid of the `Option` here by setting this field to `Arc::new(|config| config)` by default.

romac:
> Validators that are far behind
> * The UX is not great: starts with `consensus.enable=false`, when close to the tip, the validator needs to be manually restarted with updated config: `consensus.enable=true`
> * For this use case we should have config reload or even better rejoining consensus when getting close to the tip

Agreed, let's go with the latter option for now. For this we might want to add an internal config option to instruct the engine to catch up via sync and then re-enable consensus when close enough to the tip (for some definition of "close enough" that we have to figure out).

This option could be enabled by starting the node with a command-line flag that would propagate down the engine. What do you think?

hvanz:
I also like the idea of enabling consensus when the node is close to the tip. This could be by default for all nodes, including validators. 

Now, how to determine when a node is close to the tip? Querying the network for the latest height and passing it to the node when it starts could be an option. But this height can become outdated if the chain is very advanced and catching up takes, let's say, hours. An alternative is for the node to enable consensus when it's getting "close enough" to the max tip of all its peers, even when we don't know if those peers are validators.

Probably this can be left for a separate PR.

code/crates/test/framework/src/node.rs
ancazamfir: 9f4b24a59e5cb466dab6fce43984749f45014fb1

ancazamfir:
cc @ams9198
@romac 
> Agreed, let's go with the latter option for now. For this we might want to add an internal config option to instruct the engine to catch up via sync and then re-enable consensus when close enough to the tip (for some definition of "close enough" that we have to figure out).
> 
> This option could be enabled by starting the node with a command-line flag that would propagate down the engine. What do you think?

...
@hvanz 
> I also like the idea of enabling consensus when the node is close to the tip. This could be by default for all nodes, including validators.

....
@nenadmilosevic95 
> Can we achieve similar thing by subscribing and unsubscribing to a consensus topic somehow?

I thought about this, started to implement something and the changes are more complex than at first look.

The problem with automatic enable/disable for consensus is that the sync protocol is currently not secure. The sync Status has no proofs, open to sybil attacks, any node can claim it’s at a height high enough to keep consensus disabled forever. 

At least we should have some heuristics to detect false tip advertisement. Right now, if byzantine nodes fail to provide blocks beyond tip height, we timeout and adjust scores. But we still try to sync from them. We should either blacklist or switch to consensus if we haven’t made progress for a “long” time (tbd). So we need to think about how we should adjust  the scoring on response timeouts. There are other attack surfaces.

We can imagine other ways to do it (e.g. use proofs, percentile median status heights etc.), but these need to be spec-ed imo. @nenadmilosevic95 do you want to drive/ work on the spec for this?

ancazamfir:
Full nodes, regardless being in sync-only mode or consensus enabled (main repo behavior) are still participating in consensus operations: broadcasting liveness messages, requesting values from app, restreaming. This prevents clean full-node operation. So we need to add some more checks.

Drafted two approaches for filtering:
- Engine-level approach
https://github.com/ancazamfir/malachite/compare/feat/fullnode-sync-only...engine-filter

- Consensus-core approach  
https://github.com/ancazamfir/malachite/compare/feat/fullnode-sync-only...consensus-filter

I propose we decide on the high level approach and then we continue in the PR branch with that.
cc @romac @nenadmilosevic95

romac:
> Drafted two approaches for filtering:
> 
> * Engine-level approach
>   [ancazamfir/malachite@feat/fullnode-sync-only...engine-filter](https://github.com/ancazamfir/malachite/compare/feat/fullnode-sync-only...engine-filter)
> * Consensus-core approach
>   [ancazamfir/malachite@feat/fullnode-sync-only...consensus-filter](https://github.com/ancazamfir/malachite/compare/feat/fullnode-sync-only...consensus-filter)
> 
> I propose we decide on the high level approach and then we continue in the PR branch with that. cc @romac @nenadmilosevic95

I am leaning towards the `consensus-core approach`, since we already have similar checks there and also because that facilitates integration of Malachite by teams building their own engine around `consensus-core`.

nenadmilosevic95:
> Drafted two approaches for filtering:
> 
> Engine-level approach
> [ancazamfir/malachite@feat/fullnode-sync-only...engine-filter](https://github.com/ancazamfir/malachite/compare/feat/fullnode-sync-only...engine-filter)
> Consensus-core approach
> [ancazamfir/malachite@feat/fullnode-sync-only...consensus-filter](https://github.com/ancazamfir/malachite/compare/feat/fullnode-sync-only...consensus-filter)
> I propose we decide on the high level approach and then we continue in the PR branch with that. cc @romac @nenadmilosevic95

Agree here with @romac as we already make a distinction between a validator and non-validator at the core-level, we might want to keep it there. 

> We can imagine other ways to do it (e.g. use proofs, percentile median status heights etc.), but these need to be spec-ed imo. @nenadmilosevic95 do you want to drive/ work on the spec for this?

I fully agree here as now the full node vs validator distinction is not as simple as before where the validator was just not signing messages. I can start working on this but it would be nice to have a short call to scope this well.

ancazamfir:
re: the two approaches
For the record, I prefer the engine filter and remove also the checks from consensus-core. We have all checks in one place and we don't pollute consensus core with node setup details.

I know we had the argument before about the complexity for consensus API level integrators but for me it's becoming weak.

romac:
I guess it can make sense to remove those checks from core-consensus and move them to engine, which is basically saying that a non-validator node should never have to call into core-consensus. If we can guarantee the latter, then I'd be fine with that approach.

nenadmilosevic95:
> I guess it can make sense to remove those checks from core-consensus and move them to engine, which is basically saying that a non-validator node should never have to call into core-consensus. If we can guarantee the latter, then I'd be fine with that approach.

That would be nice but how would a non-validator node decide then?

ancazamfir:
> I guess it can make sense to remove those checks from core-consensus and move them to engine, which is basically saying that a non-validator node should never have to call into core-consensus. If we can guarantee the latter, then I'd be fine with that approach.

Could you clarify a bit? by non-validator do you mean a node that is not a validator or a validator that has consensus disabled?

For engine filtering:
On the send out (to network and apps) side we filter the Effects.
On the receive side, a node that doesn't want to see consensus messages will not subscribe to consensus topics and should never call into core-consensus with network messages. So we are left with `StartHeight` and `ProposeValue`. The former has to be sent to core (so we can sync), the latter no (but shouldn't occur since it's the response to `GetValue`). 
Is what you mean by " should never have to call into core-consensus...guarantee"?

ancazamfir:
> That would be nice but how would a non-validator node decide then?

We don't filter the HostMsg::Decided sent to the app. The same logic that applies at core will be moved to engine. Let me finish the engine filter impl in the branch so we see how it looks like.

nenadmilosevic95:
> We don't filter the HostMsg::Decided sent to the app. The same logic that applies at core will be moved to engine. Let me finish the engine filter impl in the branch so we see how it looks like.

I am not sure I understand, but let's see when you make all the changes it might become clear to me.

ancazamfir:
The branch was ok, I removed the validator check from the core (left just one place for the Role). Added two more filters for vote extension related effects.
@nenadmilosevic95 let's have a sync call.

The diffs https://github.com/ancazamfir/malachite/compare/feat/fullnode-sync-only...engine-filter

ancazamfir:
> * Consensus-core approach
>   [ancazamfir/malachite@feat/fullnode-sync-only...consensus-filter](https://github.com/ancazamfir/malachite/compare/feat/fullnode-sync-only...consensus-filter)

Let's go with this and we can revise later.

code/crates/core-consensus/src/state.rs
```diff
@@ -266,6 +266,15 @@ where
             .is_some()
     }
 
+    /// Check if this node is an active validator.
+    ///
+    /// Returns true only if:
+    /// - Consensus is enabled in the configuration, AND
+    /// - This node is present in the current validator set
+    pub fn is_active_validator(&self) -> bool {
+        self.params.enabled && self.is_validator()
```
romac: Do we need the distinction between `is_validator` and `is_active_validator` or could we merge them into one method?

romac:
One issue with the config customizer approach here is that for now it only works with the test app. Will see if I can find a way to generalize it so that it can be made to work for other apps as well.

code/crates/core-consensus/src/state.rs
ancazamfir: Good question. Probably not. The only place where we use this now is here:
https://github.com/ancazamfir/malachite/blob/8aedab3f3e430e6e957c430f1eab055f8070ef36/code/crates/core-consensus/src/handle/driver.rs#L60-L66 to setup the `role` which is sent to the app. Currently only used for logging. 
If we replace with `is_active_validator` the only difference will be that all log will show a round with no one being `Proposer`.

I will remove it if you are ok with it. Keep only `is_active_validator`

romac: > If we replace with is_active_validator the only difference will be that all log will show a round with no one being Proposer.

Not sure I follow? Given that sync-only nodes will never be proposer, then that wouldn't change the actual logs, no?

> I will remove it if you are ok with it. Keep only is_active_validator

Sounds good to me

ancazamfir: > Not sure I follow? Given that sync-only nodes will never be proposer, then that wouldn't change the actual logs, no?

But even in validator sync-only mode we still send `StartedRound` (including the role) to the app, that's where the app prints a log message.

> Sounds good to me
Ok, will make the change.

romac:
@ancazamfir I cannot merge this as the earlier commits are lacking signatures. Would you mind resigning them all with this command and force-pushing?

```
$ git rebase --exec 'git commit --amend --no-edit -n -S' -i main'
```

romac (approved):

PR #1177, opened 2025-08-20
chore(CCHAIN275): Proposal keeper panics with batch/parallel sync and dynamic validator set

ancazamfir:
Closes: #XXX

With sync ahead and batching we panic if validator set is dynamic. The problem is that we determine the proposer of height `H` (the height of the value in the response) based on the validator set for height `h` (the current height when the response is received). We store this proposer and when we reach `H` we verify that the proposer recorded is what we expect for `H`.
https://github.com/circlefin/malachite/blob/cefb3d752a0cc57dc2b45b6c165e2330c2c35187/code/crates/core-driver/src/proposal_keeper.rs#L79-L90

The change proposed here is to buffer all responses for higher heights. When the height is reached we process the response in the handler `on_value_response()` 

There are two flows resulting in `on_value_response()` being called:

  **Response coming from the network**
1.   `NetworkEvent::SyncResponse` (consensus.rs:484-507)
2.   `process_sync_response()` (consensus.rs:502) 
3.   `process_input()` with `SyncValueResponse` (consensus.rs:662-670)
4.   `malachitebft_core_consensus::process!()` (consensus.rs:314)
5.   `handle_input()` (handle.rs:40-68)
6.   `Input::SyncValueResponse` match (handle.rs:60)
7.   `on_value_response()` (sync.rs:6) 
8.   `if response height is > current height` we buffer, otherwise we process

**Later, when height advances on StartHeight Message**
1. `reset_and_start_height()` (start_height.rs:6)
2. `replay_pending_msgs()` (start_height.rs:64) 
3. `shift_and_take()` from Input Queue (start_height.rs:81)
4. `handle_input()` for each pending input (start_height.rs:91)
5. `Input::SyncValueResponse` match (handle.rs:60)
6. `on_value_response()` (sync.rs:6)
→ `if consensus_height < cert_height` re-buffer like above, else
→ process certificate if invalid send `Effect::InvalidValue` to the engine (engine will notify Sync)
→ if valid send `Effect::SyncValue` to the engine (engine will notify the application)

The proposer is set once in the engine on `Effect::SyncValue`, when the response height matches the current height and the validator set is known.

---

#### For all contributors


code/crates/core-consensus/src/effect.rs
```diff
@@ -134,7 +134,12 @@ where
     /// Notifies the application that consensus has received a value response.
     ///
     /// Resume with: [`resume::Continue`]
-    SyncValue(ValueResponse<Ctx>, resume::Continue),
+    SyncValue(ValueResponse<Ctx>, Ctx::Address, resume::Continue),
+
+    /// Notifies the engine that consensus has received an invalid value.
+    ///
+    /// Resume with: [`resume::Continue`]
+    InvalidValue(PeerId, Ctx::Height, Error<Ctx>, resume::Continue),
```
romac:
```suggestion
/// Notifies the engine that consensus has received a valid sync value.
   ///
   /// Resume with: [`resume::Continue`]
    ValidSyncValue(ValueResponse<Ctx>, Ctx::Address, resume::Continue),

    /// Notifies the engine that consensus has received an invalid sync value.
    ///
    /// Resume with: [`resume::Continue`]
    InvalidSyncValue(PeerId, Ctx::Height, Error<Ctx>, resume::Continue),
```

code/crates/engine/src/consensus.rs
```diff
@@ -1248,7 +1220,36 @@ where
...
+                    }
+                }
+
+                Ok(r.resume_with(()))
+            }
+
+            Effect::SyncValue(value, proposer, r) => {
+                // NOTE: The state.height is not yet updated if this is an effect that is triggered by the
+                // Msg::StartHeight(height), with buffered sync value for height `height`. So this will panic.
+                // assert_eq!(value.certificate.height, state.height);
```
romac: Do you find it useful to leave the commented out assertion here or should we remove it?

code/crates/core-consensus/src/handle/sync.rs
```diff
@@ -12,28 +12,40 @@ pub async fn on_value_response<Ctx>(
...
+    info!(
+        value.certificate.height = %cert_height,
         signatures = value.certificate.commit_signatures.len(),
         "Processing value response"
     );
 
-    if state.driver.height() < value.certificate.height {
-        debug!("Received value response for higher height, queuing for later");
+    let proposer = state
+        .get_proposer(cert_height, value.certificate.round)
```
melekes, line 42: curious if we perform any additional checks on `cert_height`. like if a peer sends a certificate with `height=0`, what will happen?

code/crates/engine/src/consensus.rs
ancazamfir: no, i meant to remove after discussing with you. It was pretty difficult to debug the cause of assert failure, not sure if there is anything we can do about it (have the `state.height` updated somehow), couldn't figure out.

code/crates/core-consensus/src/handle/sync.rs
ancazamfir: oh, great catch @melekes...lost the check with my changes. Let me look again, I think there's another one I missed.

ancazamfir: main is also broken, we lost the checks with the batching feature.

ancazamfir: I added the check in 70ba27b82f837a1189ff960615fff6746f3e71a9. Without it the sync value is fully processed until it's dropped at consensus (if the app doesn't do it before).
The other check I thought I missed is [there](https://github.com/ancazamfir/malachite/blob/70ba27b82f837a1189ff960615fff6746f3e71a9/code/crates/engine/src/consensus.rs#L1272).
It checks that the commit certificate and value IDs match.

code/crates/engine/src/consensus.rs
ancazamfir: 64613ad573463696663cf18d00c3413df85fb06c

melekes (approved):
👍

PR #1178, opened 2025-08-21
chore(CCHAIN309): Check all scenarios when the driver panics

nenadmilosevic95:
Closes: #XXX

Triggered by the panic observed [here](https://github.com/circlefin/malachite/pull/1177), we decided to take a closer look to see when the driver can panic and whether it can happen as part of regular execution or only due to bugs in our code. 

There are 3 scenarios when driver can panic: 
1. in `./src/proposal_keeper.rs` when it tries to store two proposals from two different proposers. This happened in the case linked above, and we need to make sure that a proposal received from a non-proposer is dropped before it reaches the driver so this panic can never happen. 
```rust 
        // Check if the proposal is from the same validator as the first one recorded.
        // This should never happen, as proposals should always come from the same validator (round's proposer).
        if let Some(first) = self.get_first_proposal() {
            if first.validator_address() != proposal.validator_address() {
                panic!(
                    "BUG: Received proposals from different validators in the same round.\n\
                    Existing: {:?}, New: {:?}",
                    first.validator_address(),
                    proposal.validator_address()
                );
            }
        }
```
        

2. in `./src/driver.rs` we have two panics when dealing with round certificates. Namely, when the vote keeper generates `Skip` or `PrecommitAny` certificate event based on the votes it received, the driver tries to store these certificates in its `round_certificate` field. Before storing driver checks if votes for these certificates exist in the vote keeper (actually it checks if vote keeper contains votes from the same round as certificate), and if not it panics. This should never happen because the same vote keeper generated these certificate events in the first place because it has these votes.

```rust

fn store_precommit_any_round_certificate(&mut self, vote_round: Round) {
        let Some(per_round) = self.vote_keeper.per_round(vote_round) else {
            panic!("Missing the PrecommitAny votes for round {vote_round}");
        };

fn store_skip_round_certificate(&mut self, vote_round: Round) {
        let Some(per_round) = self.vote_keeper.per_round(vote_round) else {
            panic!("Missing the SkipRoundvotes for round {vote_round}");
        };

```

---

#### For all contributors

### DIFF
--- a/code/crates/core-driver/src/driver.rs
+++ b/code/crates/core-driver/src/driver.rs
@@ -475,6 +475,8 @@ where
 
         match &output {
             VKOutput::PolkaValue(val) => self.store_polka_certificate(vote_round, val),
+            // TODO: Can we have a scenario where this event is generated for a lower round
+            // and we set our round certificate for a certificate from a lower round?
             VKOutput::PrecommitAny => self.store_precommit_any_round_certificate(vote_round),
             VKOutput::SkipRound(round) => self.store_skip_round_certificate(*round),
             _ => (),
--- a/code/crates/test/app/src/app.rs
+++ b/code/crates/test/app/src/app.rs
@@ -3,7 +3,7 @@ use std::time::Duration;
 use eyre::eyre;
 use malachitebft_app_channel::app::engine::host::Next;
 use tokio::time::sleep;
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, warn};
 
 // use malachitebft_app_channel::app::config::ValuePayload;
 use malachitebft_app_channel::app::streaming::StreamContent;
@@ -70,9 +70,26 @@ pub async fn run(
                 let pending = state.store.get_pending_proposals(height, round).await?;
                 info!(%height, %round, "Found {} pending proposals, validating...", pending.len());
                 for p in &pending {
-                    // TODO: check proposal validity
-                    state.store.store_undecided_proposal(p.clone()).await?;
-                    state.store.remove_pending_proposal(p.clone()).await?;
+                    // Check if proposer matches current proposer
+                    if p.proposer == proposer {
+                        state.store.store_undecided_proposal(p.clone()).await?;
+                        state.store.remove_pending_proposal(p.clone()).await?;
+                        info!(
+                            height = %p.height,
+                            round = %p.round,
+                            proposer = %p.proposer,
+                            "Moved valid pending proposal to undecided"
+                        );
+                    } else {
+                        state.store.remove_pending_proposal(p.clone()).await?;
+                        warn!(
+                            height = %p.height,
+                            round = %p.round,
+                            actual_proposer = %p.proposer,
+                            expected_proposer = %proposer,
+                            "Removed invalid pending proposal - wrong proposer"
+                        );
+                    }
                 }
 
                 // If we have already built or seen values for this height and round,
@@ -254,9 +271,9 @@ pub async fn run(
             // It may happen that our node is lagging behind its peers. In that case,
             // a synchronization mechanism will automatically kick to try and catch up to
             // our peers. When that happens, some of these peers will send us decided values
-            // for the heights in between the one we are currently at (included) and the one
-            // that they are at. When the engine receives such a value, it will forward to the application
-            // to decode it from its wire format and send back the decoded value to consensus.
+            // for the current height only (not for future heights). When the engine receives
+            // such a value, it will forward to the application to decode it from its wire format
+            // and send back the decoded value to consensus.
             AppMsg::ProcessSyncedValue {
                 height,
                 round,
@@ -276,6 +293,7 @@ pub async fn run(
                         validity: Validity::Valid,
                     };
 
+                    // TODO: Should we do some validation at the app side for these synced values?
                     state.store_synced_value(proposal.clone()).await?;
 
                     if reply.send(Some(proposal)).is_err() {
--- a/code/crates/test/app/src/state.rs
+++ b/code/crates/test/app/src/state.rs
@@ -117,6 +117,9 @@ impl State {
         }
 
         // Verify the proposal signature
+        // TODO: Here we assume that we can check the signature of the proposal always because the validator set is fixed
+        // In a real application, we would need to delay this verification until we know the actual validator set for this height
+
         match self.verify_proposal_signature(&parts) {
             Ok(()) => {
                 // Signature verified successfully, continue processing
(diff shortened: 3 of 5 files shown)


code/crates/core-driver/src/driver.rs
```diff
@@ -475,6 +475,8 @@ where
 
         match &output {
             VKOutput::PolkaValue(val) => self.store_polka_certificate(vote_round, val),
+            // TODO: Can we have a scenario where this event is generated for a lower round
+            // and we set our round certificate for a certificate from a lower round?
```
ancazamfir: Yes, I think so. Do you see a problem with that?

code/crates/test/app/src/app.rs
```diff
@@ -70,9 +70,26 @@ pub async fn run(
...
+                        state.store.store_undecided_proposal(p.clone()).await?;
+                        state.store.remove_pending_proposal(p.clone()).await?;
+                        info!(
+                            height = %p.height,
+                            round = %p.round,
+                            proposer = %p.proposer,
+                            "Moved valid pending proposal to undecided"
+                        );
+                    } else {
+                        state.store.remove_pending_proposal(p.clone()).await?;
```
ancazamfir: we can move this outside `if` since we do it in both cases

code/crates/test/app/src/app.rs
```diff
@@ -276,6 +293,7 @@ pub async fn run(
                         validity: Validity::Valid,
                     };
 
+                    // TODO: Should we do some validation at the app side for these synced values?
```
ancazamfir: Yes. We don't do much in our test apps though. We should add a comment at least.

code/crates/test/app/src/state.rs
```diff
@@ -117,6 +117,9 @@ impl State {
         }
 
         // Verify the proposal signature
+        // TODO: Here we assume that we can check the signature of the proposal always because the validator set is fixed
+        // In a real application, we would need to delay this verification until we know the actual validator set for this height
+
```
ancazamfir: Here we should store in pending if the value is for higher height. And not do any validation at this point.

code/crates/core-driver/src/driver.rs
nenadmilosevic95: See [ac06e621](https://github.com/circlefin/malachite/pull/1178/commits/ac06e6211406ead7c231d41142179542c6078097)

code/examples/channel/src/app.rs
```diff
@@ -57,28 +57,43 @@ pub async fn run(state: &mut State, channels: &mut Channels<TestContext>) -> eyr
...
-                            "Removed invalid pending proposal - wrong proposer"
-                        );
+                let pending_parts = state
+                    .store
+                    .get_pending_proposal_parts(height, round)
+                    .await?;
+                info!(%height, %round, "Found {} pending proposal parts, validating...", pending_parts.len());
+
+                for parts in &pending_parts {
+                    match state.validate_proposal(parts) {
```
romac:
```suggestion
                    match state.validate_proposal_parts(parts) {
```

code/examples/channel/src/state.rs
```diff
@@ -112,6 +120,75 @@ impl State {
             .unwrap_or_default()
     }
 
+    /// Validates a proposal by checking both proposer and signature
+    pub fn validate_proposal(&self, parts: &ProposalParts) -> Result<(), ProposalValidationError> {
```
romac:
```suggestion
    pub fn validate_proposal_parts(&self, parts: &ProposalParts) -> Result<(), ProposalValidationError> {
```

code/examples/channel/src/state.rs
```diff
@@ -112,6 +120,75 @@ impl State {
...
+
+        // If proposer is correct, verify the signature
+        self.verify_proposal_signature(parts)
+            .map_err(ProposalValidationError::Signature)?;
+
+        Ok(())
+    }
+
+    /// Verify proposal signature
+    fn verify_proposal_signature(
```
romac:
```suggestion
    fn verify_proposal_parts_signature(
```

romac (commented):
Very nice! Left a couple nitpicks for some methods names

code/crates/test/app/src/app.rs
```diff
@@ -18,7 +18,7 @@ use malachitebft_test::{Genesis, Height, TestContext};
 use crate::state::{decode_value, State};
 
 pub async fn run(
-    genesis: Genesis,
+    _genesis: Genesis,
```
romac: Let's remove this argument if it is not used and can be accessed from the state.

code/crates/test/app/src/state.rs
```diff
@@ -393,75 +500,31 @@ impl State {
...
+            .filter_map(|part| part.as_data())
+            .fold(1, |acc, data| acc * data.factor);
+
+        Ok(ProposedValue {
+            height: parts.height,
+            round: parts.round,
+            valid_round: init.pol_round,
+            proposer: parts.proposer,
+            value: Value::new(value),
+            validity: Validity::Valid, // TODO: Check signature in Fin part
```
romac: I believe this TODO is outdated since the proposal has already been validated.

code/crates/test/app/src/state.rs
```diff
@@ -393,75 +500,31 @@ impl State {
...
-            .signing_provider
-            .verify(&hash, &fin.signature, &proposer.public_key)
-        {
-            return Err(SignatureVerificationError::InvalidSignature);
-        }
-
-        Ok(())
+    /// Re-assemble a [`ProposedValue`] from its [`ProposalParts`].
+    ///
+    /// This is done by multiplying all the factors in the parts.
```
romac, line 508:
```suggestion
    /// This is done by multiplying all the factors in the parts.
    ///
    /// ## Important
    /// This method assumes that the proposal parts have already been validated by `validate_proposal_parts`
```

code/crates/test/app/src/state.rs
```diff
@@ -88,6 +131,82 @@ impl State {
...
+            hasher.update(init.round.as_i64().to_be_bytes());
+
+            // The correctness of the hash computation relies on the parts being ordered by sequence
+            // number, which is guaranteed by the `PartStreamsMap`.
+            for part in parts.parts.iter().filter_map(|part| part.as_data()) {
+                hasher.update(part.factor.to_be_bytes());
+            }
+
+            hasher.finalize()
+        };
```
ancazamfir, line 192: We should verify the proposal based on the hash sent by proposer.
In validate_proposal, if the proposal was signed by the proposer for the height and round, we should check our hash (computing it like above) against the proposer's hash and return something that translates in Validity::Invalid.

code/crates/test/app/src/app.rs
nenadmilosevic95: Done, see [ffd89a](https://github.com/circlefin/malachite/pull/1178/commits/ffd89a70c5ac28bc5de35106746a097b785d0f53)

code/crates/test/app/src/state.rs
nenadmilosevic95: Done, check new code

code/crates/test/app/src/state.rs
nenadmilosevic95: After live sync with @ancazamfir, @romac and @sergio-mena  we have decided to tackle this later.

PR #1184, opened 2025-08-28
feat(code): limit the number of values in a sync response based on upper limit

insumity:
Closes: #1164 

This PR is the **extension** of #1171 that I do not have access to update anymore. Look at all the commits after [_Re-request range from any peer, not necessarily the same one_](https://github.com/circlefin/malachite/pull/1171/commits/7160ec404a158b20f412cf02ed14fd73460912fd) to see the changes.

**Testing**
To see that indeed `max_response_size` is working, we ran the `response_size_limit_exceeded` test against the `main` branch with the following params:
```
    test.build()
        .run_with_params(
            Duration::from_secs(60),
            TestParams {
                enable_value_sync: true,
                // Values are around ~900 bytes, so we canNOT have 2 values in one batch.
                rpc_max_size: ByteSize::b(1000),
                batch_size: 2,
                parallel_requests: 1,
                ..Default::default()
            },
        )
        .await
```
in which case the test fails because it tries to send 2 values at once something that cannot be done because `rpc_max_size` is set to `1000` bytes.

---

#### For all contributors

### DIFF
--- a/code/crates/engine/src/sync.rs
+++ b/code/crates/engine/src/sync.rs
@@ -516,7 +516,7 @@ where
 impl<Ctx, Codec> Actor for Sync<Ctx, Codec>
 where
     Ctx: Context,
-    Codec: SyncCodec<Ctx> + Send + std::marker::Sync + 'static,
+    Codec: SyncCodec<Ctx>,
 {
     type Msg = Msg<Ctx>;
     type State = State<Ctx>;


code/crates/engine/src/sync.rs
```diff
@@ -296,6 +298,33 @@ where
                         .success_or(eyre!("Failed to get decided value for height {height}"))?;
 
                     if let Some(value) = value {
+                        let value_response = ValueResponse(sync::ValueResponse::new(
+                            *range.start(),
+                            vec![value.clone()],
+                        ));
+
+                        let result = ractor::call!(self.gossip, move |reply_to| {
+                            NetworkMsg::GetResponseSize(value_response.clone(), reply_to)
```
jmalicevic: This implies another round of of communication. How expensive is this?  Actually what might be really expensive is that `GetResponseSize` encodes the value again just to get its size. Can we not change the response type to also include it's size as a variable?

code/crates/engine/src/sync.rs
```diff
@@ -296,6 +298,33 @@ where
...
+                            Ok(value_in_bytes) => value_in_bytes,
+                            Err(e) => {
+                                error!("Failed to get response size for value, stopping at for height {}: {:?}", height, e);
+                                break;
+                            }
+                        };
+
+                        // check if adding this value would exceed the max-response limit
+                        if response_size_bytes + total_value_size_bytes
+                            > self.sync_config.max_response_size
```
jmalicevic, line 341: Deleted my previous suggestion to change the check because I realize that it would break early. But if it is equal then we should push the value and break, not go another round and only then break. So a separate 

So you add a check below line 328:

```
if response_size_bytes + total_value_size_bytes == self.sync_config.max_response_size {

break;

 }
```

code/crates/sync/src/handle.rs
```diff
@@ -306,14 +306,44 @@ where
...
+            // values are fully processed.
+            // To tackle this, we first update the current pending request with the range of values
+            // it provides we received, and then issue a new request with the remaining values.
+            let new_start = requested_range
+                .start()
+                .increment_by(response.values.len() as u64);
+
+            let end = *requested_range.end();
+
+            if response.values.is_empty() {
```
jmalicevic, line 341: Should this check not be performed outside the IF on line 320 as well?

code/crates/sync/src/handle.rs
```diff
@@ -306,14 +306,44 @@ where
...
+                error!(%request_id, %peer_id, "Received response contains no values");
+            } else {
+                // The response of this request only provided `response.values.len()` values,
+                // so update the pending request accordingly
+                let updated_range = *requested_range.start()..=new_start.decrement().unwrap();
+                state.update_request(request_id, peer_id, updated_range);
```
jmalicevic, line 347: Here you are updating (overwriting the requestID range with a new range).  

Say the old request looked like this: `request_id: 1, range [1-10]`. 
You received [1-5]. 

After this function this will look : `request_id:1, range [6-10]` . 

Now consensus receives the same range back. It starts processing values and finds an error in value for height 2. It notifies the sync reactor to re-request it.  The sync reactor will try to [get](code/crates/sync/src/state.rs) the `request_id`  [iterating](https://github.com/insumity/malachite/blob/6f23c04693358734ff3e457f507d928dbccceab3/code/crates/sync/src/state.rs#L139) through the pending_requests map and not find it.

insumity: Not sure I see your argument but maybe I'm missing something.
`update_request` updates the range of the request with the given `request_id`.
The range we use to update `request_id:1` is computed as follows:
```
let updated_range = *requested_range.start()..=new_start.decrement().unwrap();
```
So, the `updated_range` is `[*requested_range.start(), new_start - 1]`. `requested_range.start()` corresponds to `1` because the original requested range was `[1, 10]`. Then `new_start` corresponds to:
```
 let new_start = requested_range
      .start()
      .increment_by(response.values.len() as u64);
```
so, `new_start` is `1 + 5 = 6` because in your example we received `[1, 5]` so `5 = response.values.len()` values. Therefore the `updated_range` ends up being `[1, 6 - 1] = [1, 5]`.

> After this function this will look : request_id:1, range [6-10] .

Based on the above, after the `update_request` call, the range is not `[6,10]` but is `[1, 5]`.

>The sync reactor will try to get the request_id iterating through the pending_requests map and not find it.

Therefore this should not be the case.

code/crates/sync/src/handle.rs
insumity: `range_len` is always positive here because it's computed as `end - start + 1` so even if the range was just one element, then `range_len` is `1`. Also, note that we never request an empty range of values; see the `range.is_empty` check in `request_values_from_peer`.
In other words if `response.values.is_empty`,  the condition `if responses.values.len() < range_len as usize` is true.

code/crates/engine/src/sync.rs
insumity: > This implies another round of of communication. How expensive is this?

Not clear. I don't expect this to be expensive but we'll need to do benchmarking to be certain.

>  Actually what might be really expensive is that GetResponseSize encodes the value again just to get its size. Can we not change the response type to also include it's size as a variable?

If the expensive part is the encoding, then I do not see a clear way on how we can only do the encoding once without substantial refactoring. This is because encoding takes place at the network actor before the value is sent over the network. So, we would need a way to check at this network-actor stage that we do not exceed the `max_response_size` if we wanted to only do the encoding once. If we see the current approach causes performance issues later on, we can revisit this idea.

Even if the host was to return the response size through `GetDecidedValue`, it would still need to contact the network/gossip actor to get the response size, so we would need to encode as well. Returning the response size as well, might be an option and it does eliminate one actor call/message, but again, I do not expect an actor message to be expensive.

code/crates/engine/src/sync.rs
insumity: Seems like a quite unlikely scenario. In any case, I added an extra check.

code/crates/sync/src/handle.rs
jmalicevic: Apologies for the noise, I missed that. You are right, all good.

code/crates/engine/src/sync.rs
jmalicevic: Agreed this is a potential improvement if things are not performing too well so not urgent to do as part of this PR.

code/crates/sync/src/handle.rs
jmalicevic: Thanks for clarifying. If `range_len` is never 0 then leaving it like this is ok.

code/crates/core-consensus/src/util/bounded_queue.rs
```diff
@@ -46,6 +47,8 @@ where
             if &index < max_index {
                 let max_index = max_index.clone();
 
+                warn!("Bounded queue is full, dropping value");
```
hvanz, line 50: This message is very useful to spot when the queue overflows. Thanks for adding it!

code/crates/engine/src/consensus.rs
```diff
@@ -663,6 +675,7 @@ where
             .await
         {
             error!(%height, "Error when processing received synced block: {e}");
+            return Err(e.into());
```
hvanz: Just a suggestion: replace the content of this function with the following (Github UI does not allow me to select the lines to do the suggestion). It will skip one level of error message, but I think that's fine.
```
        self.process_input(
            myself,
            state,
            ConsensusInput::SyncValueResponse(CoreValueResponse::new(
                peer,
                value.value_bytes.clone(),
                value.certificate.clone(),
            )),
        )
        .await
        .map_err(|e| e.into())
```

code/crates/sync/src/handle.rs
```diff
@@ -515,6 +545,33 @@ where
...
+    state: &mut State<Ctx>,
+    metrics: &Metrics,
+    range: RangeInclusive<Ctx::Height>,
+) -> Result<(), Error<Ctx>>
+where
+    Ctx: Context,
+{
+    // NOTE: We do not perform a `max_parallel_requests` check here in contrast to what is done, for
+    // example in `request_values`. This is because `request_values_range` is only called for retrieving
+    // partial responses, which means the original request is not on the wire anymore.
```
hvanz: I understand the original request is not in the wire (that is, its response was received) but it's still in `state.pending_requests`. At least for the received range that is being processed. Is this correct? Then, after the new request issued here, there will be `max_parallel_requests + 1` active requests in `state.pending_requests`. I'm not sure if this is a problem, but two requests will need to be removed from `state.pending_requests` before a new one can be issued.

jmalicevic: This is a good point, although I think in practice the behavior stays the same. 
For example you have `state.pending_requests = {req1[1-10],req2[11-20]}` . You split 1-10 because the range is too big and get : `state.pending_requests = {req1[1-3],req2[11-20], req3[4-10]}` 
The problem is that when `req2` is done, you cannot issue another request because you are not done processing `req1` and `req3` . If we compare this to how it was before batching , we are making things slightly worse by not allowing `req4` until we process everything from the original `req1`. 

However, in practice this delay is probably not very big compared to the initial use case. In any case, I would not block this PR on this optimization so that people can integrate with the API changes and we can then optimize this under the hood. 

WDYT @hvanz @insumity ?

insumity: Thanks for this @jmalicevic. Yes, what you describe is my view as well.

insumity:
Thanks @hvanz for your comments!
> As noted by @jmalicevic there will a performance hit for the extra encoding. Have you seen any performance decrease in your tests? Maybe a small benchmark test would be useful. As a solution, the encoded value could be cached somewhere but this will take some refactoring (for another PR). I'm not worried about the extra communication (GetResponseSize) between actors; this is just a couple of function calls.

I did not see any performance decrease in the tests but that's because the `block_size` is pretty small (i.e., ~900 bytes per value). I have a [PR on the `informalsystems` repo](https://github.com/informalsystems/malachite/pull/6) where I can tune `block_size` in tests and noticed the following rough numbers:
- for a single value of 1 MB of random data, the `ProtobufCodec` encoding takes around ~50 µs;
- for a value of 10 MB, encoding takes ~700 µs;
- for a value of 25 MB, encoding takes ~1.5 ms;
- for a value of 50 MB, encoding takes ~3 ms.

In any case, I changed the code so we can retrieve the encoded length without actually having to encode the data. So, now,  computing the encoded length takes a few tens of microseconds irrespectively of the value size. Note that this only holds for the `ProtobufCodec` and not for the `JsonCodec`.
_**Update:**_ I performed some slightly more [serious benchmarking](https://gist.github.com/insumity/32977ab6b86b668b8ec0db725fed04c8) and the results can be found [here](https://gist.github.com/insumity/95c1b2a0787011a92b56be0bdef18984). To encode and then get the length of a 100MB value we need ~5.5 ms but to get immediately the length of the encoded value using `encoded_len` we just need 85 ns. Also verified that `encoded(sth).len() == encoded_len(sth)`.
Let me know if you want me to add the benchmark code in this PR. 


Please do let me know what you think!

code/crates/sync/src/handle.rs
hvanz: Yes, I think we can live that. I'm thinking the extreme scenario where blocks are so big that they fill a whole response message. In the example, `req1[1-10]` would be split in 10 requests. I think this is not something that we should try to fix here. That would either be a problem in the configuration, or the whole setup is not well thought out. Maybe we could add a warning when `state.pending_requests.len() > state.max_parallel_requests`, to see in the logs when this case is happening very often.

hvanz:
Thanks for the benchmarks! They look pretty good. 10Mb is already quite big for a value and would probably reach the typical size limit of a response, so I think we don't need to worry for 700 µs.

code/crates/sync/src/handle.rs
insumity: > Maybe we could add a warning when state.pending_requests.len() > state.max_parallel_requests, to see in the logs when this case is happening very often.

Added some logging here based on your comment.

insumity:
> Thanks for the benchmarks! They look pretty good. 10Mb is already quite big for a value and would probably reach the typical size limit of a response, so I think we don't need to worry for 700 µs.

@hvanz So, we're okay with encoding the data and then getting the length as was initial done? If that's the case, I can remove this [commit](https://github.com/circlefin/malachite/pull/1184/commits/cb40f73ef7b115793ac6099c46c6f9157086e8c9).

hvanz:
> So, we're okay with encoding the data and then getting the length as was initial done? If that's the case, I can remove this [commit](https://github.com/circlefin/malachite/pull/1184/commits/cb40f73ef7b115793ac6099c46c6f9157086e8c9).

@romac WDYT? Do we need this optimization? Encoding a value of 1Mb takes about 50 µs and for 10 MB it takes ~700 µs.

jmalicevic:
@romac @hvanz , are we ok to merge this? Should we unblock in our fork if you don't have capacity?

jmalicevic:
@romac @hvanz  we have merged this in our fork and also added a follow up PR fixing a few other issues we found in the sync : https://github.com/informalsystems/malachite/pull/15 

The PR is long, happy to upstream it if you deem its useful.

romac:
> @romac @hvanz , are we ok to merge this? Should we unblock in our fork if you don't have capacity?

Apologies, I completely missed your question…

> @romac @hvanz we have merged this in our fork and also added a follow up PR fixing a few other issues we found in the sync: https://github.com/informalsystems/malachite/pull/15

Glad to hear you took the initiative of merging it in your fork and my apologies again for not getting to this sooner.

Unfortunately, we will likely not be able to merge this before a week or so.

> The PR is long, happy to upstream it if you deem its useful.

That would be awesome if it's not too much to ask, otherwise I am happy to upstream it myself from your fork. We can also wait until this PR is merged though, to avoid stacking them.

jmalicevic:
No worries, I understand :) Yes lets merge this PR and then we will upstream the other one and happy to do a sync review if you find it useful as its quite big.

romac:
Closing in favor of rebased and re-signed version in https://github.com/circlefin/malachite/pull/1251

PR #1187, opened 2025-09-01
fix(code/wal): Improve error handling in WAL

romac:
Closes: #XXX

---

Improve error handling in WAL, making sure that all errors are properly logged.

Also fixes a bug where a reply would not be sent to the consensus actor when the WAL actor receives a mismatched `Append` message, logging `Failed to send Append command to WAL actor: Messaging failed because channel is closed`.

---

#### For all contributors

### DIFF
--- a/code/crates/engine/src/wal.rs
+++ b/code/crates/engine/src/wal.rs
@@ -4,6 +4,7 @@ use std::path::PathBuf;
 use eyre::eyre;
 use ractor::{async_trait, Actor, ActorProcessingErr, ActorRef, RpcReplyPort, SpawnErr};
 use tokio::sync::{mpsc, oneshot};
+use tracing::warn;
 use tracing::{debug, error, info};
 
 use malachitebft_core_types::{Context, Height};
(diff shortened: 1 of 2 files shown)


code/crates/engine/src/wal.rs
```diff
@@ -99,7 +100,7 @@ where
 
             Msg::Append(height, entry, reply_to) => {
                 if height != state.height {
-                    debug!("Ignoring append at height {} != {}", height, state.height);
+                    warn!(wal.height = %state.height, entry.height = %height, "Ignoring append, mismatched height");
```
melekes, line 103: curious under which scenario this could happen

melekes (approved):
👍

sergio-mena (approved):

code/crates/engine/src/wal.rs
cason: Indeed.

sergio-mena (approved):
LGTM

code/crates/engine/src/wal.rs
romac: Since #1185, that is not supposed to happen anymore, as consensus inputs are only processed when consensus reaches the corresponding height.

melekes: Should we `assert!` here now?

melekes (approved):
👍

code/crates/engine/src/wal.rs
romac: before adding assertions, i would first like to check if we see these logs at all on the testnet or not. i don't want to add assertions that we have not confirmed won't be hit (not that lookign at logs is a definite proof but still better nothing)

PR #1189, opened 2025-09-04
fix!: Remove `GetValidatorSet` effect

romac:
Closes: #XXX

⚠️ Stacked on top of #1188

## Summary

After #1188, the `GetValidator` effect is not necessary anymore since we will only ever verify signatures for messages at the current height. We can therefore remove it, and with it a potential footgun for integrators.

---

#### For all contributors


melekes (approved):
lgtm

code/crates/core-consensus/src/handle/validator_set.rs
```diff
@@ -1,20 +1,22 @@
-use std::borrow::Cow;
-
 use crate::prelude::*;
 
 pub async fn get_validator_set<'a, Ctx>(
```
ancazamfir: Not sure how useful this is anymore. All callers verify already `state.height() == <msg>.height()` (they log if not, before calling `get_validator_set()`. They may as well use `state.validator_set()` instead of calling this.

code/crates/core-consensus/src/handle/validator_set.rs
```diff
@@ -1,20 +1,22 @@
...
-) -> Result<Option<Cow<'a, Ctx::ValidatorSet>>, Error<Ctx>>
+) -> Result<Option<&'a Ctx::ValidatorSet>, Error<Ctx>>
 where
     Ctx: Context,
 {
-    if state.driver.height() == height {
-        return Ok(Some(Cow::Borrowed(state.driver.validator_set())));
-    }
+    if state.height() == height {
+        Ok(Some(state.driver.validator_set()))
```
ancazamfir:
```suggestion
        Ok(Some(state.validator_set()))
```

ancazamfir (dismissed):
Approved but I think we should do some more cleanup, see comments.
Also, there are many places where we do:
```
    let height = state.driver.height();
    let consensus_round = state.driver.round();
```
instead of 
```
    let height = state.height();
    let consensus_round = state.round();
```

nenadmilosevic95 (approved):
LGTM!

code/crates/core-consensus/src/handle/validator_set.rs
romac: Agreed and we can move the height check directly within `state.validator.set()`

code/crates/core-consensus/src/handle/validator_set.rs
romac: 4c2f3c8b

code/crates/core-consensus/src/handle/validator_set.rs
romac: f1d38065

ancazamfir (approved):

PR #1190, opened 2025-09-05
chore: Remove initial height and validator set from consensus params

romac:
Closes: #XXX

These parameters are a remnant from a very early time in Malachite's lifespan where consensus would always start at the initial height with a static validator set and move from height to height on its own.

These parameters have therefore not had any effect for a while, given that it is now up to the app to tell consensus which height to start at after receiving the `ConsensusReady` message. 

As such, they can be removed without any loss of functionality.

---

#### For all contributors

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -5,6 +5,7 @@
 ### `malachitebft-core-types`
 
 - Move `SigningProvider` and `SigningProviderExt` traits into new `malachitebft-signing` crate ([#1191](https://github.com/informalsystems/malachite/pull/1191))
+- Remove `initial_validator_set` and `initial_height` fields from `Params` struct ([#1190](https://github.com/circlefin/malachite/pull/1190))
 
 ### `malachitebft-signing`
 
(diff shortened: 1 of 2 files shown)


romac:
@jmalicevic Any concerns about removing these two parameters? 

EDIT: I updated the PR description with the motivation for this change.

code/crates/core-consensus/src/handle/start_height.rs
```diff
@@ -18,7 +18,9 @@ where
     perform!(co, Effect::ResetTimeouts(Default::default()));
 
     #[cfg(feature = "metrics")]
-    metrics.step_end(state.driver.step());
+    {
+        metrics.step_reset();
```
jmalicevic: I guess this has nothing to do with the actual change in this PR?

jmalicevic:
> @jmalicevic Any concerns about removing these two parameters?
> 
> EDIT: I updated the PR description with the motivation for this change.

As the current workflow is to set this anyways when using `StartHeight` I am fine with this particular change,

code/crates/core-consensus/src/handle/start_height.rs
romac: Reverted

PR #1198, opened 2025-09-08
fix(CCHAIN-407): Nodes fail to continue syncing after network restart

ancazamfir:
Closes: #XXX

Full nodes fail to resume syncing after validator network restarts. 
Similarly a validator node not included in persistent peers of the restarting set is affected for both consensus and sync. 

---

#### For all contributors


code/crates/test/framework/src/params.rs
```diff
@@ -16,6 +16,10 @@ pub struct TestParams {
     pub value_payload: ValuePayload,
     pub max_retain_blocks: usize,
     pub stable_block_times: bool,
+    pub enable_discovery: bool,
+    /// Node IDs that should not be added as persistent peers for other nodes
+    /// (simulates nodes that joined after initial network setup)
+    pub exclude_from_persistent_peers: Vec<u64>,
```
melekes, line 23: is this an analog of banning? kinda like dead peers?

ancazamfir: kind of. It's just for testnet setup so we can add a node to the tesnet without adding its address in the persistent peers of other nodes.

ancazamfir:
Socket leak issue was seen with this PR on devnet and testnet. Here is a summary on what I believe happens:

- The problem:
  Malachite nodes with `listen_addr = "/ipv4/0.0.0.0:27000"` experience massive socket accumulation (6000+ sockets) leading to resource exhaustion, while explicit IP addresses work fine.

- Address Advertisement Difference

  With `0.0.0.0` binding:
     - Node advertises: `[/ip4/127.0.0.1/tcp/27000, /ip4/172.20.0.12/tcp/27000]`
     - Multiple nodes advertise the SAME loopback address (`127.0.0.1`)

   With explicit IP binding (e.g., `172.19.0.20`):
     - Node advertises: `[/ip4/172.19.0.20/tcp/27000]`
     - Each node advertises UNIQUE addresses


- The peer identification bug (`bootstrap_nodes`)

   With `0.0.0.0`:
     - Node1 advertises `[172.20.0.11, 127.0.0.1]` → correctly updates Entry 0
     - Node2 advertises `[127.0.0.1, 172.20.0.12]` → incorrectly matches Entry 0 due to shared `127.0.0.1`
   (we currently filter on any address: https://github.com/circlefin/malachite/blob/bd6a7c81db5cfea34dacd23da756b85c8827c0a1/code/crates/discovery/src/handlers/identify.rs#L59-L62)

     - Result: Entry 0 gets Node2's data, Entry 1 remains unidentified
     - Duplicate entries for same logical peer cause flip-flopping behavior

   With explicit IPs - Correct Behavior:
     - Node1 advertises `[172.19.0.20]` → matches Entry 0 
     - Node2 advertises `[172.19.0.21]` → matches Entry 1
     - Result: No overlap, each peer correctly identified

- Duplicate bootstrap_nodes entries cause:
  - Flip-flopping: Entry 0 (identified) vs Entry 1 (unidentified)
  - Continuous dial attempts: "Successfully initiated dial None" every second
  - Connection churn: New connections faster than cleanup (`TIME_WAIT` 60sec timeout)
  - Socket accumulation: 6000+ sockets per node


The solution: Loopback Address Filtering
Fix: 
- Filter out loopback addresses (127.x.x.x, ::1) during peer identification to prevent cross-contamination.
- ~~In `reconnect_to_persistent_peers()` do not try to connect to un-identified peers (`peer_id == None`)~~


Root cause summary
`0.0.0.0` binding causes libp2p to advertise shared loopback addresses, creating address cross-contamination in the `bootstrap_nodes` matching logic.

ancazamfir:
CI failures in this PR happen due to https://github.com/circlefin/malachite/pull/1211

ancazamfir:
> ~~* In `reconnect_to_persistent_peers()` do not try to connect to un-identified peers (`peer_id == None`)~~

That was a mistake as we want a first dial to trigger identify. Reverted the commit.

ancazamfir:
Few more issues:
- [x] We never remove peers from discovered list (i.e. we never call `self.discovered_peers.remove(&peer_id) `) ending up with stale peers. This has impact on discovery. (fixed in b73b233ea2c32a56ef520fc9cd3aa51cb596dffc)
- [x] The changes break some local setups where we use loopback addresses like `[127.0.0.1:27000, 127.0.0.1:27001,..]`. (fixed in 25f7e1f3c27b6c227485a0feb060f95e388928c5)
- [x] Discovery test fails in CI (fixed in 5f971abb81ca238bf8c63f36b07af36500fb0d58)

ancazamfir:
All tests described in https://github.com/ancazamfir/malachite/blob/fullnode-reconnect/code/TESTNET_SCENARIOS.md were failing. Should now be fixed with b73b233ea2c32a56ef520fc9cd3aa51cb596dffc and 25f7e1f3c27b6c227485a0feb060f95e388928c5

cason:
I spent some time in this PR, using the documentation, dockers and networks.

In the `make testnet` setup, the full node is not able to reconnect to the validators whatsoever.

In the `make testnet-multi` setup, the full node is able to reconnect to all validators after restarting.

In the `make tesnet-nat` setup, , the full node is able to reconnect to all validators after restarting.

Is this the expected behaviour at the current state of this PR?

ancazamfir:
> I spent some time in this PR, using the documentation, dockers and networks.
> 
> In the `make testnet` setup, the full node is not able to reconnect to the validators whatsoever.
> 
> In the `make testnet-multi` setup, the full node is able to reconnect to all validators after restarting.
> 
> In the `make tesnet-nat` setup, , the full node is able to reconnect to all validators after restarting.
> 
> Is this the expected behaviour at the current state of this PR?

All are supposed to succeed. What's the last commit you see? There was a problem in the Makefile, not finding the config files, fixed today in 6acb82b4af4102bc79a34179fd7d7ef77e981f9d (did some renaming and forgot to update the Makefile). But you should have seen a make error. Could you update and try again? We can debug together if you still issues.

ancazamfir:
> CI failures in this PR happen due to #1211

We need this to be first merged. Also waiting on devnet test results.

code/Makefile
```diff
@@ -46,3 +48,194 @@ tests: ## Run all the tests.
...
+		--home ./deployments/volumes/malachite
+
+	@echo "Copying single-net configs..."
+	@cp ./single-net-configs/node0-config.toml ./deployments/volumes/malachite/0/config/config.toml
+	@cp ./single-net-configs/node1-config.toml ./deployments/volumes/malachite/1/config/config.toml
+	@cp ./single-net-configs/node2-config.toml ./deployments/volumes/malachite/2/config/config.toml
+	@cp ./single-net-configs/node3-config.toml ./deployments/volumes/malachite/3/config/config.toml
+
+	@echo "Detecting host IP address..."
+	@HOST_IP=$$(hostname -I 2>/dev/null | awk '{print $$1}' || ipconfig getifaddr en0 2>/dev/null || ipconfig getifaddr en1 2>/dev/null || echo "127.0.0.1"); \
```
cason: `hostname -I 2>/dev/null | awk '{print $$1}'` returns 0 even thought the first part returns false, as in my laptop. As a result, the IP is empty.

cason:
> In the `make testnet` setup, the full node is not able to reconnect to the validators whatsoever.

Update: this now works with the basic setup, where validators are off for few seconds. But by adding a `30s` sleep to the restart validators logic, the full node is not able to reconnect to the nodes with a message such as:

```
malachite_node3  | 2025-10-01T15:23:10.692099Z ERROR node{moniker=app-3}:Actor{id="0.0"}:network: Failed to dial peer None at [/ip4/192.168.0.83/tcp/27000] after 5 trials
```

cason:
> In the `make testnet-multi` setup, the full node is able to reconnect to all validators after restarting.

This happens because the connections attempts fail with a different error, a timeout. If the validators stay off longer, say 60s, this setup also does not work.

In other words, this solution works as long as the validators are back and accept connections within the 5 re-connection attempts the full node will use. This is somehow problematic...

code/Makefile
ancazamfir: fixed in 6b2b7ce89f0e76aa2a779886c7ea9d27668ca6b8. If you want to retry.

ancazamfir:
> > In the `make testnet-multi` setup, the full node is able to reconnect to all validators after restarting.
> 
> This happens because the connections attempts fail with a different error, a timeout. If the validators stay off longer, say 60s, this setup also does not work.
> 
> In other words, this solution works as long as the validators are back and accept connections within the 5 re-connection attempts the full node will use. This is somehow problematic...

Good catch @cason ! Yes there is a Fibonacci retry on connection failure, 1 sec step and retry stops after 5 steps resulting in abandoning the retries after 19 sec. 
I made a tentative change in 24e7188d9aaad9dfdf9a90788c32b6645c924e33, will see if there is a better way.

code/nat-configs/node0-config.toml
```diff
@@ -0,0 +1,64 @@
...
+[consensus]
+timeout_propose = "3s"
+timeout_propose_delta = "500ms"
+timeout_prevote = "1s"
+timeout_prevote_delta = "500ms"
+timeout_precommit = "1s"
+timeout_precommit_delta = "500ms"
+timeout_rebroadcast = "5s"
+value_payload = "proposal-and-parts"
+queue_capacity = 100
```
cason:
```suggestion
queue_capacity = 100
enabled = true
```

cason: Needed for all, after feat(CCHAIN-274): Allow nodes to run in sync-only mode without participating in consensus (#1174)

code/makefile-scripts/monitor-sockets.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+echo "Monitoring socket connections for socket leak detection..."
+echo "Press Ctrl+C to stop"
+echo
+
+# Function to count sockets for a container
+count_sockets() {
+    local container_name="$1"
+    local count=$(docker exec "$container_name" ss -t 2>/dev/null | grep :27000 | wc -l || echo "0")
```
cason: The container has no `ss` installed

code/crates/discovery/src/handlers/dial.rs
```diff
@@ -175,7 +198,46 @@ where
...
+            // We check by address since bootstrap nodes may not have peer_id
+            let already_attempted = listen_addrs.iter().any(|addr| {
+                self.controller
+                    .dial
+                    .is_done_on(&crate::controller::PeerData::Multiaddr(addr.clone()))
+            });
+
+            if already_attempted {
+                debug!(
+                    "Skipping bootstrap node dial - retry chain in progress: addrs={:?}",
```
cason: This message is pretty verbose, printed every second essentially.

code/crates/discovery/src/handlers/dial.rs
```diff
@@ -175,7 +198,46 @@ where
...
+                    continue;
+                }
+            }
+
+            // For bootstrap nodes, check if already attempted (done_on flag)
+            // This prevents overlapping Fibonacci retry sequences since done_on is only cleared
+            // after all retries are exhausted
+            // The Fibonacci retry sequence is started when a connection fails, see handle_failed_connection()
+            // We check by address since bootstrap nodes may not have peer_id
+            let already_attempted = listen_addrs.iter().any(|addr| {
```
cason, line 206: There is a very similar test as part of `should_dial` method.

cason: Which, by the way, is invoked below.

So essentially this part is more for tracing and understanding the behaviour.

code/crates/discovery/src/handlers/dial.rs
```diff
@@ -175,7 +198,46 @@ where
 
     pub fn dial_bootstrap_nodes(&mut self, swarm: &Swarm<C>) {
         for (peer_id, listen_addrs) in &self.bootstrap_nodes.clone() {
-            self.add_to_dial_queue(swarm, DialData::new(*peer_id, listen_addrs.clone()));
+            let dial_data = DialData::new(*peer_id, listen_addrs.clone());
+
+            // Skip if already connected
+            if let Some(id) = peer_id {
```
cason: Also this test is part of `should_dial`

code/crates/discovery/src/handlers/identify.rs
```diff
@@ -9,6 +9,84 @@ impl<C> Discovery<C>
...
+        );
+
+        // Skip if peer is already identified (avoid duplicate work)
+        if self
+            .bootstrap_nodes
+            .iter()
+            .any(|(existing_peer_id, _)| existing_peer_id == &Some(peer_id))
+        {
+            debug!(
+                "Peer {} already identified in bootstrap_nodes - skipping",
```
cason, line 50: I couldn't see this in tests, it would be interesting.

code/crates/discovery/src/handlers/identify.rs
```diff
@@ -9,6 +9,84 @@ impl<C> Discovery<C>
...
+    /// - Identify protocol: peer sends identity information including supported protocols
+    /// - Protocol check: only compatible peers reach `handle_new_peer()`
+    /// - Bootstrap matching: **This function** matches the peer against bootstrap nodes:
+    ///    - Check if peer's advertised addresses match any bootstrap node addresses
+    ///    - If match found: update `bootstrap_nodes[i].0 = Some(peer_id)`
+    ///
+    /// Called after connection is established but before peer is added to active_connections
+    fn update_bootstrap_node_peer_id(&mut self, peer_id: PeerId) {
+        debug!(
+            "Checking peer {} against {} bootstrap nodes",
```
cason, line 38: A bit verbose too.

cason (commented):
A new round of experiments with the improved solution and it works.

I left some comments, mainly:
- This version does not work from scratch, node configurations must be fixed
- Some log messages are maybe too verbose, we should consider simplifying their production

code/nat-configs/node0-config.toml
ancazamfir: try to update, we fixed the config default in main.

code/makefile-scripts/monitor-sockets.sh
ancazamfir: sorry I will remove it, this is an old script. use `check-socket-leaks-simple.sh` instead.

code/crates/discovery/src/handlers/dial.rs
ancazamfir: `should_dial` ORs with some other conds. Without this you see a fibonacci retry sequence started every second ending up with 6 dial attempts per peer every second.

code/crates/discovery/src/handlers/identify.rs
ancazamfir: You'll see it if you remove the `if already_attempted` check above.

code/crates/discovery/src/handlers/identify.rs
ancazamfir: I don't see this. Will check again.

ancazamfir:
> A new round of experiments with the improved solution and it works.
> 
> I left some comments, mainly:
> 
> * This version does not work from scratch, node configurations must be fixed

Should work if you update to latest

> * Some log messages are maybe too verbose, we should consider simplifying their production

Yes, definitely.

code/crates/discovery/src/handlers/dial.rs
ancazamfir: 06ac66e06a1be5ee417515874fd1a6a519896825

code/crates/discovery/src/handlers/dial.rs
ancazamfir: 06ac66e06a1be5ee417515874fd1a6a519896825

code/makefile-scripts/monitor-sockets.sh
ancazamfir: Renamed `check-socket-...` one to `monitor-sockets.sh`. You can also `make monitor-sockets`
fa4065af22e8cfbe7acaa1948d7269dd333bd82e

code/crates/test/tests/it/value_sync.rs
```diff
@@ -510,3 +510,234 @@ pub async fn full_node_sync_after_all_persistent_peer_restart() {
...
+        .wait_until(HEIGHT)
+        .crash()
+        .restart_after(Duration::from_secs(3))
+        .wait_until(HEIGHT + 5)
+        .success();
+
+    // Node 4: full node that that syncs and needs to reconnect after all validators have restarted
+    test.add_node()
+        .full_node()
+        .start_after(1, Duration::from_secs(3))
```
nenadmilosevic95, line 670: Any particular reason why up we wait 12s to start and here and below 3s?

code/crates/discovery/src/handlers/close.rs
```diff
@@ -95,6 +85,37 @@ where
...
+                );
+                bootstrap_node.0 = None; // Reset to None so it can be re-identified
+                self.controller
+                    .dial_clear_done_for_peer(peer_id, &bootstrap_node.1);
+                return;
+            }
+        }
+
+        // Handle non-bootstrap peers when discovery is disabled
+        if !self.is_enabled() {
```
nenadmilosevic95, line 116: Would we need to also clean in case we enable discovery?

code/TESTNET_SCENARIOS.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Usage
+
+### Commands:
+```bash
+make testnet          # Single network + host IP scenario
+make testnet-mixed    # Single network + host IP scenario + mixed binaries
+make testnet-multi    # Multi-network address mismatch
+make testnet-nat      # True NAT with gateway
+make testnet-mixed    # Mixed version testing
```
nenadmilosevic95: Why two times `make testnet-mixed`?

code/TESTNET_SCENARIOS.md
```diff
@@ -0,0 +1,250 @@
...
+```bash
+# Check socket connections for any running testnet
+./makefile-scripts/check-socket-leaks-simple.sh
+
+# Monitor socket connections for any running testnet
+./makefile-scripts/check-socket-leaks-simple.sh monitor
+
+# View logs for specific scenarios
+docker compose logs node3 --follow                             # Standard testnet
+docker compose -f docker-compose-multi-network.yml logs node3  # Multi-network
```
nenadmilosevic95: Is it supposed to work even without -f option, because here it somehow works?

code/makefile-scripts/setup-mixed-testnet.sh
```diff
@@ -0,0 +1,133 @@
...
+fi
+
+# Build latest version
+RUST_MIN_STACK=16777216 cross build --release --target x86_64-unknown-linux-gnu
+cp target/x86_64-unknown-linux-gnu/release/informalsystems-malachitebft-example-channel target/mixed-testnet/latest/malachite
+
+echo ""
+echo " Creating mixed-version docker compose.yml..."
+
+# Create the mixed-version docker compose file
```
nenadmilosevic95: Why do we need to do this dynamically in this case?

code/makefile-scripts/monitor-sockets.sh
```diff
@@ -0,0 +1,134 @@
+#!/bin/bash
+
+echo "Socket Analysis - $(date)"
+echo ""
+
+# Summary table
+echo "Socket counts per node (ports 27000-27003 only):"
+echo "Node     | Total | ESTABLISHED | LISTEN | TIME_WAIT | CLOSE_WAIT | Other"
```
nenadmilosevic95: It will be nice (if possible) to add expected values for each of the raws?

code/crates/network/test/tests/discovery.rs
```diff
@@ -94,6 +94,7 @@ pub async fn circular_graph_n() {
 // Testing correctness when discovery is disabled. Especially, the nodes should
 // not accept more connections than the defined number of inbound peers in the
 // configuration.
+// With discovery disabled all peers are considered inbound.
```
nenadmilosevic95, line 97: What was happening in CI?

code/crates/discovery/src/handlers/dial.rs
```diff
@@ -158,6 +158,29 @@ where
 
                 self.metrics.increment_total_failed_dials();
 
+                // For bootstrap nodes, clear the done_on flag so they can be retried
```
nenadmilosevic95, line 161: The alternative @ancazamfir and I discussed here is to allow just one retry mechanism for boostrap nodes. Namely, disable Fibonacci retry mechanism for bootstrap nodes, and only rely on retry mechanism triggered from the main loop. This is interesting as it mitigates having two retry mechanism enabled at the same time that you need to sync somehow.

ancazamfir:
> > CI failures in this PR happen due to #1211
> 
> We need this to be first merged. Also waiting on devnet test results.

CI clean after merge

code/TESTNET_SCENARIOS.md
```diff
@@ -0,0 +1,250 @@
+# Testnet Scenarios
```
nenadmilosevic95: We should update this file

nenadmilosevic95 (commented):
I did a full pass commit by commit.

Amazing work @ancazamfir, it was super hard for me to review this, I can imagine how hard it was to debug it! 

I left some comments.

When testing it, `make testnet` and `make testnet-multi` worked without any problems. I tested with different T for `restart-validators` up to 125s. Everything looks good! 

On the other side `make testnet-mixed` is not working on my laptop. And `make testnet-nat` is showing that `node3` is only connected to `node0` and not to all three. Also upon restart validators I am seeing some `TIME_WAIT` and `FIN_WAIT_2`.

code/TESTNET_SCENARIOS.md
cason: Because the name of the nodes are the same.

code/crates/discovery/src/handlers/dial.rs
cason: But in this case we would retry at fixed intervals, even for nodes that are "forever" down or down for a very long period of time. The Fibonacci approach allows us to slow it a bit down.

code/TESTNET_SCENARIOS.md
cason: Actually, this shouldn't be merged, this was a description of the tested env.

code/crates/test/tests/it/value_sync.rs
cason: The 12s is on the previous test, for context.

code/makefile-scripts/setup-mixed-testnet.sh
cason: You're right, we don't need that, as the file is already tracked.

code/TESTNET_SCENARIOS.md
cason: The -f parameter tells from where to read the containers definition.

cason:
I think this PR contains two parts:

1. The solution of a problem
2. Testnets, scripts, makefile allowing observing the solution in place to solve a problem (a long restart of a node)

I think only (1) should be merge to main/master, but I feel bad dropping (2), which can be useful again in the future.

How can we then proceed?

code/makefile-scripts/setup-mixed-testnet.sh
ancazamfir: yes, forgot to update the script. This setup was meant to simulate and test the mixed binary environment used to check the fix on devnet and more generally rolling update. I will remove it altogether.

code/crates/discovery/src/handlers/close.rs
ancazamfir: Yes. We will start an epic on discovery enabled case and will do as part of that.

code/crates/discovery/src/handlers/dial.rs
ancazamfir: Let's leave it like this for now. We can re-evaluate as part of the bigger effort on p2p/discovery that we will start immediately after this PR is merged.

code/crates/network/test/tests/discovery.rs
ancazamfir: All bootstrap connections are considered incoming. The maximum number would be 4. With the limit `num_incoming_peers: 2` different peers will reject/ delete different connections and, with the retries, the exact connections may and up differently than what the test expected.

code/crates/test/tests/it/value_sync.rs
ancazamfir: When I wrote these tests I tried to catch the timing such that node 4 is in sync mode when the validators restart. After a few tries these values worked so I stopped there.

code/crates/network/src/lib.rs
```diff
@@ -291,6 +289,11 @@ async fn run(
         };
     }
 
+    // Timer to periodically try reconnecting to persistent peers
+    // TODO: Using 1 second for now, for faster reconnection during testing
+    // Maybe adjust via config in the future
+    let mut persistent_peer_timer = tokio::time::interval(std::time::Duration::from_secs(1));
+
```
ancazamfir, line 296: Any comments on this? 
The actual retry intervals given the fibonacci retry on connection failure will be:
1, 1, 2, 3, 5, 8, 1, 1, 2, 3, 5, 8,...
@cason @nenadmilosevic95 @romac ??

romac: I am not sure what the question is? Are you asking if we should eventually switch to fibonacci backoff?

romac: Just a note that the first tick of this interval timer will complete immediately, which is perhaps not desired?

If not, we can use [`interval_at`](https://docs.rs/tokio/latest/tokio/time/fn.interval_at.html) instead:

```rust
let period = Duration::from_secs(1);
let mut persistent_peer_timer = tokio::time::interval_at(Instant::now() + period, period);
```

We should also probably [set the missed tick behavior](https://docs.rs/tokio/latest/tokio/time/struct.Interval.html#method.set_missed_tick_behavior) to skip any missed ticks instead of the default burst strategy:

```rust
persistent_peer_timer.set_missed_tick_behavior(MissedTickBehavior::Skip);
```

See the docs for [`MissedTickBehavior`](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html) for more info.

ancazamfir: > I am not sure what the question is? Are you asking if we should eventually switch to fibonacci backoff?

Sorry, the question was for the TODO, if we should leave it hardcoded, 1sec? other?

ancazamfir: > Just a note that the first tick of this interval timer will complete immediately, which is perhaps not desired?

I think it's fine for this PR, we used to `dial_bootstrap_nodes()` before the loop.

nenadmilosevic95: I think it makes sense to make it larger, as we are maintaining the Fibonacci retry for now, so this one should be used only in case all retries triggered by Fibonacci failed, so I guess this one should be >20s?

cason: 10s?

cason: This is complex, in any case, since we are considering only the sleep time of the Fibonacci, not the time to a connection attempt to fail, which may vary a lot. We want here to pick guys to whom attempts have failed but are set as persistent.

cason: Being more precise here. In the worst case, we check a persistent peer, it is dialling in the Fibonacci approach, then we sleep, it is still in the Fibonacci approach, then we sleep again and it is available to dialling. If after the first sleep, the Fibonacci approach is just about to finish, we wait for a full sleep interval.

cason:
Did you just dropped all tesnets and testing? :/

I think you can open for reviews.

ancazamfir:
> Did you just dropped all tesnets and testing? :/

Yes, moved them in `debug-socket-leak` branch.

code/TESTNET_SCENARIOS.md
ancazamfir: cleaned in `debug-socket-leak`

code/makefile-scripts/setup-mixed-testnet.sh
ancazamfir: removed

code/TESTNET_SCENARIOS.md
ancazamfir: removed

code/crates/network/src/lib.rs
ancazamfir: Let's leave it 1sec for now. We can revise later.

romac (approved):
Amazing work 🚀

cason:
Closes https://circlepay.atlassian.net/browse/CCHAIN-407

PR #1200, opened 2025-09-12
chore: remove qa.yml workflow

xiuming-circle:
remove `qa.yml` workflow


nenadmilosevic95 (commented):
There are also scripts that only this workflow is using, should we remove them also or not?

hvanz:
> There are also scripts that only this workflow is using, should we remove them also or not?

I'd say, yes. If those scripts are not used anywhere else, then remove them so we don't have to maintain them. 

Is the whole `./qa` directory that we would need to remove? Many of the files in there will eventually come back to the `circle-chain-reth` repo as part of the Quake testnet tool.

PR #1211, opened 2025-09-22
fix(CCHAIN-489): Full node sync sometimes stuck after all peers restart

ancazamfir:
Closes: #XXX
See new added test `full_node_sync_after_all_persistent_peer_restart`:

- full node connects to validators
- starts sync
- validators are stopped while there are pending requests
- connections are torn down
- sync requests time out:
```
2025-09-22T14:09:57.216391Z  WARN test{id=1}:node{id=4}:node{moniker=node-4}:sync{height.tip=6 height.sync=8}: Sync request timed out peer_id=12D3KooWBTia5TdzdZ7mDiGu1dSPJFHYMguV9QQCUcm1RBzSqEvL range=7..=7
2025-09-22T14:09:57.216501Z  INFO test{id=1}:node{id=4}:node{moniker=node-4}:sync{height.tip=6 height.sync=8}: Re-requesting values from peer request_id=7 except_peer_id=Some(PeerId("12D3KooWBTia5TdzdZ7mDiGu1dSPJFHYMguV9QQCUcm1RBzSqEvL"))
2025-09-22T14:09:57.216624Z ERROR test{id=1}:node{id=4}:node{moniker=node-4}:sync{height.tip=6 height.sync=8}: No peer to request sync from range.sync=7..=7
```
- the timed out requests are stuck in pending queue and never retried
- sync blocks at first timed out request

cc: @nenadmilosevic95 @hvanz 

---

#### For all contributors


cason:
Similar behaviour was observed in the first testnet (now devnet).

code/crates/sync/src/handle.rs
```diff
@@ -498,8 +498,23 @@ where
...
+            if state
+                .pending_requests
+                .values()
+                .any(|(range, _)| range.contains(&next_height))
+            {
+                break;
+            }
+            end_height = next_height;
+        }
+
```
ancazamfir: Maybe we can reduce the complexity here by not doing the check `max_batch_size` times but rather check for range overlap. 

Something like:
```
        let mut end_height = start_height.increment_by(max_batch_size - 1);
        
        end_height = state
            .pending_requests
            .values()
            .find_map(|(r, _)| {
                // Check if this range overlaps and would require truncation
                if r.start() <= &end_height && r.end() >= &start_height {
                    if let Some(truncate_at) = r.start().decrement() {
                        if truncate_at >= start_height {
                            return Some(truncate_at);
                        }
                    }
                }
                None
            })
            .unwrap_or(end_height);
```
Ideally we should pull this into a function and write UTs for it. The proposal for example assumes that the ranges in `pending_requests` are disjoint (which I think is the case). It would need to be re-written if not true or we don't want to make the assumption.

code/crates/sync/src/handle.rs
```diff
@@ -554,8 +569,22 @@ where
...
+    // Find the next height that's not covered by any pending request
+    let mut next_sync_height = max(state.sync_height, range.end().increment());
+    while let Some((covered_range, _)) = state
+        .pending_requests
+        .values()
+        .find(|(r, _)| r.contains(&next_sync_height))
+    {
+        next_sync_height = covered_range.end().increment();
+    }
+    state.sync_height = next_sync_height;
```
ancazamfir: It's unfortunate that we do this update here and not in `request_values()`. Maybe there is a way to clean the logic across the two functions.
And also called from `re_request_values_from_peer_except()`.

ancazamfir:
Looks great @nenadmilosevic95. My only comment is if we can have the tests in table format so it's easier to read them. I made the changes in a branch, see https://github.com/ancazamfir/malachite/blob/table_format/code/crates/sync/src/handle.rs, let me know what you think.

nenadmilosevic95:
I took a look, and you are right it is much easier to read with this table approach! I like it a lot! Thanks for doing this! The only downside is that you can't run individual experiment, but I think in this case this is not really necessary. So I would keep the approach you suggested!

code/crates/sync/src/handle.rs
```diff
@@ -629,3 +639,382 @@ impl<'a, Ctx: Context> core::fmt::Display for DisplayRange<'a, Ctx> {
         write!(f, "{}..={}", self.0.start(), self.0.end())
     }
 }
+
+/// Find the next uncovered range starting from initial_height.
+///
+/// Builds a contiguous range of the specified max_size from initial_height.
+///
+/// # Assumptions
+/// - All ranges in pending_requests are disjoint (non-overlapping)
```
melekes, line 648: side note: we've discovered in Comet that duplicating requests (ie. sending requests to two nodes) yields significantly faster sync bacause a single node could be slow and this forces you to wait a timeout (not sure how it's done here) before reissuing the request.

melekes (approved):
utACK 👍

code/crates/sync/src/handle.rs
romac: Interesting, we might consider doing this in the future as well then!

romac (approved):
Amazing work, tests and all! 🚀

PR #1222, opened 2025-10-03
fix(code): Process reply to `AppMsg::Decided` asynchronously

romac:
Closes: #XXX

---

#### For all contributors

### DIFF
--- a/code/crates/app-channel/src/connector.rs
+++ b/code/crates/app-channel/src/connector.rs
@@ -208,11 +208,14 @@ where
                     })
                     .await?;
 
-                let next = rx.await?;
-
-                if let Err(e) = reply_to.send(next) {
-                    error!("Failed to send next height and validator set: {e}");
-                }
+                // Do not block processing of other messages while waiting for the next height
+                tokio::spawn(async move {
+                    if let Ok(next) = rx.await {
+                        if let Err(e) = reply_to.send(next) {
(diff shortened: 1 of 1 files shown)


ancazamfir:
looks good. Will do some testing. For sync but also because now timing will change for consensus as it is more likely to receive "future" messages compared with before.

ancazamfir:
Sync looks good! Helper with 2sec decide continuously responds to sync requests.

cason:
Questions:

1. Could this also apply to other messages?
2. If 1, should we generalize the solution, namely have a method to asynchronously send any reply?
3. How is this thread creating cost, in general? Is it relevant?

ancazamfir:
Consensus also looks good.

romac:
> 1. Could this also apply to other messages?

It could be, we should review this carefully at some point, and perhaps encode whether or not a message can be processed sync or async in the type system to prevent such mismatch of semantics.

> 2. If 1, should we generalize the solution, namely have a method to asynchronously send any reply?

Yes, definitely something we would look into then.

> 3. How is this thread creating cost, in general? Is it relevant?

Strictly speaking, `tokio::spawn` creates an async task not a system thread, the former are very cheap, whose cost is pretty much negligible.

cason:
https://circlepay.atlassian.net/browse/CCHAIN-557

PR #1227, opened 2025-10-08
feat(code)!: Allow dynamic adjustment of timeout parameters

melekes:
This pull request introduces support for passing consensus timeouts as part of the initialization and height-changing logic in the consensus engine. The changes propagate the `Timeouts` trait through the configuration, message types, and core consensus logic, allowing dynamic adjustment of timeout parameters when starting or restarting consensus at a new height. Additionally, the previous static timeout configuration is removed from the config crate, centralizing timeout management in the consensus core.

```rs
// ConsensusReady message
ConsensusReady {
        /// Channel for sending back the height to start at,
        /// the validator set for that height, and optionally the timeouts
        reply: Reply<(
            Ctx::Height,
            HeightParams {
              validator_set: Ctx::ValidatorSet,
              timeouts: Ctx::Timeouts,
            },
        )>,
    },

// StartHeight message
StartHeight(
        Ctx::Height,
        HeightParams {
           validator_set: Ctx::ValidatorSet,
           timeouts: Ctx::Timeouts,
        },
    ),

// RestartHeight message
RestartHeight(
        Ctx::Height,
        HeightParams {
           validator_set: Ctx::ValidatorSet,
          timeouts: Ctx::Timeouts,
        },
    ),
```

---

#### For all contributors

### DIFF
--- a/code/crates/test/tests/it/timeout_updates.rs
+++ b/code/crates/test/tests/it/timeout_updates.rs
@@ -95,33 +95,6 @@ async fn different_timeouts_per_node() {
     test.build().run(Duration::from_secs(90)).await
 }
 
-/// Test that timeout changes only apply at height boundaries, not mid-round
-#[tokio::test]
-async fn timeout_changes_at_height_boundaries() {
-    const HEIGHT: u64 = 4;
-    const CHANGE_HEIGHT: u64 = 2;
-
-    let middleware = TimeoutChangingMiddleware {
-        change_height: CHANGE_HEIGHT,
-        new_propose_timeout: Duration::from_millis(100),
-    };
-
-    let mut test = TestBuilder::<()>::new();
-
-    // All three nodes change timeouts at height 2
-    for _ in 0..3 {
-        test.add_node()
-            .with_middleware(middleware)
-            .start()
-            .wait_until(HEIGHT)
-            .success();
-    }
-
-    // The test succeeds if consensus reaches HEIGHT, which means
-    // that timeout changes applied correctly at height boundaries
-    test.build().run(Duration::from_secs(60)).await
-}
-
 /// A middleware that uses extremely short timeouts to test that they're actually being applied
 #[derive(Copy, Clone, Debug)]
 struct VeryShortTimeouts;


code/crates/starknet/p2p-types/Cargo.toml
```diff
@@ -9,6 +9,7 @@ publish = false
 
 [dependencies]
 malachitebft-core-types.workspace = true
+malachitebft-engine.workspace = true
```
romac: Is this new dependency actually needed?

ancazamfir:
Looks good after a first pass.
One question on:
> Option<malachitebft_app::types::core::Timeouts>,

can this just be `malachitebft_app::types::core::Timeouts,`? If these are going to be stored in a contract and we need to retrieve them at every height, we could simplify the code to always send them, even if they don't change. We have the same approach for `ValidatorSet`. It is true that consensus parameters will rarely change but also validator set is not going to change too often. Was the infrequent change the reason you chose `Option` here? Or were there other considerations?

melekes:
> we need to retrieve them at every height

I don't think that's a hard requirement 🤷 

> Was the infrequent change the reason you chose Option here? Or were there other considerations?

I was designing with other developers in mind. Like it's easier to pass `None` than `Timeouts::default()`, no? If they don't change in your app or change very infrequently. Furthermore, I was thinking of transforming this into:

```rs
StartHeight(
        Ctx::Height,
        HeightUpdates{
              validator_set: Option<Ctx::ValidatorSet>,
              timeouts: Option<malachitebft_app::types::core::Timeouts>,
        }
),
```

because val set also can stay constant.

ancazamfir:
> I was designing with other developers in mind. Like it's easier to pass None than Timeouts::default(), no? If they don't change in your app or change very infrequently. Furthermore, I was thinking of transforming this into:

Good points 👍

cason:
Interesting changes. But since we are changing a bit this, I would:

1. Have `Timeouts` as an interface, I think it would be a `trait` in Rust
2. What consensus needs is `timeout_duration(kind, round)`, something like that should be the method
3. We keep the current implementation as linear timeouts, or something like that, where timeout `kind` at `round` = `timeout_kind + round * timeout_kind_delta`

What do you think?

melekes:
> Have Timeouts as an interface, I think it would be a trait in Rust

That's how I implemented it first. What benefits do you see from this being an interface rather than a fixed struct? Do you ever expect someone needing a non-linear implementation?

cason:
> Do you ever expect someone needing a non-linear implementation?

Yes.

melekes:
> > Do you ever expect someone needing a non-linear implementation?
> 
> Yes.

Done ✅

melekes:
```rs
StartHeight(
        Ctx::Height,
        HeightUpdates{
              validator_set: Option<Ctx::ValidatorSet>,
              timeouts: Option<malachitebft_app::types::core::Timeouts>,
        }
),
```

should I proceed with this change?

romac:
> ```rust
> StartHeight(
>         Ctx::Height,
>         HeightUpdates{
>               validator_set: Option<Ctx::ValidatorSet>,
>               timeouts: Option<malachitebft_app::types::core::Timeouts>,
>         }
> ),
> ```
> 
> should I proceed with this change?

Go ahead! :)

code/crates/engine/src/consensus.rs
```diff
@@ -364,12 +346,7 @@ where
         let is_restart = matches!(msg, Msg::RestartHeight(_, _));
 
         match msg {
-            Msg::StartHeight(height, validator_set) | Msg::RestartHeight(height, validator_set) => {
-                // Check that the validator set is not empty
```
melekes, line 379: note this check was dropped.

romac:
Do we really need `HeightUpdates` for `RestartHeight`? I would think that we'd want to always restart it with the same parameters, wdyt?

melekes:
> Do we really need `HeightUpdates` for `RestartHeight`? I would think that we'd want to always restart it with the same parameters, wdyt?

I would say no (no need), but will defer to @ancazamfir and @cason

code/crates/test/app/src/app.rs
```diff
@@ -40,7 +40,12 @@ pub async fn run(state: &mut State, channels: &mut Channels<TestContext>) -> eyr
                     .get_validator_set(start_height)
                     .expect("Validator set should be available");
 
-                if reply.send((start_height, validator_set)).is_err() {
+                let height_updates = HeightUpdates {
+                    validator_set: Some(validator_set),
+                    timeouts: None,
+                };
```
romac: Can we add a fluent interface to this type to make the code slightly nicer?


```suggestion
                let height_updates = HeightUpdates::default().with_validator_set(validator_set);
```

BREAKING_CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ### `malachitebft-core-types`
```
romac, line 5: ❤️

code/crates/starknet/p2p-types/Cargo.toml
melekes: yes, wal.rs test uses it.

cason:
> Do we really need `HeightUpdates` for `RestartHeight`? I would think that we'd want to always restart it with the same parameters, wdyt?

We should not even have this event. But since we have it, it looks natural to me that we can change anything we want.

romac:
> I think that the core-driver should not store timeouts. In particular because they don't belong are not used there. I think timeout implementation belongs to core-consensus.

Yes good point, especially since it does not make use of them!

@melekes Can you please pull them out of the driver?

melekes:
> @melekes Can you please pull them out of the driver?

will do

melekes:
> > @melekes Can you please pull them out of the driver?
> 
> will do

done ✅

code/crates/engine/src/consensus.rs
```diff
@@ -313,6 +274,9 @@ where
         state: &mut State<Ctx>,
         input: ConsensusInput<Ctx>,
     ) -> Result<(), ConsensusError<Ctx>> {
+        // Extract timeouts before the process! macro to avoid borrow checker issues
+        let timeouts = *state.consensus.timeouts();
```
melekes: I think this is okay, but lmk

melekes: I.e. there's no strict need to store timeouts in HandlerState because timeouts are stateless now, but didn't find a better way

romac: I believe we can simplify this further by removing the timeouts from the core-consensus state, since it actually never uses them. See my attempt at https://github.com/melekes/malachite/pull/1, let me know what you think?

romac:
> I am still not sure if `HeightUpdates` is the best name here

Any suggestions?

cason:
> > I am still not sure if `HeightUpdates` is the best name here
> 
> Any suggestions?

How bad is the old familiar `ConsensusParams`?

melekes:
> > > I am still not sure if `HeightUpdates` is the best name here
> > 
> > 
> > Any suggestions?
> 
> How bad is the old familiar `ConsensusParams`?

isn't it too abstract? given the context and naming of enum

romac:
Maybe `ConsensusParamsUpdates` or just `ParamsUpdates`? (a bit shorter)

melekes:
> Maybe `ConsensusParamsUpdates` or just `ParamsUpdates`? (a bit shorter)

or just `Updates` 😄

romac:
> > Maybe `ConsensusParamsUpdates` or just `ParamsUpdates`? (a bit shorter)
> 
> or just `Updates` 😄

Sounds good!

melekes:
> > > Maybe `ConsensusParamsUpdates` or just `ParamsUpdates`? (a bit shorter)
> > 
> > 
> > or just `Updates` 😄
> 
> Sounds good!

@cason wdyt?

melekes:
> > Maybe `ConsensusParamsUpdates` or just `ParamsUpdates`? (a bit shorter)
> 
> or just `Updates` 😄

done ✅

romac (approved):

code/crates/core-driver/src/driver.rs
```diff
@@ -106,21 +106,30 @@ where
...
         let proposal_keeper = ProposalKeeper::new();
 
+        // Update the validator set if provided
+        if let Some(validator_set) = validator_set {
+            self.validator_set = validator_set;
+        }
+
         // Reset the vote keeper
-        let vote_keeper = VoteKeeper::new(validator_set.clone(), self.threshold_params);
+        // XXX: must be called after updating the validator set
```
ancazamfir: Remove XXX?

code/crates/test/app/src/app.rs
```diff
@@ -216,8 +218,13 @@ pub async fn run(state: &mut State, channels: &mut Channels<TestContext>) -> eyr
                             .get_validator_set(state.current_height)
                             .expect("Validator set should be available");
 
+                        let updates = Updates {
+                            validator_set: Some(validator_set),
```
ancazamfir: Could we change the code here to compare old and new validator sets and pass None if the same? This way we can test the `None` case for valset in malachite.

ancazamfir:
Should we add some tests for timeout changes now that we allow it without restart?

code/crates/core-driver/src/driver.rs
melekes: can't do that because I'm using `self.validator_set` which is updated above. that's why XXX is there

code/crates/test/app/src/app.rs
melekes: https://github.com/circlefin/malachite/pull/1227/commits/1f14c3f7a4af732bd4a6949272d10a01bc4a82a9

melekes:
> test timeout updates

https://github.com/circlefin/malachite/pull/1227/commits/b91d517e800b6b044e8334a398e8ff273bfe756e

code/crates/app-channel/src/msgs.rs
```diff
@@ -300,27 +299,27 @@ pub enum AppMsg<Ctx: Context> {
 /// Messages sent from the application to consensus.
 #[derive_where(Debug)]
 pub enum ConsensusMsg<Ctx: Context> {
-    /// Instructs consensus to start a new height with the given validator set.
-    StartHeight(Ctx::Height, Ctx::ValidatorSet),
+    /// Instructs consensus to start a new height with optional updates.
+    StartHeight(Ctx::Height, Updates<Ctx>),
```
romac: I am only realizing now that having validator set and timeouts being optional here opens the door to an application not providing the initial validator set and timeouts for the first height. We could document it but I think I'd rathersolve statically by making both required.

romac: I can take a stab at this.

code/crates/core-driver/src/driver.rs
ancazamfir: I meant remove "XXX" from the comment

code/crates/app-channel/src/msgs.rs
melekes: > I am only realizing now that having validator set and timeouts being optional here opens the door to an application not providing the initial validator set and timeouts for the first height. We could document it but I think I'd rathersolve statically by making both required.

Note that was not the case when we had `initial_validator_set` and `initial_timeouts`, which I personally liked for ^ exact reason - it gives developer a way to provide 1st validator set / timeouts.

melekes: Now it's convoluted, yeah. I can make both args required if folks think it's a way to go. @romac feel free to change also.

code/crates/test/app/src/app.rs
```diff
@@ -232,8 +260,13 @@ pub async fn run(state: &mut State, channels: &mut Channels<TestContext>) -> eyr
...
```
ancazamfir: I think we can pass `None` here as we restart the same height with same valset as before.
```suggestion
                            validator_set: None,
```

code/crates/app-channel/src/msgs.rs
romac: Yeah there is some unfortunate interplay between the two PRs, but imho removing these `initial_` parameters was the right thing to do outside of the scope of this PR. And with this PR (which is otherwise great by the way!) I'd rather make parameters required both so that they cannot be forgotten but also because it makes the internal logic less stateful.

I am on it, will push a couple commits soon.

code/crates/test/tests/it/timeout_updates.rs
```diff
@@ -0,0 +1,211 @@
...
+        .start()
+        .wait_until(HEIGHT)
+        .success();
+
+    test.build().run(Duration::from_secs(90)).await
+}
+
+/// Test that timeout changes only apply at height boundaries, not mid-round
+#[tokio::test]
+async fn timeout_changes_at_height_boundaries() {
```
ancazamfir: Do we need this? Seems to be the same as `change_timeouts_between_heights()` just different values.

code/crates/test/app/src/app.rs
romac: The `validator_set` is now always required: https://github.com/circlefin/malachite/pull/1227/commits/3db51032ce028e5f60486852efdfc3661522f54c

code/crates/test/tests/it/timeout_updates.rs
romac: Yeah, and I am not sure if it's possible to test what the comment says. I think we can just remove this test.

melekes:
Thanks for the last push @romac 🙏

code/crates/test/tests/it/timeout_updates.rs
ancazamfir: let's remove

romac:
@melekes Thanks for the great work!

PR #1230, opened 2025-10-14
chore(code): Lower some warnings to debug or info

romac:
Closes: #XXX

Story: https://circlepay.atlassian.net/browse/CCHAIN-602

---

#### For all contributors

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,9 @@ We are grateful to StarkWare Industries for prompting the initial discussions of
 We are also thankful for the collaboration with Farcaster.
 This led to further refinements and maturing of the Malachite codebase, and their approach to building complex systems and shipping valuable products is an inspiration for us.
 
+We are grateful as well for our collaboration with the Shielded Labs team.
+Their Crosslink system, a PoS + PoW hybrid design for upgrading Zcash, has helped deepen our understanding of BFT system design and push the limits of what Malachite is capable of supporting in terms of novel decentralized systems.
+
 These collaborations reflect work completed prior to the Malachite team joining Circle.
 
 ## License


code/crates/core-consensus/src/state.rs
cason, line 1: Having all these messages in a single line would be helpful.

code/crates/engine/src/wal.rs
```diff
@@ -100,7 +99,7 @@ where
 
             Msg::Append(height, entry, reply_to) => {
                 if height != state.height {
-                    warn!(wal.height = %state.height, entry.height = %height, "Ignoring append, mismatched height");
+                    debug!(wal.height = %state.height, entry.height = %height, "Ignoring append, mismatched height");
```
cason: This was changed by other PR, should remain `warn`

code/crates/engine/src/util/msg_buffer.rs
```diff
@@ -24,7 +24,7 @@ impl<Ctx: Context> MessageBuffer<Ctx> {
             self.messages.push_back(msg);
             true
         } else {
-            warn!("Buffer is full, dropping message: {msg:?}");
+            info!("Buffer is full, dropping message: {msg:?}");
```
cason: How often do we see this message?

The issue here is that once this condition is reached we are potentially dropping lots of messages...

code/crates/core-consensus/src/handle/propose.rs
```diff
@@ -90,7 +90,7 @@ where
     Ctx: Context,
 {
     if state.height() != local_value.height {
-        warn!(
+        debug!(
```
cason: Is this supposed to happen? We receive a proposed value at the wrong height/round?

code/crates/core-consensus/src/handle/driver.rs
cason, line 1: I think that wrong heights shouldn't happen with the driver, should them?

If this should not happen, we should use `warn!` or even worse, use assertions.

cason (commented):
Thank for this. I have some doubts if we should change the driver's messages, in particular.

code/crates/engine/src/util/msg_buffer.rs
romac: Good point, should remain a warning indeed

romac:
There are a lot of other warnings in the discovery crate which I have not touched so far.

@ancazamfir Maybe we can look at those once you are done with the ongoing discovery work?

code/crates/core-consensus/src/handle/driver.rs
ancazamfir: It cannot/ shoud not happen.  for `Proposal`, `Vote`, `CommitCertificate` we discard lower heights and we buffer in core-consensus handlers if for higher heights, to be processed when we reach that height.
We discard `PolkaCertificates` for different heights or lower rounds.
So yes, these should stay warn!

code/crates/core-consensus/src/handle/propose.rs
ancazamfir: Cannot happen for different height. For some apps can potentially happen for different round I believe. If they use the timeout and the proposed value is not built in time and we timeout, move to the next round, etc. In any case these should stay warn.

code/crates/engine/src/consensus.rs
ancazamfir, line 1: There is an `info!` in line 561:
```
NetworkEvent::RoundCertificate(...) =>
    info!(
        %from
         ...
```
that we should remove as it's done in the handler

code/crates/core-consensus/src/handle/driver.rs
```diff
@@ -159,7 +159,7 @@ where
 
         if let Some(valid) = &state.driver.valid_value() {
             if state.driver.step_is_propose() {
-                info!(
+                debug!(
```
ancazamfir: I think info! should be fine, this is a rare event and we want to "see" it.

code/crates/engine/src/consensus.rs
romac: https://github.com/circlefin/malachite/pull/1230/commits/56d1d0e939cb089118e6a28f5671abd809d45ceb

code/crates/core-consensus/src/handle/driver.rs
romac: https://github.com/circlefin/malachite/pull/1230/commits/e8eb1bfb98550df586029d4ab56f8f5275f33b94

code/crates/core-consensus/src/handle/driver.rs
romac: https://github.com/circlefin/malachite/pull/1230/commits/70523f1722c6815c49ca6e1c70463d7feea2cc8a

ancazamfir (approved):
👍

Issue #1232, opened 2025-10-15
WAL entries emitted when replaying pending inputs on `StartHeight` are ignored by the WAL


romac:
When processing a consensus input, consensus may yield WAL entries to be appended to the WAL. Each entry is associated with a height. In order for a WAL entry to be appended to the WAL, the height associated with that entry must match the height that the WAL is presently at, otherwise the entry is ignored with a `Ignoring append, mismatched height` message.

The height associated with each entry is taken from the `HandlerState`, and is set right before processing the input. That height is then retrieved in the `WalAppend` effect handler and passed along the entry to `wal_append`.

When `consensus-core` processes the `StartHeight` input, it will replay any pending inputs for that height, and potentially emit WAL entries to be appended to the WAL. The issue is that the height we associate with these inputs is by that time outdated. Indeed, for `StartHeight`, the height associated with these entries is the height prior to the height having been started. This causes any WAL entry being emitted while processing a `StartHeight` input to be ignored by the WAL.

PR #1233, opened 2025-10-15
fix(code/wal): Fix mismatched height of WAL entries emitted when processing `StartHeight` input

romac:
Closes: #1232

**Story:** <https://circlepay.atlassian.net/browse/CCHAIN-578>

---

#### For all contributors

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,6 +10,7 @@
 - Fix bug in WAL recovery logic where a corrupted entry would not be detected in some circumstances ([#1127](https://github.com/informalsystems/malachite/pull/1127))
 - Add facility for app to request a consensus state dump at any time ([#1176](https://github.com/informalsystems/malachite/pull/1176))
 - Make libp2p protocol names configurable ([#1161](https://github.com/informalsystems/malachite/issues/1161))
+- Fix mismatched height of WAL entries emitted when processing `StartHeight` input ([#1232](https://github.com/circlefin/malachite/issues/1232))
 
 ## 0.5.0
 


code/crates/engine/src/consensus.rs
```diff
@@ -314,19 +313,13 @@ where
         state: &mut State<Ctx>,
         input: ConsensusInput<Ctx>,
     ) -> Result<(), ConsensusError<Ctx>> {
-        let height = state.height();
-
-        // By the time the effect is processed the state height might have changed.
-        // This happens for input Msg::StartHeight(height), so height is potentially stale.
```
ancazamfir, line 320: There is another similar comment below for sync value that we should remove.

code/crates/engine/src/wal.rs
```diff
@@ -100,12 +99,20 @@ where
 
             Msg::Append(height, entry, reply_to) => {
                 if height != state.height {
-                    warn!(wal.height = %state.height, entry.height = %height, "Ignoring append, mismatched height");
+                    debug!(
```
ancazamfir: I would keep this `warn!` as something is broken if it happens.

code/crates/engine/src/consensus.rs
romac: ef4bc852

code/crates/engine/src/wal.rs
romac: 0c9c3b9e

romac: I also removed the noisy debug log that I had previously added below: 5ce78a9b

PR #1292, opened 2025-12-09
chore(stepsecurity): update workflows to use custom hosted runners with built-in StepSecurity

circle-ops-repo-updater:
## Summary
This PR updates GitHub Actions workflows to use custom hosted runners that have StepSecurity built-in, removing the need for the explicit StepSecurity harden-runner action.

## What Changed
- Removed step-security/harden-runner action steps (no longer needed as StepSecurity is built into custom runners)
- Removed id-token: write permissions (no longer needed without the StepSecurity action)
- Updated runs-on from ubuntu-latest to github-hosted-small (custom runners with built-in StepSecurity)
- Converted non-circlefin action versions to commit SHAs with version comments for security pinning (e.g., actions/checkout@abc123 # v3.6.0)
- circlefin GitHub actions remain unchanged

## Purpose
Our custom hosted runners (github-hosted-small) now have StepSecurity built-in at the runner level, so we no longer need to add it as an explicit step in each workflow. This simplifies our workflows while maintaining the same security posture.

## Testing
- All workflow syntax changes have been validated
- No functional changes to workflow behavior
- StepSecurity protection is maintained via the custom runners
- Review the diff to ensure only intended changes occurred

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,10 +1,1 @@
-/specs/ @cason @josef-widder
-/code/ @ancazamfir @romac
-/code/examples @ancazamfir @romac
-/code/crates/discovery @bastienfaivre
-/docs/ @adizere @ancazamfir @romac
-/scripts/ @ancazamfir @romac
-/qa/ @greg-szabo @nenadmilosevic95
-/*.md @adizere @ancazamfir @romac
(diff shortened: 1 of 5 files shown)


romac:
This PR is outdated and should likely be recreated.

PR #1371, opened 2026-01-14
feat: Upstream Arc-related changes

romac:
Closes: #1364

> [!IMPORTANT]
> This PR must be **rebased** and **NOT squashed**


adizere:
cc @insumity @jmalicevic there are many good changes here which accumulated in the last period, and we're upstreaming here -- the release notes should be up-to-date, and each commit should be self-contained!

ams9198 (approved):

PR #1408, opened 2026-01-27
feat(docs): ADR on the consensus Write-Ahead Log (WAL)

### LINKED ISSUE #469: spec: Consensus Write-Ahead Log (WAL)
In order to support the crash-recovery failure model (#578), the consensus implementation should persist all relevant events that have lead it to its current state. When recovering from a crash, the implementation is initialized from its initial state of the latest active height `H`, then has to replay all the information persisted before it has crashed.

The valid events processed by the consensus implementation are therefore typically persisted in a Write-Ahead Log (WAL), an append-only log that was originally conceived to ensure atomicity of transactions in databases.

## Definition of Done

- [ ] Write an ADR discussing the problem and defining:
   - Which events should be persisted to the WAL
   - The strategy for writing data to the WAL (e.g., synchronously versus asynchronously)
   - The procedure for replaying the content of the WAL when a node is (re)started
- [ ] Check if the implementation aligns with the design

cason:
Closes #469.

[Rendered](https://github.com/cason/malachite/blob/adr-wal/docs/architecture/adr-007-write-ahead-log.md)


docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,154 @@
...
+collects single inputs to produce complex events for the consensus state machine.
+The [driver](./adr-001-architecture.md#consensus-driver) is also responsible for
+removing the non-determinism present in Tendermint's [pseudo-code][pseudo-code],
+where in the case when multiple `upon` clauses can be activated, any of them
+could be chosen.
+The driver establishes priorities between them so that to render the operation
+deterministic.
+So, ideally, it should be the right layer to invoke the WAL primitive for the
+logging of inputs and for replaying WAL entries retrieved from persistent
+storage upon recovery.
```
romac: It's not clear to me why that is, given all the downsides listed below?

What is the actual upside that you see of making the driver aware of the WAL versus leaving it in core-consensus?

cason: Testing.

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+the [malachitebft-core-consensus][consensus-crate] crate,
+is the interface between consensus and the host.
+On the one hand, it receives inputs from the network, the application, or other
+components, process and forward them to the driver.
+On the other hand, it interacts the host to request some actions
+([Effects](./adr-004-coroutine-effect-system.md#effect) in in Malachite's parlance)
+to be performed.
+For instance, core consensus layer receives messages from the network
+(`Proposal` or `Vote` inputs), verifies its signatures (`VerifySignature`
+effect), and forwards it to the driver.
```
ancazamfir:
```suggestion
(`Proposal` or `Vote` inputs), verifies their signatures (`VerifySignature`
effect), and forwards them to the driver.
```

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+It the result of this processing is a message, core consensus layer requests
```
ancazamfir:
```suggestion
If the result of this processing is a message, core consensus layer requests
```

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+4. Application input `ProposedValue`: received consensus value `v` and its
+   validity `valid(v)`;
+
+The case of consensus messages is straightforward, as their reception leads to
+progress in the consensus protocol.
+
+The case of expired timeouts is less evident.
+Timeouts are scheduled when some conditions on the received messages are
+observed.
+While their expiration leads to state transitions, provided that the process is
```
ancazamfir:
```suggestion
Their expiration leads to state transitions, provided that the process is
```

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+> For reviewers: while looked obvious in the first design, persisting the
+> application inputs (values) appears not to be so crucial, when a
+> well-behaving application is considered.
+
+### Checkpoints
+
+A WAL enables crash-recovery behavior in systems that can be modelled as
+deterministic state machines.
+This, for example, that starting from an initial state `s0` and applying
+inputs `i1` and `i2`, the system transitions to states `s1` and `s2`, respectively.
```
ancazamfir:
```suggestion
For example when starting from an initial state `s0` and applying
inputs `i1` and `i2`, the system transitions to states `s1` and `s2`, respectively.
```
Not sure it's the best but the sentence doesn't read well.

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+every processed input preceding the production of the output.
+This definition enables the following operational design for the WAL component:
+
+1. Log all processed inputs, in the order they have been processed, without
+   persistence guarantees -> asynchronous writes;
+2. When an output is produced, and before emitting it, persist all inputs that
+   were not yet persisted -> synchronous writes or `flush()`.
+
+Put in different words, inputs that do not (immediately) lead to an output, or
+to a relevant state transition, can be logged in foreground and in a best
```
ancazamfir:
```suggestion
to a relevant state transition, can be logged in background and in a best
```

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+effort manner.
+While the production of an output demands a synchronous, blocking call to
+persist all the outstanding inputs.
+
+> It is not 100% clear to me if we adopted this approach.
+> I have the impression that all WAL append calls are synchronous.
```
ancazamfir: When it has an input, engine calls `wal_append` which sends a blocking message (`ractor:call!`) to the WAL actor, which forwards it to a dedicated WAL thread that writes the entry to the OS file buffer (a fast syscall, not a disk sync) and responds. This round-trip blocks consensus briefly but doesn't wait for disk. 
Only when consensus produces an output, it calls `wal_flush` which triggers an fsync() so all buffered entries are persisted to disk before the output is emitted.

Maybe we could use `ractor::cast!` for appends to avoid even the message-passing between actors. But not sure. @romac is the expert!

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+
+### Replay
+
+All discussion of previous boils down to this point: how is the operation of a
```
ancazamfir:
```suggestion
All previous discussion boils down to this point: how is the operation of a
```

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+recovery process?
+A first and relevant consideration is that Malachite, a priori, does not
+know if it either in ordinary or recovery operation.
+The consensus layer, in either operation mode, waits for a `StartHeight` input
+from the application indicating the height number `H` to start.
+At this point Malachite should open and load the WAL contents to check if it
+includes entries (inputs) belonging to height `H`:
+if there are, it is in recovery mode; otherwise, in regular operation.
+If the concept of [Checkpoints](#checkpoints) is adopted, this verification is
+even simpler and already described on the associated section (Case 2).
```
ancazamfir:
```suggestion
even simpler and already described in the associated section (Case 2).
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+input, the consensus [driver](./adr-001-architecture.md#consensus-driver) is
+configured with its initial state, set for height `H` with some parameters
+applied (e.g., the validator set).
+If there are WAL entries belonging to height `H`, the process is in recovery
+mode: all height `H` inputs are replayed in the order with which they appear in
+the WAL.
+Once there are no (further) inputs to be replayed, the process starts its
+ordinary operation, processing inputs from the network, from the application,
+and from other protocols.
+
```
ancazamfir, line 297: fwiw, incoming live messages are buffered while in Recovery state.

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+
+The implementation adopted the strategy of producing [checkpoints](#checkpoints)
+at every height, so that the WAL only contains inputs belonging to the latest
+started height.
+The `reset` operation of the `Log` type, implementing the WAL, receives a
+height number, write it to the WAL header, and truncates the WAL.
+It is invoked whenever `StartHeight` input is received from the application
+for the first time for a height.
+
+> It is also use in the case of the `RestartHeight` input, an _unsafe command_
```
ancazamfir:
```suggestion
> It is also used in the case of the `RestartHeight` input, an _unsafe command_
```

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+> that instructs Malachite to _ignore_ the WAL's contents.
+
+Note that all entries in the WAL must belong to the height the WAL is
+configured for, written as part of its header.
+This means that messages from future heights `H'> H`, where `H` is the highest
+height for which a `StartHeight` input was processed, are not persisted.
+The are only buffered in main memory and therefore lost upon restarts.
```
ancazamfir:
```suggestion
They are only buffered in main memory and therefore lost upon restarts.
```

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+when an input produces outputs or relevant state transitions, a blocking
+`flush()` call is needed, implementing a synchronous batch write.
+
+Synchronous writes, using the `wal_flush` consensus engine method, are performed:
+
+1. When a new round of consensus is started, via the `StartRound` effect;
+2. When a consensus message is broadcast, via the `PublishConsensusMsg` effect;
+3. When a value is finalized by the consensus, via the `Decide` effect.
+
+> FIXME: is there any particular reason for item 1, i.e. starting a round?
```
ancazamfir: I can't recall the exact reasoning. `StartedRound` notifies the host, which might be considered an "output" that could trigger external actions (e.g., proposer starts building a block). Flushing makes sure that the state leading to this round is persisted before the host acts on it. Maybe @romac knows.

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+
+A current limitation of the persistence approach, however, is the fact that
+calls to `wal_append` are also blocking, while they do not have to.
+This is possibly associated with error handling, but it would be good to find a
+way to propagate and handle errors asynchronously.
+
+> For reviewers: can you double check the above affirmation?
```
ancazamfir: see my comment above. But we can probably do better

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+**must be critical**.
+At the moment, from this [commit](https://github.com/circlefin/malachite/commit/38f113f6c81da0af32a748718b2d87ab64e3a72f),
+consensus hangs forever in case of any WAL operational error.
+
+A second source of errors is when replaying the WAL, where one particular error
+requires attention: the corruption of the last entries of the WAL.
+Notice that crashes can happen at any time of the execution, including the
+instant at which a WAL entry is being persisted to stable storage.
+A crash at this point will likely render a suffix of the WAL corrupted.
+This, however, is not a treat for safety because the outputs produced by the
```
ancazamfir:
```suggestion
This, however, is not a threat for safety because the outputs produced by the
```

docs/architecture/adr-006-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+associated state-machine transition are only emitted after the WAL is
+[persisted](#persistence-1).
+Since the append operation has not been concluded with success, the output was
+not emitted, therefore it is just like it has never been produced.
+
+In summary, as discussed in this [Jira ticket][jira-corrupt], corruptions at
```
ancazamfir: Not sure if reference to internal Jira ticket is best here.

ancazamfir (commented):
Very nice doc, reads very well!
We should update the other adrs, especially adr-003 which was written more than 2 years ago.

docs/architecture/adr-006-write-ahead-log.md
cason: https://github.com/circlefin/malachite/pull/1408/commits/836c4dcb91e811d448afe1f65809e540409efb3a

docs/architecture/adr-007-write-ahead-log.md
cason: Good point. Does it applies also for other messages, e.g., from the application?

cason: [6484d11](https://github.com/circlefin/malachite/pull/1408/commits/6484d110384114bce09d399089332da91041dcf8)

docs/architecture/adr-006-write-ahead-log.md
cason: It is listed here: https://github.com/circlefin/malachite/issues/469#issuecomment-2488960658, so it was a requirement. But it is not documented why.

docs/architecture/adr-006-write-ahead-log.md
cason: Yes, this comment is a bit outdated, see line 502.

The write is indeed asynchronous, but the `wal_append` method blocks waiting for the reply from the WAL thread. I assume it blocks because the call is like:

```
Effect::WalAppend(height, entry, r) => {
                self.wal_append(height, entry, state.phase).await?;
```

You are right regarding the `ractor:call!`, this is still kind of greek to me.

cason:
> We should update the other adrs, especially adr-003 which was written more than 2 years ago.

It is actually 4 months ago and mostly up-to-date. But I agree that ADR 001 is historical right now.

adizere (approved):
thanks Daniel, lgtm!

docs/architecture/adr-006-write-ahead-log.md
romac: > Maybe we could use ractor::cast! for appends to avoid even the message-passing between actors

Do you mean use `cast!` to simply send a message without waiting for a reply?

romac: Right now, disk writes are async but we still wait for the reply from the WAL to detect any failures at this stage and stop consensus.

docs/architecture/adr-006-write-ahead-log.md
romac: > Flushing makes sure that the state leading to this round is persisted before the host acts on it. 

Yes, I think the reasoning was that `StartRound` is a state machine output and therefore needs to trigger a WAL flush.

ancazamfir:
> > We should update the other adrs, especially adr-003 which was written more than 2 years ago.
> 
> It is actually 4 months ago and mostly up-to-date. But I agree that ADR 001 is historical right now.

oh, i meant adr-001 indeed.

docs/architecture/adr-006-write-ahead-log.md
cason: So my assumption regarding error handling was true. But is it not simpler to just monitor if the WAL thread/actor "died"?

cason: https://github.com/circlefin/malachite/issues/1435, linked to the ADR.

docs/architecture/readme.md
```diff
@@ -30,4 +30,5 @@ To suggest an ADR, please make use of the [ADR template](./adr-template.md) prov
 | [001](./adr-001-architecture.md)            | High Level Architecture                     | Accepted |
 | [002](./adr-002-node-actor.md)              | Node Architecture using the Actor Model     | Accepted |
 | [003](./adr-003-values-propagation.md)      | Propagation of Proposed Values              | Accepted |
-| [004](./adr-004-coroutine-effect-system.md) | Coroutine-Based Effect System for Consensus | Accepted |
\ No newline at end of file
+| [004](./adr-004-coroutine-effect-system.md) | Coroutine-Based Effect System for Consensus | Accepted |
+| [006](./adr-006-write-ahead-log.md)         | Consensus Write-Ahead Log (WAL)             | Accepted |
```
nenadmilosevic95: The number conflicts with PoV ADR, can you use 007 instead?

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+> Notice that consistency is not only a "nice to have" property, but a required one.
+> A process that is not consistent, does operate arbitrarily and becomes a
+> potential Byzantine process.
+>
+> A textbook example of the impact of not properly handling the recovery of a
+> process is the locking mechanism present in multiple consensus algorithms.
+> In Tendermint, a process that emits a `Precommit` for a value `v` also
+> _locks_ `v` at that round.
+> The lock is a promise to not accept a value different than `v` in future
```
nenadmilosevic95, line 28: Unless we have L28-30 but I think for the purpose of this example this is precise enough

docs/architecture/adr-007-write-ahead-log.md
ancazamfir: > Good point. Does it applies also for other messages, e.g., from the application?

yes, mostly. [These](https://github.com/circlefin/malachite/blob/ba38f41388af078ad21dae254f136173e2a0e018/code/crates/engine/src/consensus.rs#L1455) we don't buffer. I think we do buffer:

- consensus messages via NetworkEvent (votes, proposals from peers)
- application messages, `ProposeValue`, `ProposedValue`
- timeout events
- sync responses
- `RestartHeight`

cason: https://github.com/circlefin/malachite/pull/1408/commits/ba70e06c421433b4fea00b15bc949b55149c077f

cason:
Can someone make the checks to run? They are always canceled for some reason.

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+> rounds.
+> So, if a value different than `v` is proposed in the next round `r`, the
+> process rejects it by emitting a `Prevote` for `nil`.
+>
+> The lock on `v` is part of the state of the process.
+> If the process is restarted but its consensus state is not properly
+> recovered, then the process can receive a proposal for a value `v'` in round
+> `r` and misbehave in two ways:
+>
+> * Amnesia: by "forgetting" about the promise associated to the pre-crash
```
nenadmilosevic95, line 39: Amnesia can also be when process doesn't re-propose its valid_value this can cause equivocation.

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+>   lock on `v`, accept the proposed value `v' != v`;
+> * Equivocation: for accepting `v'`, to emit a `Prevote` for `id(v')` in
```
nenadmilosevic95: Minor again but to me this is more a double-signing than equivocation

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+The second layer of the consensus implementation is the
+[driver](./adr-001-architecture.md#consensus-driver) that collects
+single inputs to produce complex events for the consensus state machine.
+Implemented in the [malachitebft-core-driver][driver-crate] crate,
+the driver is also responsible for removing the non-determinism present in
+Tendermint's [pseudo-code][pseudo-code], which in situations when multiple
+`upon` clauses can be activated, assumes that any of them could be chosen at
+random.
+The driver establishes priorities between the potentially complex events, so
+that to render Malachite's operation deterministic.
```
nenadmilosevic95:
```suggestion
The driver establishes priorities among potentially complex events, thereby making Malachite’s operation deterministic.
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+
+The third layer of the consensus implementation,
+the [malachitebft-core-consensus][consensus-crate] crate,
+is the interface between consensus and the host.
+On the one hand, it receives inputs from the network, the application, or other
+components, process and forward them to the driver.
+On the other hand, it interacts the host to request some actions
```
nenadmilosevic95:
```suggestion
On the other hand, it interacts with the host to request some actions
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+([Effects](./adr-004-coroutine-effect-system.md#effect) in in Malachite's parlance)
```
nenadmilosevic95:
```suggestion
([Effects](./adr-004-coroutine-effect-system.md#effect) in Malachite's parlance)
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,572 @@
...
+to be performed.
+For instance, core consensus layer receives messages from the network
+(`Proposal` or `Vote` inputs), verifies their signatures (`VerifySignature`
+effect), and forwards them to the driver.
+If the result of this processing is a message, core consensus layer requests
+its signing (`SignProposal` or `SignVote` effects) to the signer and its
+broadcast (`PublishConsensusMsg` effect) to the network.
```
nenadmilosevic95:
```suggestion
If this processing results in a message, core consensus layer requests
its signing from the signer (`SignProposal` or `SignVote` effects) and its
broadcast to the network (`PublishConsensusMsg` effect) .
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+Since the WAL is a functionality implemented by the host, the interaction of
+the consensus implementation with the WAL will happen through effects, produced
+by core consensus layer.
+
+The core consensus layer is also responsible for implementing the value
+dissemination feature of Malachite.
```
nenadmilosevic95: It is a bit strange to me that here we say that it is implemented by core consensus and then below we say it is the responsibility of the application.

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+The dissemination is usually a role implemented by the application, that
+reports to Malachite received values and their validity via the
+`ProposedValue` core consensus input.
+This input is then typically combined with `Proposal` core consensus input,
+representing the reception of the corresponding message from the network,
+to form the `Proposal` input provided to the driver and the state machine,
+which are unaware of the operation of the value dissemination protocol.
+
+The fourth and last layer of the consensus implementation is the 
+[malachitebft-engine][engine-crate] crate, which provide a standard
```
nenadmilosevic95:
```suggestion
[malachitebft-engine][engine-crate] crate, which provides a standard
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+implementation for all the effects, i.e., interactions with the host system,
+requested the core consensus layer.
```
nenadmilosevic95:
```suggestion
requested by the core consensus layer.
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+It is the engine the actually implements or interacts with the network layer,
```
nenadmilosevic95:
```suggestion
It is the engine that actually implements or interacts with the network layer,
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+### Inputs
+
+A priori, all inputs that change the consensus protocol state or produce an
+output should be persisted to the WAL.
+More specifically:
+
+1. Consensus messages: `Proposal` and `Vote`, the last representing both
+   `Prevote` and `Precomit` votes;
+2. Expired timeouts `TimeoutElapsed(step)`, where `step` is one of
+   `{propose,prevote, precommit}`;
```
nenadmilosevic95:
```suggestion
   `{propose, prevote, precommit}`;
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+When the process is replaying inputs from the WAL during recovery, the ordinary
+consensus execution should schedule the same timeouts scheduled before the
+process has crashed, while processing the same inputs.
+But since it takes time for the timeouts to expire, it is hard to ensure that
+the state of the process when the timeout expires will be the same it was
+before it had crashed.
+As the next state and outputs of the timeout expiration event depends on the
+process state, it must be ensured that the `TimeoutElapsed` inputs are
+replayed during recover in _the same relative order_ to other inputs they were
+before crashing.
```
nenadmilosevic95, line 174: Any reason why wouldn't we reply the events in the same order?

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+In other words, since _time_ is non deterministic, time-based event should be
+logged.
+
+The values proposed by the local instance of the application, when the
```
nenadmilosevic95: I would be specific here and say that we talk about `LocallyProposedValue`

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+The `ProposedValue` inputs received from the application are typically combined
+with the `Proposal` consensus message received by the process to produce the
+`Proposal` input that is processed by Tendermint's state machine.
+This operation is also discussed in [ADR 003][adr-003].
+In the same way as for the `LocallyProposedValue` input, the application is
+supposed to be deterministic and consistent, replaying the same inputs when the
+process recovers.
+But since the reception of these inputs typically lead to state transitions and
+outputs, it is safer to just store the value returned by the application, which
+it is supposed to be small, together with its application-evaluated validity.
```
nenadmilosevic95:
```suggestion
is supposed to be small, together with its application-evaluated validity.
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+2. OR height `H` was previously started, the WAL contains inputs for height `H`
+   that have to be replayed, since this is a recovery;
+
+Case 1. is the ordinary case, with no crashes or restarts involved.
+The process can just **reset** the WAL to height `H`.
+Namely, to remove all inputs possibly present in the WAL,
+that by design must refer to previous heights,
+and set up the WAL to log inputs of height `H`.
+
+Case 2. refers to when the process is restarting or recovering,
```
nenadmilosevic95: `Restarting or recovering` here refers to the same thing?

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+**before** they are applied to the state machine.
+A more precise requirement establishes the relation of persistence of inputs
+and production of outputs: all inputs that have lead to the production
+of an output, as a result of a state machine transition, must be persisted
+before **the output is emitted** (to the "external word").
+The reason is the adopted definition of consistency, which is derived from the
+outputs produced by a process during regular execution and recovery.
+
+Although seemingly complex, there is very simple definition of
+"all inputs that have lead to the production of an output":
```
nenadmilosevic95:
```suggestion
"all inputs that have led to the production of an output":
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+every processed input preceding the production of the output.
```
nenadmilosevic95, line 260: I don't get this sentence

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+This definition enables the following operational design for the WAL component:
+
+1. Log all processed inputs, in the order they have been processed, without
+   persistence guarantees -> asynchronous writes;
+2. When an output is produced, and before emitting it, persist all inputs that
+   were not yet persisted -> synchronous writes or `flush()`.
```
nenadmilosevic95, line 266: At first when I read it I liked this a lot and it seems very intuitive to me, but then I started thinking if there is a way this can be exploited. Do we discuss anywhere what are these outputs that trigger flush ?

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+effort manner.
+While the production of an output demands a synchronous, blocking call to
+persist all the outstanding inputs.
+
+### Replay
+
+All previous discussion boils down to this point: how is the operation of a
+recovery process?
+A first and relevant consideration is that Malachite, a priori, does not
+know if it either in ordinary or recovery operation.
```
nenadmilosevic95:
```suggestion
A key initial consideration is that, a priori, Malachite does not know whether it is operating in ordinary or recovery mode.
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+from the application indicating the height number `H` to start.
+At this point Malachite should open and load the WAL contents to check if it
+includes entries (inputs) belonging to height `H`:
+if there are, it is in recovery mode; otherwise, in regular operation.
+If the concept of [Checkpoints](#checkpoints) is adopted, this verification is
+even simpler and already described in the associated section (Case 2).
+
+When the application requests Malachite to start height `H` via `StartHeight`
+input, the consensus [driver](./adr-001-architecture.md#consensus-driver) is
+configured with its initial state, set for height `H` with some parameters
```
nenadmilosevic95, line 289: what does `set` refer to here?

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+
+There is also an important corner case to be considered.
+Crashes can occur at any time, so in particular they can occur when an output
+was produced but not yet emitted.
+Assume that WAL replay is implemented in a way that outputs derived from
+replayed inputs are produced but not emitted.
+So there is a case where the process transitions to a particular state (say,
+the `precommit` step of a round) but no process sees the output produced by
+that state transition (in the case, a `Precommit` message for that round)
+because it is not emitted during recover.
```
nenadmilosevic95:
```suggestion
because it is not emitted during recovery.
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+
+In practical terms, however, the question is whether is acceptable, during recovery,
```
nenadmilosevic95:
```suggestion
In practical terms, however, the question is whether it is acceptable, during recovery,
```

ancazamfir (approved):

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+as discussed in the [Inputs](#inputs-1) section below.
+
+The Write-Ahead Log (WAL) is implemented as an
+[Actor](./adr-002-node-actor.md#write-ahead-log-(wal)-actor)
+and runs in its own thread, as part of the
+[malachitebft-engine][engine-crate] crate.
+The interaction between the consensus engine and the WAL routine uses a set of
+messages (`enum WalMsg`), the most relevant being `Append` and `Flush`, `Reset`
+and `StartedHeight`.
+The first messages two are related to the [persistence](#persistence-1) of
```
nenadmilosevic95:
```suggestion
The first two messages are related to the [persistence](#persistence-1) of
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+outputs.
+However, the adopted definition of consistency (see [Context](#context))
+requires the logging of single inputs.
+For instance, if a process only receives `2f` matching `Precommit` messages,
+there is no complex input to process or persist, nor produced output.
+But the reception of an additional matching `Precommit` message leads to a
+complex input and an observable state transition.
+If the first `2f Precommit` messages are not persisted to the WAL, a recovering
+process will not produce the same state transition when receiving the missing
+`Precommit`.
```
nenadmilosevic95, line 379: But if we don't flush this unless we have the output what is the difference here?

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+- `TimeoutElapsed` is directly mapped to `WalEntry::Timeout`
+
+So, all the [list of inputs](#inputs) that must be persisted were contemplated.
+There are, however, important exceptions: some inputs produced by the
+synchronization protocol (`SyncValueResponse`, which includes a `CommitCertificate`)
+and by the liveness protocol (`PolkaCertificate` and `RoundCertificate`) are
+not persisted.
+They are all **Certificate**s, namely types that aggregate multiple `Vote` inputs.
+By not persisting those inputs - namely, the `Vote` inputs not already persisted -
+recovering processes may behave inconsistently (see [issue #1445][issue-certs]).
```
nenadmilosevic95, line 433: Hmm, nice catch, we should consider and discuss this!

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+
+To conclude the list of `Input`s, `StartHeight` is not persisted to the WAL but
+this input leads to either:
```
nenadmilosevic95, line 436: This is actually interesting, because `StartHeight` is issued from the app, so the app is responsible to restart the same height?

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+
+### Replay
+
+The replay of the WAL is triggered by the reception of `StartHeight` input from
+the application for a height matching the WAL's height.
+The `fetch_entries` method is used to iterate through the WAL and collect all stored
+`WalEntry` instances, that are returned to the consensus engine.
+If `WalEntry` instances are returned, the consensus engine is set to
+`Phase::Recovering` and the persisted inputs are replayed by the `wal_replay` method.
+This method reconstructs the associated consensus `Input`s and apply then using the
```
nenadmilosevic95:
```suggestion
This method reconstructs the associated consensus `Input`s and apply them using the
```

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+`process_input` method, which is the same used to process ordinary inputs.
+The `Phase::Recovering` flag is actually only used to block the `WalAppend`
+effect from appending again to the WAL inputs that are being replayed,
+as long as blocking calls to the associated WAL's `flush()` method.
```
nenadmilosevic95: I don't understand this sentence?

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+
+As a result, replayed inputs produce outputs, `Effect`s in the consensus engine
+parlance, in the same way as ordinary inputs, the exception being only the
+effects related to persisting inputs to the WAL.
+In any case, the existence of the `Phase::Recovering` flag allows filtering out
+behaviors, effects, an inputs that are not needed or redundant
+in recovery mode - although it should be used very carefully.
+Once replaying is done, the `Phase::Recovering` flag is cleared and processed
+inputs are appended to the existing WAL, which is not [reset](#reset-1) in this case.
```
nenadmilosevic95: I would polish this part a bit

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+
+1. When a new round of consensus is started, via the `StartRound` effect;
+2. When a consensus message is broadcast, via the `PublishConsensusMsg` effect;
+3. When a value is finalized by the consensus, via the `Decide` effect.
+
+Notice that synchronous writes are performed whenever the effect is
+externalized, either to the application (1 and 3) or the network (2).
+
+A current limitation of the persistence approach, however, is the fact that
+calls to `wal_append` are also blocking, while they do not have to.
```
nenadmilosevic95, line 506: This means that we at the moment we don't do asynchronous writes?

docs/architecture/adr-007-write-ahead-log.md
```diff
@@ -0,0 +1,576 @@
...
+
+The Write-Ahead Log is a crucial component for the operation of a consensus
+process.
+As a result, errors when attempting to append inputs to the WAL
+**must be critical**.
+At the moment, from this [commit](https://github.com/circlefin/malachite/commit/38f113f6c81da0af32a748718b2d87ab64e3a72f),
+consensus hangs forever in case of any WAL operational error.
+
+A second source of errors is when replaying the WAL, where one particular error
+requires attention: the corruption of the last entries of the WAL.
```
nenadmilosevic95:
```suggestion
A second source of errors occurs during WAL replay and requires special attention: an error at a given entry implies that all subsequent WAL entries are corrupted.
```

docs/architecture/README.md
```diff
@@ -30,4 +30,5 @@ To suggest an ADR, please make use of the [ADR template](./adr-template.md) prov
 | [001](./adr-001-architecture.md)            | High Level Architecture                     | Accepted |
 | [002](./adr-002-node-actor.md)              | Node Architecture using the Actor Model     | Accepted |
 | [003](./adr-003-values-propagation.md)      | Propagation of Proposed Values              | Accepted |
-| [004](./adr-004-coroutine-effect-system.md) | Coroutine-Based Effect System for Consensus | Accepted |
\ No newline at end of file
+| [004](./adr-004-coroutine-effect-system.md) | Coroutine-Based Effect System for Consensus | Accepted |
+| [007](./adr-007-write-ahead-log.md)         | Consensus Write-Ahead Log (WAL)             | Accepted |
```
nenadmilosevic95, line 36: Thanks for changing to 007, I wanted to let you use 006 since you are older :)

nenadmilosevic95 (commented):
Wonderfull ADR! Great job! Thanks @cason, very well written and extremely useful!!! I left some comments!

docs/architecture/adr-007-write-ahead-log.md
cason: I will change the verb, it intermediates?

docs/architecture/adr-007-write-ahead-log.md
cason: In Comet we scheduled the timeouts again and wait for their expiration. So we could have messages arriving either before or after the timeout expiration, leading to potential equivocation,

docs/architecture/adr-007-write-ahead-log.md
cason: Yes... good point.

docs/architecture/adr-007-write-ahead-log.md
cason: Every input processed before the production of the output. Namely, everything that happens until the output is produced.

docs/architecture/adr-007-write-ahead-log.md
cason: Yes, we do in the decision part.

docs/architecture/adr-007-write-ahead-log.md
cason: The driver is set to height `H`.

docs/architecture/adr-007-write-ahead-log.md
cason: We don't flush, but the write opportunistically flushes.

docs/architecture/adr-007-write-ahead-log.md
cason: We do, but they are blocking.

docs/architecture/adr-007-write-ahead-log.md
nenadmilosevic95: That is better!

docs/architecture/adr-007-write-ahead-log.md
cason: Yes, it is.

docs/architecture/adr-007-write-ahead-log.md
cason: 3771ad8c65a911113d2a572b49de8ecd97d5f41d

docs/architecture/adr-007-write-ahead-log.md
cason: Yes, there are multiple cases, I am just considering the ignoring lock misbehaviour as an example. This block was not in the initial version, then from some comments I realized that the rationale of replaying was not straightforward as I thought. So I am giving here an example.

nenadmilosevic95: I would then just write it slightly different, something like "misbehave in different ways:", so it doesn't seem there are only two ways. Anyway, this is just nitpicking so please ignore if not needed

cason: https://github.com/circlefin/malachite/pull/1408/commits/9287ce25b9d10d280bd5f81e454b7a71e6ecf676

romac (dismissed):

Issue #1376, opened 2026-01-25
WAL Actor Deadlock on Append Errors


vikions:
## Summary
`WalMsg::Append` may exit without sending a reply to the caller, which can cause `ractor::call!` to block indefinitely and potentially deadlock the consensus task.

## Problem
In `wal/thread.rs`, the `WalMsg::Append` handler does not send a reply when:
- `encode_entry()` fails (the error is propagated via `?`)
- the encoded buffer is empty

## Code reference
```rust
WalMsg::Append(entry, reply) => {
    let mut buf = Vec::new();
    encode_entry(&entry, codec, &mut buf)?; // ← no reply on error

    if !buf.is_empty() {
        // reply only sent here
    }
    // ← no reply if buffer is empty
}
```

## Impact
Callers using `ractor::call!(self.wal, WalMsg::Append, ...)` will wait indefinitely for a response, which can stall or deadlock the consensus actor and effectively freeze the node.

## Reproduction idea
Inject a codec error (for example, make `encode_entry` return `Err` for a specific entry) and trigger a WAL append. The caller will block forever waiting for the WAL actor reply.

## Suggested fix
One possible fix is to always send a reply, even when encoding fails or the buffer is empty, by capturing the result and replying unconditionally:

```rust
let result = encode_entry(&entry, codec, &mut buf)
    .and_then(|_| {
        if !buf.is_empty() {
            log.append(&buf).map_err(Into::into)
        } else {
            Ok(())
        }
    });

if reply.send(result).is_err() {
    error!("Failed to send WAL append reply");
}
```

If this approach aligns with the intended semantics for empty entries, I’d be happy to submit a PR.

romac:
Good catch, that's indeed not the intended behavior. The fix you suggest looks good to me, so feel free to open a PR. Thanks a lot for reporting this and offering to help fix it 🙏

PR #1410, opened 2026-01-28
Fix WAL actor deadlock on Append errors

vikions:
Closes: #1376

This PR addresses the deadlock issue in the WAL actor's `Append` handler by ensuring a reply is always sent to the caller, even when encoding fails or produces an empty buffer.

## Changes
- Modified `WalMsg::Append` handler to capture the result of `encode_entry` and unconditionally send a reply
- Empty buffers now return `Ok(())` instead of silently skipping the reply
- Preserved existing error logging behavior

## Testing
The fix prevents `ractor::call!` from blocking indefinitely in scenarios where:
1. `encode_entry()` returns an error
2. The encoded buffer is empty

### DIFF
--- a/code/crates/engine/src/wal/thread.rs
+++ b/code/crates/engine/src/wal/thread.rs
@@ -113,23 +113,27 @@ where
             let entry_type = wal_entry_type(&entry);
 
             let mut buf = Vec::new();
-            encode_entry(&entry, codec, &mut buf)?;
-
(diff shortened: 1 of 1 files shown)


cason:
Is that duplicated?

vikions:
> Is that duplicated?


Yes, here is the already adopted PR https://github.com/circlefin/malachite/pull/1411

PR #1414, opened 2026-01-29
refactor(sync): Sync actor notifies Consensus of sync responses

romac:
> [!NOTE]
> This change lead to the discovery of the bug fixed by #1413. It therefore needs #1413 to be merged for tests to pass.

Change the Sync actor to notify the Consensus actor when there is a sync response to process instead of having the Consensus actor directly pick them from the network events, leaving the responsibility of which responses to process or not to the Sync actor.

To enable this, this PR replaces the direct `SyncRef` dependency in Consensus with an `OutputPort` that allows late binding of the Sync actor subscription. This enables spawning the Consensus actor before the Sync actor while maintaining communication between them.

Changes:
- Consensus now sends sync messages through an `OutputPort` instead of directly casting to `SyncRef`
- Sync actor receives a `ConsensusRef` to forward value responses
- Move sync response processing from network event handler to dedicated `Msg::ProcessSyncResponse` variant
- Reorder actor spawning: `Consensus` before `Sync`, with port subscription after both are initialized

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -43,6 +43,7 @@
 - Remove signing of proposal parts
 
 ### `sync`
+- Fix partial range request not being tracked in pending requests
 - Initial random (fixed) period adjustment in sync status ticker
 - Refactor sync actor to notify consensus of sync responses
 - Support batch retrieval of decided values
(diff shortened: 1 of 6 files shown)


ancazamfir (approved):
👍

PR #1418, opened 2026-01-29
feat(CCHAIN-485): Surface proposal and vote equivocation evidence

sergio-mena:
Closes: #911
Continues: #999 (thanks @bastienfaivre for the base PR!)

I'm opening a new PR because we are not pushing PR branches to `circlefin`.

This PR is surfacing to the application the vote and proposal equivocation evidence detected by Malachite Core.
It is surfaced via events (immediately), and also in the `AppMsg::Decided` (and `HostMsg::Decided`) message.
 * N.B.: I'm not sure we want to keep the event-based logic, as the one in `Decided` is good enough for our purpose

An integration test that makes to validators share the same key, makes sure that the equivocation is surfaced. This was the simplest way I found to it-test the new infra. But I think further unit tests could be written later on.

---

**Story:** <https://circlepay.atlassian.net/browse/CCHAIN-485>

**Requested Reviewers:** @hvanz @alesforz 

---

#### Contribution eligibility

#### For all contributors

### DIFF
--- a/code/crates/test/tests/it/equivocation.rs
+++ b/code/crates/test/tests/it/equivocation.rs
@@ -58,7 +58,7 @@ pub async fn equivocation_two_vals_same_pk() {
     test.add_node()
         .start()
         .on_vote(|_v, _s| Ok(HandlerResult::SleepAndContinueTest(VOTE_DURATION)))
-        .on_decided(move |_c, evidence, _s| {
+        .on_decided(|_c, evidence, _s| {
             check_decided_impl(&evidence);
             let result = if evidence.proposals.is_empty() {
                 HandlerResult::WaitForNextEvent
@@ -73,7 +73,7 @@ pub async fn equivocation_two_vals_same_pk() {
     test.add_node()
         .start()
         .on_vote(|_v, _s| Ok(HandlerResult::SleepAndContinueTest(VOTE_DURATION)))
-        .on_decided(move |_c, evidence, _s| {
+        .on_decided(|_c, evidence, _s| {
             check_decided_impl(&evidence);
             let result = if evidence.votes.is_empty() {
                 HandlerResult::WaitForNextEvent


code/crates/engine/src/util/events.rs
```diff
@@ -62,6 +66,16 @@ pub enum Event<Ctx: Context> {
...
+    ProposalEquivocationEvidence {
+        proposal_height: Ctx::Height,
+        address: Ctx::Address,
+        evidence: (SignedProposal<Ctx>, SignedProposal<Ctx>),
+    },
+    VoteEquivocationEvidence {
+        vote_height: Ctx::Height,
+        address: Ctx::Address,
+        evidence: (SignedVote<Ctx>, SignedVote<Ctx>),
+    },
```
sergio-mena: Are we sure we want to keep these two events?

code/crates/core-votekeeper/src/evidence.rs
```diff
@@ -14,7 +14,8 @@ where
     Ctx: Context,
 {
     #[allow(clippy::type_complexity)]
```
romac: This can likely be removed

```suggestion
```

code/crates/app-channel/src/run.rs
```diff
@@ -25,14 +25,19 @@ use crate::msgs::{ConsensusRequest, NetworkRequest};
 use crate::spawn::{spawn_host_actor, spawn_network_actor};
 use crate::Channels;
 
-pub struct EngineHandle {
+pub struct EngineHandle<Ctx: Context> {
     pub actor: NodeRef,
+    pub consensus: ConsensusRef<Ctx>,
```
romac: Was this field added because it's needed by the Arc app?

cason:
> An integration test that makes to validators share the same key, makes sure that the equivocation is surfaced. This was the simplest way I found to it-test the new infra. But I think further unit tests could be written later on.

It is failing, pay attention with those tests for equivocation, they were super flaky in Comet.

code/crates/core-types/src/lib.rs
```diff
@@ -47,12 +47,18 @@ pub type PrivateKey<Ctx> = <<Ctx as Context>::SigningScheme as SigningScheme>::P
 /// Type alias to make it easier to refer the `Signature` type.
 pub type Signature<Ctx> = <<Ctx as Context>::SigningScheme as SigningScheme>::Signature;
 
-/// A signed vote
+/// A signed vote.
 pub type SignedVote<Ctx> = SignedMessage<Ctx, <Ctx as Context>::Vote>;
 
+/// A double signed vote (e.g., conflicting votes).
+pub type DoubleVote<Ctx> = (SignedVote<Ctx>, SignedVote<Ctx>);
```
cason, line 54: DoubleVotes?

code/crates/core-votekeeper/src/evidence.rs
```diff
@@ -32,19 +33,50 @@ where
...
+                address == vote.validator_address() && conflicting == vote
+            })
+            .cloned()
+    }
+
+    /// Add evidence of equivocating votes, ie. two votes submitted by the same validator,
+    /// but with different values but for the same height and round.
+    ///
+    /// # Precondition
+    /// - Panics if the two conflicting votes were not proposed by the same validator.
```
cason:
```suggestion
    /// - Panics if the two conflicting votes are not from the same validator.
```

ARCHITECTURE.md
```diff
@@ -182,10 +185,32 @@ it only constraints that any implementation satisfies the `Ctx::Proposal` trait.
...
-https://github.com/circlefin/malachite/blob/53a9d9e071e773ff959465f2836648d8ad2a5c74/code/crates/core-types/src/proposal.rs#L5-L28
+<https://github.com/circlefin/malachite/blob/53a9d9e071e773ff959465f2836648d8ad2a5c74/code/crates/core-types/src/proposal.rs#L5-L28>
 
 Inputs, Effects, and the Context are the three key types that make up the Malachite consensus library.
 
+### Equivocation Evidence
+
+Malachite detects and surfaces validator equivocation in two ways:
+
+- Immediate events: When a conflicting proposal or vote is received, the engine emits `Event::ProposalEquivocationEvidence` or `Event::VoteEquivocationEvidence` with the full pair of signed artifacts.
```
cason:
```suggestion
- Immediate events: When a conflicting proposal or vote is received, the engine emits `Event::ProposalEquivocationEvidence` or `Event::VoteEquivocationEvidence` with the full pair of signed messages.
```

ARCHITECTURE.md
```diff
@@ -182,10 +185,32 @@ it only constraints that any implementation satisfies the `Ctx::Proposal` trait.
...
+  These are useful for real-time telemetry and alerting.
+- Decide-time evidence: On `Effect::Decide`, Malachite includes a `MisbehaviorEvidence` payload that aggregates all proposal and vote equivocations observed during the height.
```
cason:
```suggestion
- Decide-time evidence: On `Effect::Decide`, Malachite includes a `MisbehaviorEvidence` that aggregates all equivocating proposals and votes observed in the decided height.
```

cason: So, to be clear: during height `H` we produce only evidence/misbehaviour output regarding height `H`, correct? So it is not about "when" the evidence was observed, but "to which" height it belongs.

cason: Also, it is not that if an equivocated message is received after the decision we are reporting anyway, is it?

code/crates/core-consensus/src/handle/decide.rs
```diff
@@ -81,9 +82,14 @@ where
         }
     }
 
+    let evidence = MisbehaviorEvidence {
+        proposals: state.driver.proposal_evidence().to_owned(),
```
cason: `to_owned()` is like a `clone()` with asteroids?

code/crates/core-votekeeper/src/keeper.rs
```diff
@@ -257,8 +257,12 @@ where
                     "Received equivocating vote {:?}, existing {:?}",
                     conflicting, existing
                 );
-                self.evidence.add(existing, conflicting);
-                return None;
+                self.evidence.add(existing.clone(), conflicting.clone());
+
+                return Err(RecordVoteError::ConflictingVote {
```
cason: But we actually ignore this error, don't we?

code/crates/test/tests/it/equivocation.rs
```diff
@@ -0,0 +1,69 @@
...
+    test.add_node().start().success();
+
+    // Node 2 (same validator key as node 1)
+    test.add_node().start().success();
+
+    // Node 3 -- checking proposal equivocation evidence
+    test.add_node()
+        .start()
+        .on_proposal_equivocation_evidence(|_height, _address, (p1, p2), _state| {
+            assert_ne!(p1.message.value.value, p2.message.value.value);
```
cason: The test is failing, I don't know why. You should add some error message to all those asserts.

code/crates/engine/src/util/events.rs
```diff
@@ -82,7 +96,16 @@ impl<Ctx: Context> fmt::Display for Event<Ctx> {
...
                 )
             }
-            Event::Decided(cert) => write!(f, "Decided(value: {})", cert.value_id),
+            Event::Decided {
+                commit_certificate,
+                evidence,
+            } => {
+                write!(
+                    f,
+                    "Decided(value: {}, evidence: {:?})",
```
cason: Can we only write the evidence part when some evidence is produced? This would help a log logging and debugging.

code/crates/test/tests/it/equivocation.rs
```diff
@@ -0,0 +1,69 @@
...
+    };
+    let mut test = TestBuilder::<()>::new();
+
+    // Node 1
+    test.add_node().start().success();
+
+    // Node 2 (same validator key as node 1)
+    test.add_node().start().success();
+
+    // Node 3 -- checking proposal equivocation evidence
```
cason, line 57: Both nodes 3 and 4 look for both evidences. Can this be the reason for the test failure?

code/crates/app-channel/src/run.rs
sergio-mena: Good catch!
This was on the base patch, and I didn't realize we don't need it in the current version. I just removed it

code/crates/core-consensus/src/handle/decide.rs
sergio-mena: When having a reference, I prefer `to_owned()` because it makes the code more readable IMO

code/crates/core-types/src/lib.rs
sergio-mena: No, it is one double vote

code/crates/core-votekeeper/src/keeper.rs
sergio-mena: Yes, I think you're right.
This has to do with not being sure of the approach we want to follow. But, yeah, as it is now, and if nobody else objects to it, your comment fully applies. Let me fix it

code/crates/engine/src/util/events.rs
sergio-mena: @romac @ancazamfir

sergio-mena:
> Can we have a version where this is no instantaneous report of detected evidence, for neither application and the events handler?

This is what I went for (out of the options that were discussed in the original PR #999).
Also, check [this comment](https://github.com/circlefin/malachite/pull/1418#discussion_r2742543867): the current events are not used by the app, they're only there for observability. The app won't use them, it will use the info in `Decide`.

code/crates/core-driver/src/driver.rs
```diff
@@ -212,6 +212,16 @@ where
...
     }
 
+    /// Return recorded evidence of proposal equivocation for this height.
+    pub fn proposal_evidence(&self) -> &proposal_keeper::EvidenceMap<Ctx> {
+        self.proposal_keeper.evidence()
+    }
+
+    /// Return recorded evidence of vote equivocation for this height.
+    pub fn vote_evidence(&self) -> &malachitebft_core_votekeeper::EvidenceMap<Ctx> {
+        self.votes().evidence()
```
romac: Nitpick: for consistency with the method above:

```suggestion
        self.vote_keeper.evidence()
```

code/crates/core-driver/src/proposal_keeper.rs
```diff
@@ -254,7 +265,8 @@ where
     Ctx: Context,
 {
     #[allow(clippy::type_complexity)]
```
romac: This one can likely be removed as well

ARCHITECTURE.md
sergio-mena: This is the MVP we agreed on:
* only equivocation from the current height are returned with the `Decide` message
* no persistence (at malachite core level) for crash-recovery
* no gossiping

code/crates/engine/src/util/events.rs
romac: imho having the evidence in `Decided` is enough, no need for new events. unless there are tests that we can only write with those events, eg. a test where no decision can be made because a node equivocates, but I am not sure how useful a test that would be. wdyt?

code/crates/test/app/src/node.rs
```diff
@@ -67,6 +67,15 @@ pub struct App {
     pub middleware: Option<Arc<dyn Middleware>>,
 }
 
+impl App {
+    fn get_nw_keypair(&self) -> Keypair {
```
romac: Wasn't sure if this was a typo for `new` or an abbreviation for `network`, which  means that it's probably worth expanding to:

```suggestion
    fn get_network_keypair(&self) -> Keypair {
```

code/crates/engine/src/util/events.rs
sergio-mena: Sure! Good idea

code/crates/test/tests/it/equivocation.rs
sergio-mena: I structured the test so that node 3 _only_ checks for proposal evidence, and node 4 _only_ check for vote evidence.
This way, if either vote or proposal evidences are not being produced, the test will fail.

But two nodes sharing the same key, should cause both vote and proposal evidence, shouldn't they?

code/crates/app-channel/src/connector.rs
```diff
@@ -196,6 +196,7 @@ where
             HostMsg::Decided {
                 certificate,
                 extensions,
+                evidence,
```
ancazamfir, line 199: Once we have `Finalized` would that be a better place?

code/crates/core-driver/src/proposal_keeper.rs
```diff
@@ -226,11 +228,15 @@ where
     }
 
     /// Store a proposal, checking for conflicts and storing evidence of equivocation if necessary.
-    pub fn store_proposal(&mut self, proposal: SignedProposal<Ctx>, validity: Validity) {
+    pub fn store_proposal(
+        &mut self,
+        proposal: SignedProposal<Ctx>,
+        validity: Validity,
+    ) -> Result<(), RecordProposalError<Ctx>> {
```
ancazamfir: why the change? We store the proposal and might store equivocation. Also the error is currently dropped.

code/crates/core-driver/src/proposal_keeper.rs
```diff
@@ -272,23 +284,42 @@ where
...
+        proposal: &SignedProposal<Ctx>,
+    ) -> Option<(Ctx::Address, DoubleProposal<Ctx>)> {
+        self.last
+            .as_ref()
+            .filter(|(address, (_, conflicting))| {
+                address == proposal.validator_address() && conflicting == proposal
+            })
+            .cloned()
+    }
+
```
ancazamfir: it's really annoying that we have to add this just for testing.

code/crates/core-votekeeper/src/evidence.rs
ancazamfir: Just to clarify what the precondition is
```suggestion
    /// Both votes must be from the same validator (debug-asserted).
```
And we should have the same in proposal_keeper when we add the proposal.

code/crates/test/tests/it/equivocation.rs
sergio-mena: Let me investigate

code/crates/engine/src/util/events.rs
ancazamfir: yes, they are currently used for tests. Maybe we can run the test and just check Decided has evidence.

code/crates/core-driver/src/proposal_keeper.rs
sergio-mena: Removed

code/crates/app-channel/src/connector.rs
sergio-mena: :-D
I'm looking at each with one half of my brain. Hadn't thought about combining them.

Thinking of it, it see two options:
* Leave if here (if the target height time is not set, the app doesn't get `Finalized`)
* Add it on both (`Decided` and `Finalized`) to allow for notifying misbehavior detected during finalization

IMO, the second is overkill

code/crates/core-driver/src/proposal_keeper.rs
sergio-mena: Yeah. I just reverted this to address one of @cason 's comments

code/crates/core-driver/src/proposal_keeper.rs
sergio-mena: It's not just for testing.
My understanding is that this is an optimization so that the engine doesn't need to go through all evidences stored, to see if one got just added.

Having said that, this was in the base patch from May, so I might have misunderstood.

code/crates/core-votekeeper/src/evidence.rs
sergio-mena: Yep. Done

code/crates/engine/src/util/events.rs
sergio-mena: Cool! Thanks for takin a look. I agree with you.
Let me simplify the patch.

code/crates/test/tests/it/equivocation.rs
sergio-mena: I added more logs (check tip of this branch). Trying to see if I can see what's going on

code/crates/core-driver/src/proposal_keeper.rs
ancazamfir: it's just for the testing framework, i removed them and cleaned the code that uses them. Changed the test to:
```
#[tokio::test]
pub async fn equivocation_two_vals_same_pk() {
    // Nodes 1 and 2 share a validator key to induce equivocation
    let params = TestParams {
        shared_key_group: HashSet::from([1, 2]),
        ..Default::default()
    };

    // State: count of decide events seen
    let mut test = TestBuilder::<u32>::new();

    // Node 0
    test.add_node().start().success();

    // Node 1 (shares validator key with node 2)
    test.add_node().start().success();

    // Node 2 (shares validator key with node 1)
    test.add_node().start().success();

    // Node 3 checking equivocation evidence at decide time
    test.add_node()
        .start()
        .on_decided(|_certificate, evidence, decide_count| {
            *decide_count += 1;

            let has_proposal_evidence = !evidence.proposals.is_empty();
            let has_vote_evidence = !evidence.votes.is_empty();

            // Pass test once we have both proposal and vote equivocation evidence
            if has_proposal_evidence && has_vote_evidence {
                Ok(HandlerResult::ContinueTest)
            } else if *decide_count > 3 {
                // Evidence should appear in first heights, fail after 3 decides
                bail!(
                    "Decided {} times. proposal_evidence={}, vote_evidence={}",
                    decide_count,
                    has_proposal_evidence,
                    has_vote_evidence
                )
            } else {
                Ok(HandlerResult::WaitForNextEvent)
            }
        })
        .success();

    test.build()
        .run_with_params(Duration::from_secs(30), params)
        .await;
}
```

code/crates/core-driver/src/proposal_keeper.rs
ancazamfir: Ah, i see @cason comment for the vote keeper. Do you have a commit? I don't see the revert for the proposal_keeper.

code/crates/core-driver/src/proposal_keeper.rs
ancazamfir: in case they are helpful, just pushed the changes in https://github.com/ancazamfir/malachite/tree/no-extra-events

code/crates/engine/src/util/events.rs
sergio-mena: > yes, they are currently used for tests

That can be fixed easily. We can do the asserts at decide time, bear with me

ARCHITECTURE.md
```diff
@@ -182,10 +185,32 @@ it only constraints that any implementation satisfies the `Ctx::Proposal` trait.
...
-https://github.com/circlefin/malachite/blob/53a9d9e071e773ff959465f2836648d8ad2a5c74/code/crates/core-types/src/proposal.rs#L5-L28
+<https://github.com/circlefin/malachite/blob/53a9d9e071e773ff959465f2836648d8ad2a5c74/code/crates/core-types/src/proposal.rs#L5-L28>
 
 Inputs, Effects, and the Context are the three key types that make up the Malachite consensus library.
 
+### Equivocation Evidence
+
+Malachite detects and surfaces validator equivocation in two ways:
+
+- Immediate events: When a conflicting proposal or vote is received, the engine emits `Event::ProposalEquivocationEvidence` or `Event::VoteEquivocationEvidence` with the full pair of signed messages.
```
cason: Are we still producing those events? Where?

code/crates/engine/src/util/events.rs
```diff
@@ -62,6 +66,16 @@ pub enum Event<Ctx: Context> {
     WalReplayError(Arc<ConsensusError<Ctx>>),
     WalResetError(Arc<eyre::Report>),
     WalCorrupted(Arc<io::Error>),
+    ProposalEquivocationEvidence {
```
cason: Are we still producing those events? Where?

code/crates/engine/src/util/events.rs
```diff
@@ -103,6 +130,20 @@ impl<Ctx: Context> fmt::Display for Event<Ctx> {
             Event::SkipRoundCertificate(certificate) => {
                 write!(f, "SkipRoundCertificate: {certificate:?})")
             }
+            Event::ProposalEquivocationEvidence {
```
cason: Are we still producing those events? Where?

code/crates/core-driver/src/proposal_keeper.rs
cason: Are we using this Error for anything?

code/crates/core-driver/src/proposal_keeper.rs
cason: > My understanding is that this is an optimization so that the engine doesn't need to go through all evidences stored, to see if one got just added.

It is not easier to just drain all stored ones when reporting?

cason (commented):
It is not clear to me whether this is an intermediate status towards removing the single evidence report events, that is the way I would prefer. If so, we can remove all the helpers and types supporting those events.

code/crates/test/tests/it/equivocation.rs
sergio-mena: I think I managed to fix it. Commit coming up

code/crates/core-driver/src/proposal_keeper.rs
sergio-mena: Removed it (commit upcoming)

code/crates/engine/src/util/events.rs
sergio-mena: I removed them

code/crates/engine/src/util/events.rs
sergio-mena: Done

code/crates/core-driver/src/proposal_keeper.rs
sergio-mena: `Consensus::handle_msg` uses both `is_last_equivocation` (vote and proposal) for the metrics. Do you see another way to deal with the metrics?

sergio-mena: I'm trying something. Bear with me

sergio-mena: Ok, I think I got something working, please take a look

ARCHITECTURE.md
sergio-mena: Addressed

ARCHITECTURE.md
```diff
@@ -182,10 +185,28 @@ it only constraints that any implementation satisfies the `Ctx::Proposal` trait.
 Furthermore, this trait is generic over a type which we call a `Context`.
 The full definition of `Ctx::Proposal` trait looks like this:
 
-https://github.com/circlefin/malachite/blob/53a9d9e071e773ff959465f2836648d8ad2a5c74/code/crates/core-types/src/proposal.rs#L5-L28
+<https://github.com/circlefin/malachite/blob/53a9d9e071e773ff959465f2836648d8ad2a5c74/code/crates/core-types/src/proposal.rs#L5-L28>
```
sergio-mena:
```suggestion
https://github.com/circlefin/malachite/blob/53a9d9e071e773ff959465f2836648d8ad2a5c74/code/crates/core-types/src/proposal.rs#L5-L28
```

code/crates/core-driver/src/proposal_keeper.rs
```diff
@@ -227,16 +227,17 @@ where
         &self.evidence
     }
 
+    /// Remove and return all recorded evidence.
+    pub fn take_evidence(&mut self) -> EvidenceMap<Ctx> {
+        core::mem::take(&mut self.evidence)
```
cason, line 232: Unsafe 😈

code/crates/app-channel/src/run.rs
```diff
@@ -161,7 +161,7 @@ where
     .await?;
 
     let (tx_request, rx_request) = mpsc::channel(request_ctx.channel_size);
-    spawn_consensus_request_task(rx_request, consensus);
+    spawn_consensus_request_task(rx_request, consensus.clone());
```
cason: Why? You need to rebase this branch with `main`, there are more changes in this file.

code/crates/core-driver/src/driver.rs
```diff
@@ -212,6 +212,26 @@ where
         self.proposer.as_ref()
     }
 
+    /// Return recorded evidence of proposal equivocation for this height.
```
cason: I don't like much "for this height". The driver is a per-height abstraction, in any case. And we don't ensure that recorder evidence is for the current height, do we?

cason (approved):
Thank you for that. Maybe we should save somewhere the previous version, producing "live" misbehaviour events, but I think that at the moment surfacing them upon decisions is more than enough.

sergio-mena:
> Thank you for that. Maybe we should save somewhere the previous version, producing "live" misbehaviour events

Yes, that's #999

code/crates/core-driver/src/proposal_keeper.rs
sergio-mena: Why?

code/crates/app-channel/src/run.rs
sergio-mena: OK, let me do that...

code/crates/core-driver/src/driver.rs
sergio-mena: I don't think so. Adapted the comment

code/crates/app-channel/src/run.rs
sergio-mena: I merged from `malachite/main`, and removed the `.clone()`

code/crates/core-driver/src/driver.rs
```diff
@@ -212,6 +212,26 @@ where
         self.proposer.as_ref()
     }
 
+    /// Return recorded evidence of proposal equivocation.
+    pub fn proposal_evidence(&self) -> &proposal_keeper::EvidenceMap<Ctx> {
+        self.proposal_keeper.evidence()
+    }
```
ancazamfir: is this still needed?

code/crates/core-driver/src/driver.rs
```diff
@@ -212,6 +212,26 @@ where
...
+
+    /// Remove and return recorded evidence of proposal equivocation.
+    pub fn take_proposal_evidence(&mut self) -> proposal_keeper::EvidenceMap<Ctx> {
+        self.proposal_keeper.take_evidence()
+    }
+
+    /// Return recorded evidence of vote equivocation.
+    pub fn vote_evidence(&self) -> &malachitebft_core_votekeeper::EvidenceMap<Ctx> {
+        self.vote_keeper.evidence()
+    }
```
ancazamfir: Is this still needed?

code/crates/test/tests/it/equivocation.rs
```diff
@@ -0,0 +1,85 @@
...
+            assert_eq!(v1.round(), v2.round());
+            assert_eq!(v1.vote_type(), v2.vote_type());
+            assert_ne!(v1.value(), v2.value());
+            break;
+        }
+    }
+}
+
+#[tokio::test]
+pub async fn equivocation_two_vals_same_pk() {
```
ancazamfir, line 37: Could you add a description on what this test does?

ancazamfir: Looks good by the way. I think in this setup nodes 3 and 4 will detect both proposal and vote equivocation in the heights where 1/2 is proposer but probably not important to enforce in the test.

code/crates/core-driver/src/driver.rs
sergio-mena: I would swear I saw a callsite to this function before pushing my last commit, but looks like it's not the case. Thanks for the catch!

code/crates/test/tests/it/equivocation.rs
sergio-mena: I "specialized" nodes 3 and 4 on detecting only one equivocation type because that way the test will fail if one type of evidence stops working but not the other.

sergio-mena: Description added

ancazamfir: > I "specialized" nodes 3 and 4 on detecting only one equivocation type because that way the test will fail if one type of evidence stops working but not the other.

What I meant is that, given how the test is setup, and if nodes 3 and 4 are correct they must detect both, but we only check one type (even if different per node). Anyway, this is minor.

code/crates/core-driver/src/proposal_keeper.rs
cason: https://doc.rust-lang.org/core/mem/fn.take.html - maybe I'm wrong, to ignore.

sergio-mena: We only alternative a can think of is defining the field as `Option< ... >`, and then using `Option`s `take()` method; but:
* It looks more cumbersome (the value is not optional, it's always there)
* Not sure it's more safe

Issue #1446, opened 2026-02-04
consensus: Collect late precommits for monitoring purposes


sergio-mena:
Some metrics (uptime, downtime, etc) depend on Malachite collecting -- and making available to the app -- late precommits.

Currently, as soon as enough precommits have been collected to be able to decide on a height, Malachite decides immediately, moves on to the next height, and discard any late commits for the height it just decided on.

PR #1447, opened 2026-02-04
feat(CCHAIN-890): Add finalization period for collecting additional precommits

sergio-mena:
Closes: #1446
Closes: #1466

This PR introduces a finalization period in consensus that allows collecting additional precommit votes ("late precommits") after a decision has been reached. When a configurable `target_time` is set, consensus
  waits for a specified duration to gather more precommits before moving to the next height.

  This produces extended commit certificates with more signatures, which can be used for monitoring purposes.

  The finalization period is optional and controlled by the `target_time` parameter in `HeightParams`. When not set, consensus will skip the finalization period (just as before), maintaining backward compatibility.

  Changes:

  - Added `finalization_period` boolean flag to consensus state to track when collecting late precommits
  - Consensus always emits both `Decided` and `Finalized` effects in all cases
    - `Decided` is emitted without a reply channel (notification only)
    - `Finalized` is emitted with a reply channel (for application response)
  - If `target_time` is configured, consensus enters finalization period after deciding
    - Sets `finalization_period` flag to true
    - Schedules timeout for remaining duration until target time
    - During finalization, continues accepting and aggregating precommit votes while in Commit step
    - When timeout expires, emits `Finalized` effect and clears flag `finalization_period`
  - Otherwise, emits `Finalized` effect immediately after `Decided`
  - IMPORTANT: if the current height is deciding because we received a commit certificate via sync, then
    - we emit `Finalized` effect immediately after `Decided`
    - we don't query the vote keeper, but use the commit certificate we used via sync
    - in this case, we won't be collecting further precommits in the finalization period, but we don't need them anyway because the node was late, so Data Monitoring will filter out this node's data for the present height
  - During finalization period, Commit step accepts precommit votes and timeouts, rejects all other inputs
  - New `additional_precommits` metric tracks signatures gathered during finalization

  These changes provide applications with flexibility to trade off latency for stronger consensus evidence, depending on their specific requirements.

  **IMPORTANT1**: Since the application gets the minimal commit certificate at decide time (just as before), it can make it available for Sync to have access to that current height during finalization period.

**IMPORTANT2**: I have left a few `TODO`s in the code with questions for the team

---

**Story:** <https://circlepay.atlassian.net/browse/CCHAIN-890>

---

#### Contribution eligibility

#### For all contributors

### DIFF
--- a/code/Cargo.lock
+++ b/code/Cargo.lock
@@ -1345,7 +1345,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "33d852cb9b869c2a9b3df2f71a3074817f01e1844f839a144f5fcef059a4eb5d"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2982,9 +2982,9 @@ checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
 
 [[package]]
 name = "libc"
-version = "0.2.175"
+version = "0.2.181"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a82ae493e598baaea5209805c49bbf2ea7de956d50d7da0da1164f9c6d28543"
+checksum = "459427e2af2b9c839b132acb702a1c654d95e10f8c326bfc2ad11310e458b1c5"
 
 [[package]]
 name = "libloading"
@@ -3445,9 +3445,9 @@ checksum = "d26c52dbd32dccf2d10cac7725f8eae5296885fb5703b261f7d0a0739ec807ab"
 
 [[package]]
 name = "linux-raw-sys"
-version = "0.9.2"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6db9c683daf087dc577b7506e9695b3d556a9f3849903fa28186283afd6809e9"
+checksum = "df1d3c3b53da64cf5760482273a98e575c651a67eec7f77df96b5b642de8f039"
 
 [[package]]
 name = "litemap"
@@ -4446,7 +4446,7 @@ dependencies = [
  "once_cell",
  "socket2 0.5.8",
  "tracing",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -4758,20 +4758,20 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.4.15",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
 name = "rustix"
-version = "1.0.2"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f7178faa4b75a30e269c71e61c353ce2748cf3d76f0c44c393f4e60abf49b825"
+checksum = "146c9e247ccc180c1f61615433868c99f3de3ae256a30a43b49f67c2d9171f34"
 dependencies = [
  "bitflags 2.9.0",
  "errno",
  "libc",
- "linux-raw-sys 0.9.2",
- "windows-sys 0.52.0",
+ "linux-raw-sys 0.11.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -5359,15 +5359,15 @@ checksum = "7b2093cf4c8eb1e67749a6762251bc9cd836b6fc171623bd0a9d324d37af2417"
 
 [[package]]
 name = "tempfile"
-version = "3.21.0"
+version = "3.25.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "15b61f8f20e3a6f7e0649d825294eaf317edce30f82cf6026e7e4cb9222a7d1e"
+checksum = "0136791f7c95b1f6dd99f9cc786b91bb81c3800b639b3478e561ddb7be95e5f1"
 dependencies = [
  "fastrand",
  "getrandom 0.3.1",
  "once_cell",
- "rustix 1.0.2",
- "windows-sys 0.52.0",
+ "rustix 1.1.3",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -6046,7 +6046,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
--- a/code/Cargo.toml
+++ b/code/Cargo.toml
@@ -172,7 +172,7 @@ k256               = { version = "0.13", default-features = false }
 p256               = { version = "0.13", default-features = false }
 p384               = { version = "0.13", default-features = false }
 ecdsa              = { version = "0.16", default-features = false }
-tempfile           = "3.21.0"
+tempfile           = "3.25.0"
 testdir            = "0.9.1"
 thiserror          = { version = "2.0", default-features = false }
 tikv-jemallocator  = "0.6.0"
--- a/code/crates/starknet/test/mbt/Cargo.toml
+++ b/code/crates/starknet/test/mbt/Cargo.toml
@@ -22,4 +22,4 @@ bytes = { workspace = true }
 
 
 [dev-dependencies]
-tempfile = { version = "3.21.0" }
+tempfile = { version = "3.25.0" }
--- a/code/crates/test/mbt/Cargo.toml
+++ b/code/crates/test/mbt/Cargo.toml
@@ -30,4 +30,4 @@ pretty_assertions = { workspace = true }
 multihash = { workspace = true }
 
 [dev-dependencies]
-tempfile = { version = "3.21.0" }
+tempfile = { version = "3.25.0" }
--- a/code/crates/wal/Cargo.toml
+++ b/code/crates/wal/Cargo.toml
@@ -32,5 +32,5 @@ lz4_flex = { version = "0.11.5", optional = true }
 criterion = "0.8.2"
 nix = { version = "0.29.0", features = ["signal"] }
 rand = "0.8.5"
-tempfile = "3.21.0"
+tempfile = "3.25.0"
 testdir = "0.9.1"


code/crates/core-consensus/src/handle/vote.rs
```diff
@@ -85,19 +85,48 @@ where
...
+        Effect::WalAppend(
+            signed_vote.height(),
+            WalEntry::ConsensusMsg(SignedConsensusMsg::Vote(signed_vote.clone())),
+            Default::default()
+        )
+    );
+
+    if state.driver.step_is_commit() {
+        // Directly add the vote to the vote keeper without going through state machine
+        // TODO: Do we want the state machine to process late votes in Commit step?
```
romac: cc @nenadmilosevic95 @cason

code/crates/core-consensus/src/handle/timeout.rs
```diff
@@ -33,6 +34,15 @@ where
         "Timeout elapsed"
     );
 
+    // TODO Is it correct to call finalize_height from here?
```
romac: Yes, that's the way to go I think.

code/crates/core-consensus/src/handle.rs
```diff
@@ -45,9 +46,50 @@ async fn handle_input<Ctx>(
 where
     Ctx: Context,
 {
+    // TODO is this redundant w.r.t. the state machine transitions?
+    if state.driver.step_is_commit() {
+        return match input {
+            Input::Vote(vote) => on_vote(co, state, metrics, vote).await,
+            Input::TimeoutElapsed(timeout) => on_timeout_elapsed(co, state, metrics, timeout).await,
+            Input::StartHeight(..) => {
+                panic!("Received StartHeight input during Commit step")
```
romac: Just a note that given that this is a public-facing API (via `process!`) I would suggest returning an error here instead of panicking. This point may be moot if we remove this code but just wanted to point this out in general.

code/crates/core-consensus/src/handle.rs
```diff
@@ -45,9 +46,50 @@ async fn handle_input<Ctx>(
 where
     Ctx: Context,
 {
+    // TODO is this redundant w.r.t. the state machine transitions?
```
romac: If at all possible I would have this logic live in either the driver or even the state machine. Not clear to me yet which is best, will think on it.

code/crates/core-consensus/src/handle.rs
```diff
@@ -45,9 +46,50 @@ async fn handle_input<Ctx>(
...
+                panic!("Received Propose input during Commit step")
+            }
+            _ => {
+                debug!("Ignoring input while in Commit step: {:?}", input);
+                Ok(())
+            }
+        };
+    }
+
+    // TODO is this redundant w.r.t. the state machine transitions?
```
romac: Same for this.

code/crates/app-channel/src/connector.rs
```diff
@@ -199,27 +199,62 @@ where
...
                         certificate,
                         extensions,
                         evidence,
-                        reply,
+                        reply: Some(reply),
                     })
                     .await?;
 
                 // Do not block processing of other messages while waiting for the next height
+                // TODO: do we still want to spawn here? I would say we do (app may still want to do stable times itself)
```
romac: > app may still want to do stable times itself

Yes, exactly, I would therefore keep this logic as-is.

code/crates/app-channel/src/msgs.rs
```diff
@@ -345,6 +346,30 @@ pub enum AppMsg<Ctx: Context> {
         /// Misbehavior evidence observed since last decide
         evidence: MisbehaviorEvidence<Ctx>,
```
romac: Note also that the evidence may also need to be extended in a validator equivocates in a late precommit for instance.

romac: More generally, I wonder what you think about:

- Removing the certificate, extensions and evidence from here and simply have the `ValueId` that's been decided
- Have the (potentially extended) commit certificate, extensions and evidence only in `Finalized`

code/crates/core-consensus/src/effect.rs
```diff
@@ -156,12 +156,28 @@ where
...
+    ///   * a boolean flag indicating whether a Finalize effect will follow.
+    ///     * if true, the application must not reply to this Decide message.
+    ///     * if false, the application must reply with Next.
     ///
     /// Resume with: [`resume::Continue`]
     Decide(
         CommitCertificate<Ctx>,
         VoteExtensions<Ctx>,
         MisbehaviorEvidence<Ctx>,
+        bool, // will_finalize
```
romac: Can we make this an enum to make semantics more explicit?

code/crates/engine/src/consensus.rs
```diff
@@ -1299,12 +1305,61 @@ where
                 let height = certificate.height;
 
                 // Notify the host about the decided value
+                if will_finalize {
+                    // Finalization will follow, so don't request a reply
```
romac: How about we always send `Finalized` to the app, regardless of whether a target time is set and remove the reply channel altogether for `Decided` (making it into purely informative message)?

code/crates/app-channel/src/connector.rs
romac: Actually, see my question below regarding removal of the reply field in `Decided`: https://github.com/circlefin/malachite/pull/1447#discussion_r2763565003

cason:
> When not set, consensus proceeds immediately to the next height as before, maintaining backward compatibility.

Consensus does not do that, it is the application.

cason:
> Consensus transitions to Finalize after deciding in all cases
> If target_time is configured, it will transition when a timer, set up upon deciding, expires
> No Finalize effect is emitted (backward compatibility)
> Otherwise, it will transition immmediately
> Finalize effect is emitted

It is the other way around. If configured, emits, otherwise don't (backwards compatibility).

cason:
> Waits for target_time - the target time reflects the time the app want the height to last (if possible)

Waits until the duration of the height meets `target_time` - the target time reflects the minimum time the app want the height to last

cason:
You should also use `Commit` certificates in your description all the times, to avoid confusion.

code/crates/core-consensus/src/effect.rs
```diff
@@ -156,12 +156,28 @@ where
...
+        resume::Continue,
+    ),
+
+    /// Notifies the application that a height has been finalized.
+    ///
+    /// This message includes an extended commit certificate containing all precommits
+    /// (both timely and late) that were received before the target time expired.
```
cason:
```suggestion
    /// for the decided value received until the target duration for the height.
```

code/crates/core-consensus/src/effect.rs
```diff
@@ -156,12 +156,28 @@ where
     /// In addition, it includes
     ///   * the vote extensions that were received for this height
     ///   * any misbehavior evidence (vote/proposal equivocation) detected during the height.
+    ///   * a boolean flag indicating whether a Finalize effect will follow.
+    ///     * if true, the application must not reply to this Decide message.
+    ///     * if false, the application must reply with Next.
```
cason: Replies are asynchronous. The application can always reply to `Decide`, even if it happens after `Finalize`?

code/crates/core-state-machine/src/input.rs
```diff
@@ -72,4 +72,8 @@ where
     /// Timeout waiting for precommits.
     /// L65
     TimeoutPrecommit,
+
```
cason: I don't see why the inner state-machine should be aware of this transition.

code/crates/core-state-machine/src/state_machine.rs
```diff
@@ -211,12 +211,21 @@ where
         }
 
         //
-        // From Commit. No more state transitions.
+        // From Commit. Accept transition to Finalize, otherwise invalid.
```
cason: I don't see why the inner state-machine should be aware of this transition.

code/crates/core-driver/src/driver.rs
```diff
@@ -632,11 +643,17 @@ where
 
             // The driver never receives these events, so we can just ignore them.
             TimeoutKind::Rebroadcast => return Ok(None),
+            TimeoutKind::FinalizeHeight(_) => return Ok(None),
         };
 
         self.apply_input(timeout.round, input)
     }
 
+    /// Apply the transition from Commit to Finalize step.
```
cason: Nothing happens at driver level, except for a `Step` change?

I have the impression that the driver should not be involved in this transition either.

cason: If anything would change here, it would change in `apply_vote`. Which does not care about which state we are at. So, again, in my opinion, nothing needs to change in the Driver.

code/crates/core-consensus/src/input.rs
```diff
@@ -14,7 +15,8 @@ where
 {
     /// Start consensus for the given height with an optional validator set update.
     /// The boolean indicates whether this is a restart of consensus for the given height.
-    StartHeight(Ctx::Height, Ctx::ValidatorSet, bool),
+    /// The optional Duration is the target time for this height.
+    StartHeight(Ctx::Height, Ctx::ValidatorSet, bool, Option<Duration>),
```
cason, line 19: Not really related to this PR, but can't we create a type with all "consensus params"?

code/crates/core-consensus/src/handle/vote.rs
```diff
@@ -85,19 +85,48 @@ where
...
-    if !state.driver.votes().has_vote(&signed_vote) {
-        perform!(
-            co,
-            Effect::WalAppend(
-                signed_vote.height(),
-                WalEntry::ConsensusMsg(SignedConsensusMsg::Vote(signed_vote.clone())),
-                Default::default()
-            )
-        );
+    if state.driver.votes().has_vote(&signed_vote) {
```
cason: Ideally, we should not even verify the signature here. Namely, lines 73-75 can be moved after this if.

code/crates/core-consensus/src/handle/vote.rs
cason: I think there is no harm of passing the vote to the driver, that will store it in the vote keeper.

Recall that in the current setup, once a decision is reached `step = Step::Commit` and we send a `Decide` to the application. The application may take a while to reply (ARC's app does).

In the meanwhile, inputs are received as usual, including votes, but no relevant state transition occurs (because `Commit` is a final state) and no output is produced. So, no, we should not give special treatment for `Precommit`s during `Step::Commit` except if we have a very strong reason for so.

code/crates/core-consensus/src/handle/vote.rs
```diff
@@ -85,19 +85,48 @@ where
...
+            let _ = state
+                .driver
+                .votes_mut()
+                .apply_vote(signed_vote.clone(), signed_vote.round());
+        }
+        return Ok(());
+    }
 
-        apply_driver_input(co, state, metrics, DriverInput::Vote(signed_vote)).await?;
+    // In Finalize step, ignore all votes
```
cason: This was not needed for `Step::Commit`, so no needed for `Step::Finalize` either.

I am not even sure if we need a `Step::Finalize`.

code/crates/core-consensus/src/handle/timeout.rs
```diff
@@ -33,6 +34,15 @@ where
...
+    if matches!(timeout.kind, TimeoutKind::FinalizeHeight(_)) {
```
cason: I am impressed that here we don't have a switch/case structure.

I have no idea why we pass a `RebroadcastTimeout` expiration to the driver, as `apply_timeout` in the driver just ignores it.

The same should happen to the introduced timeout: it is not a consensus timeout, it should not interact with the driver.

code/crates/core-consensus/src/handle/finalize.rs
```diff
@@ -0,0 +1,78 @@
...
+    Ctx: Context,
+{
+    assert!(
+        state.driver.step_is_commit(),
+        "finalize_height can only be called in Commit step"
+    );
+
+    let height = state.height();
+
+    let Some((proposal_round, decided_value)) = state.decided_value() else {
```
cason, line 22: Why checking this again here? We have a commit, therefore we have a decision value, the associated full value, the associated 2/3+ Precommits.

code/crates/core-consensus/src/handle/finalize.rs
```diff
@@ -0,0 +1,78 @@
...
+            certificate.clone(),
+            state.driver.validator_set().clone(),
+            state.params.threshold_params,
+        )
+        .await?
+        .is_ok(),
+        "Finalize: Commit certificate is not valid"
+    );
+
+    apply_driver_input(co, state, metrics, DriverInput::TransitionToFinalize).await?;
```
cason: This is a big noop going up and down the stack.

code/crates/core-consensus/src/handle/finalize.rs
```diff
@@ -0,0 +1,78 @@
...
+        return Err(Error::DecisionNotFound(height, state.round()));
+    };
+
+    let decided_id = decided_value.id();
+
+    let mut commits = state.restore_precommits(height, proposal_round, &decided_value);
+    let extensions = super::decide::extract_vote_extensions(&mut commits);
+    let certificate = CommitCertificate::new(height, proposal_round, decided_id, commits);
+
+    assert!(
```
cason, line 32: Why are we verifying all the signatures again?

code/crates/core-consensus/src/state.rs
```diff
@@ -35,6 +36,12 @@ where
 
     /// Last precommit broadcasted by this node
     pub last_signed_precommit: Option<SignedVote<Ctx>>,
+
+    /// Target time for the current height
+    pub target_time: Option<Duration>,
+
+    /// Start time of the current height
+    pub height_start_time: Option<Instant>,
```
cason, line 44: Can we just add a boolean here "finalized"?

code/crates/core-consensus/src/handle/driver.rs
```diff
@@ -141,6 +141,8 @@ where
 
         DriverInput::TimeoutElapsed(_) => (),
 
+        DriverInput::TransitionToFinalize => (),
```
cason: If we had a `state.finalized`, here we would set it to true. If it is already true, something is really wrong.

code/crates/core-consensus/src/handle/timeout.rs
```diff
@@ -33,6 +34,15 @@ where
...
+        if state.driver.step_is_commit() {
```
cason: If we had a `state.finalized`, here we would check if it is `false`.

code/crates/core-consensus/src/handle/decide.rs
```diff
@@ -53,7 +56,7 @@ where
         )
```
cason, line 52: All code above is ran _again_ upon `finalize`....

code/crates/core-consensus/src/handle/decide.rs
```diff
@@ -89,9 +96,52 @@ where
...
+        Effect::Decide(
+            certificate.clone(),
+            extensions.clone(),
+            evidence,
+            will_finalize,
+            Default::default()
+        )
     );
 
+    let Some(target_time) = state.target_time else {
```
cason, line 80: I don't see a point in logging that we are not doing something. And we don't need the driver... but I am being repetitive.

code/crates/core-consensus/src/handle/decide.rs
```diff
@@ -89,9 +96,52 @@ where
...
+    if elapsed < target_time {
+        // Do not transition to Finalize yet
+        let remaining = target_time - elapsed;
+        let timeout = Timeout::finalize_height(consensus_round, remaining);
+        perform!(co, Effect::ScheduleTimeout(timeout, Default::default()));
+
+        debug!(
+            height = %height,
+            remaining_ms = remaining.as_millis(),
+            "Staying in Commit step"
```
cason:
```suggestion
            "Scheduled Finalized effect"
```

code/crates/core-consensus/src/handle/decide.rs
```diff
@@ -89,9 +96,52 @@ where
...
+        );
+    } else {
```
cason: Why we don't just schedule a timeout of duration zero, or even negative? In this way we handle finalization in a single place in the code: the the timeout expires.

If we want to optimize, we optimize the implementation of `Effect::ScheduleTimeout`, right?

code/crates/core-consensus/src/handle.rs
cason: No use for that. Already today we process inputs while already in `Step::Commit` - and nothing bad happens.

code/crates/core-consensus/src/handle.rs
cason: No use for that. Already today we process inputs while already in `Step::Commit` - and nothing bad happens.

code/crates/app-channel/src/msgs.rs
```diff
@@ -345,6 +346,30 @@ pub enum AppMsg<Ctx: Context> {
...
 
+        /// Channel for instructing consensus to start the next height, if desired.
+        /// `None` if `Finalized` will follow.
+        reply: Option<Reply<Next<Ctx>>>,
+    },
+
+    /// Notifies the application that a height has been finalized after collecting late precommits.
```
cason: The are not late, their are additional. They might be received at the exactly same time.

code/crates/app-channel/src/msgs.rs
```diff
@@ -345,6 +346,30 @@ pub enum AppMsg<Ctx: Context> {
...
+    ///
+    /// This message is sent after a decision was made and a finalization period elapsed,
+    /// during which additional precommits were collected. The certificate may contain more
```
cason: is expected to contain, I'd say.

code/crates/app-channel/src/msgs.rs
```diff
@@ -345,6 +346,30 @@ pub enum AppMsg<Ctx: Context> {
...
         evidence: MisbehaviorEvidence<Ctx>,
 
+        /// Channel for instructing consensus to start the next height, if desired.
+        /// `None` if `Finalized` will follow.
+        reply: Option<Reply<Next<Ctx>>>,
+    },
+
+    /// Notifies the application that a height has been finalized after collecting late precommits.
+    ///
+    /// This message is sent after a decision was made and a finalization period elapsed,
```
cason: This elapsed period is confusing. This is the target duration for the height, which is configurable.

code/crates/app-channel/src/connector.rs
cason: Replies are asynchronous today with ARC application, they will continue like that.

code/crates/app-channel/src/connector.rs
```diff
@@ -199,27 +199,62 @@ where
...
+
+                self.sender
+                    .send(AppMsg::Finalized {
+                        certificate,
+                        extensions,
+                        reply,
+                    })
+                    .await?;
+
+                if let Ok(next) = rx.await {
```
cason: Well, here you are blocking consensus. So all logic for preventing stuff from happening during the finalized state has no much sense, right? I think the reply here should be also asynchronous, as for decide.

code/crates/engine/src/consensus.rs
sergio-mena: We can do that...

I decided to go for for the current solution in order to save other apps (not `malachite-app`) that depend on Malachite (but are not interested in adding target height times) the hassle of having to receive two messages (`Decided` and `Finalized`) with "apparently" no reason or advantage.

But, I'm also OK simplifying here, if we're willing to render non-target-time usage of Malachite a bit more cumbersome.

code/crates/engine/src/consensus.rs
```diff
@@ -1299,12 +1305,61 @@ where
...
+                }
+
+                // Notify the sync actor about the decided height
+                self.sync.send(SyncMsg::Decided(height));
+
+                Ok(r.resume_with(()))
+            }
+
+            Effect::Finalize(certificate, extensions, r) => {
+                assert!(!certificate.commit_signatures.is_empty());
```
cason, line 1300: What is the point here?

code/crates/core-consensus/src/handle.rs
sergio-mena: This part was the first thing I came up with, but it looks "patchy". On the other hand, I was afraid to touch the driver or state machine too much (I assumed you folks wouldn't like it 😄).

Then, I had second thoughts about this code, because it looks so much like "a quick hack", so decided to add a step to the state machine. Although I left these "if"'s here for the time being.

So, after your reply, I guess I can just nuke this, can't I?

code/crates/engine/src/consensus.rs
romac: > we're willing to render non-target-time usage of Malachite a bit more cumbersome.

It seems to me that an app that does not want to collect late precommits could ignore the `Decided` message and only process the `Finalized` message instead, or am I missing something?

It's a bit of breaking change for existing apps but the migration should be straightforward.

code/crates/app-channel/src/msgs.rs
romac: That said it's a bit weird that the `Commit` message does not carry the _commit_ certificate. On the other hand, it's also a bit unwieldy to receive two different but overlapping certificates and extensions. Not sure what's the best API here… Wonder what @ancazamfir thinks?

romac: Maybe we can go with what you have already, as it's backward compatible and figure out later if it's worth breaking backward compatibility.

romac:
> I cannot see any good reason to expose this transition to the driver and to the consensus state-machine.
>
> I also feel that restricting the changes to the core-consensus/engine would reduce a lot the number of changes required and make the solution much more elegant.

Agreed. I think it's worth giving it a try.

adizere:
Not a technical review, but only comment I have is that `Finalize` is an overloaded term, so I would avoid that. Please be more specific, eg "late commits" or "extended commits" or "post-commit" or "target time reached".

code/crates/core-consensus/src/handle/vote.rs
sergio-mena: Cool! If @romac also agrees I'll try to "remove the special case" here

code/crates/core-consensus/src/handle.rs
sergio-mena: @cason I wrote my reply before seeing your comments.

So can you confirm that, if I nuke these two "if"'s, I'm basically removing dead code?

code/crates/app-channel/src/connector.rs
sergio-mena: Removing TODO

code/crates/app-channel/src/connector.rs
sergio-mena: Since we have already waited, my understanding is that the app must reply quickly here (just as it used to be in decide before we started sleeping for stable times).

Having said that, in my current implementation on `malachite-app` side, I'm keeping the "target block times" code to minimze changes (following the thinking that we can do it later).

So I guess you got a point here. Will change it

code/crates/app-channel/src/msgs.rs
sergio-mena: A few comments:

The initial commit (the minimal commit, as @cason defined them in his doc) is anyway necessary at decide time, so that Sync can access it, already during the finalization period (this is also the answer to a question that @ancazamfir asked me last Fri)

About evidence on late commit. My understanding from last week's discussion is that we decided to leave them out of this effort (and tackle as an improvement later), but my memories might be betraying me. I could add them, but in that case, I'd rather do it in a different PR.

> unwieldy to receive two different but overlapping certificates and extensions.

We have a use case for that: as mentioned above, the minimal commit certificate helps Sync having the certificate as soon as it's produced.

code/crates/app-channel/src/msgs.rs
sergio-mena: Ok, I'll reword "late" to "additional" all over

code/crates/app-channel/src/msgs.rs
sergio-mena: Will reword

code/crates/app-channel/src/msgs.rs
sergio-mena: reworded

code/crates/core-consensus/src/handle/decide.rs
sergio-mena: Do you mean that:
* We shouldn't assert on `verify_commit_certificate` in Finalilze? I left it for sanity/defensive programming
* The `let (certificate, extensions) = ...` is not really the same, as Finalize is collecting from (potentially, "likely" in the happy path) more precommits

code/crates/core-consensus/src/handle/decide.rs
sergio-mena: I introduced it to debug the tests, but I can remove it

code/crates/core-consensus/src/handle/decide.rs
sergio-mena: I thought about it, but preferred this way. No _dummy_ timer scheduled when it's not needed.

However, I'm open to changing that if there's agreement it's better like that

code/crates/core-consensus/src/handle/driver.rs
sergio-mena: At some point, I had a boolean like the one you suggest; but it made all the rest of the code I added look like "a quick hack on top of the base behavior". So I decided to add the new `Step` and nuke the boolean here

code/crates/core-consensus/src/handle/finalize.rs
sergio-mena: I can change it to an `expect`, but I think returning an error is cleaner

code/crates/core-consensus/src/handle/finalize.rs
sergio-mena: As I said, it's an assert (sanity/defensive programming). I can turn it into a `debug_assert`

code/crates/core-consensus/src/handle/finalize.rs
sergio-mena: Interesting.

I had left this here while waiting on the question whether the state machine should process the late precommits (it's not doing it ATM, see `code/crates/core-consensus/src/handle/vote.rs`).

In that comment, you seem to be advocating for passing having the driver deal with the late prevotes... but now, I seem to understand that you're not keen on extending the state machine? Am I getting it right?

code/crates/core-consensus/src/handle/timeout.rs
sergio-mena: Removing TODO

code/crates/core-consensus/src/handle/timeout.rs
sergio-mena: OK, I'm starting to understand...

You actually (strongly) prefer a boolean here, rather than extending the state machine. Is this accurate?

Anyone else having strong opinions here? @romac @ancazamfir 

I see @cason seems to have a strong opinion here, so I'll probably revert to the previous version I had where I was having the boolean (I think I still have it in my `git` history)

code/crates/core-consensus/src/handle/vote.rs
sergio-mena: Makes sense. Will move the signature verification down here

code/crates/core-consensus/src/handle.rs
sergio-mena: Coming back here. If we are going to replace the extra step by a boolean, are these "if"s still unneeded? 
With a boolean, I'm not sure about that

code/crates/core-consensus/src/input.rs
sergio-mena: If you think so, please open a ticket

code/crates/core-consensus/src/state.rs
sergio-mena: We need the start time to know how long we need to sleep when we reach Decide

code/crates/core-driver/src/driver.rs
sergio-mena: So you would leave the driver out?

Again I had a version doing that, which felt too hacky. But you seem to be liking that more than involving the driver and adding a new step....

code/crates/core-state-machine/src/input.rs
sergio-mena: As mentioned above. My intention was to get rid of the boolean that I found ugly. But you seem to have a different opinion...

Also as mentioned above, if nobody else objects, I will revert back to removing this, the new Step, and reintroducing the boolean. That code feels hackier to me, but I don't have as much context as you, so I'll trust your intuition

code/crates/core-state-machine/src/state_machine.rs
sergio-mena: Ditto. Will revert to the boolean shortly

code/crates/engine/src/consensus.rs
sergio-mena: OK. So it looks like that's the way to go.
I'll remove the reply from Decided and Finalize will always be emitted

code/crates/engine/src/consensus.rs
sergio-mena: Defensive programming

code/crates/app-channel/src/msgs.rs
romac: > We have a use case for that: as mentioned above, the minimal commit certificate helps Sync having the certificate as soon as it's produced.

Ah good point, I forgot about that. Nevermind then :)

code/crates/core-consensus/src/effect.rs
sergio-mena: No need anymore. The boolean has been nuked

code/crates/engine/src/consensus.rs
sergio-mena: Ok, after some fiddling, I got it working. It just pushed a commit. Please take a look at the result.

sergio-mena:
> Agreed. I think it's worth giving it a try.

Good. I will revert to the "boolean-based" solution I previously had, which was avoiding to touch the state machine and (IIRC) the driver.

code/crates/core-consensus/src/handle/vote.rs
sergio-mena: Outdated

code/crates/core-consensus/src/handle.rs
sergio-mena: Addressed

code/crates/core-consensus/src/handle.rs
sergio-mena: @cason with the new code (no longer touching the driver of the state machine), can you confirm that ALL the checks in this if are redundant?

code/crates/app-channel/src/msgs.rs
sergio-mena: @romac now that `Finalized` is always emitted, do you think returning evidence in `Finalized` is better?

(My opinion: it's the same, but returning it in `Finalized` would be less disruptive for the user if, later, we decide to collect further evidence during the finalization period)

code/crates/core-consensus/src/effect.rs
sergio-mena: the text no longer exists

code/crates/core-state-machine/src/input.rs
sergio-mena: I reverted to a boolean-based implementation

code/crates/core-state-machine/src/state_machine.rs
sergio-mena: I reverted to a boolean-based implementation

code/crates/core-driver/src/driver.rs
sergio-mena: I reverted to a boolean-based implementation

code/crates/core-consensus/src/handle/vote.rs
sergio-mena: I reverted to a boolean-based implementation

code/crates/core-consensus/src/handle/finalize.rs
sergio-mena: I reverted to a boolean-based implementation

code/crates/core-consensus/src/state.rs
sergio-mena: I reverted to a boolean-based implementation

code/crates/core-consensus/src/handle/driver.rs
sergio-mena: I reverted to a boolean-based implementation

code/crates/core-consensus/src/handle/timeout.rs
sergio-mena: I reverted to a boolean-based implementation

sergio-mena:
Boolean-based implementation in place. Please take a look

code/crates/core-consensus/src/error.rs
```diff
@@ -57,4 +57,8 @@ where
     /// The write-ahead log is corrupted.
     #[error("Write-ahead log is corrupted: {0}")]
     WalCorrupted(Arc<io::Error>),
+
+    /// Received unexpected input during Commit or Finalize step.
```
sergio-mena:
```suggestion
    /// Received unexpected input during Commit step.
```

code/crates/core-consensus/src/error.rs
```diff
@@ -57,4 +57,8 @@ where
     /// The write-ahead log is corrupted.
     #[error("Write-ahead log is corrupted: {0}")]
     WalCorrupted(Arc<io::Error>),
+
+    /// Received unexpected input during Commit step.
```
sergio-mena:
```suggestion
    /// Received unexpected input during a step.
```

code/crates/core-consensus/src/handle/decide.rs
romac: I am unable to make up my mind about this one:

- On one hand it would be nice to have a single code path regardless of target time being set or not.
- On the other hand, it seems wasteful and counter-intuitive to schedule a 0-duration timeout (even if optimized in handling of `ScheduleTimeout` effect).

code/crates/app-channel/src/msgs.rs
romac: I agree, if there is no strong case for collecting the evidence in `Decided`, then let's only have it in `Finalized` .

code/crates/core-consensus/src/handle/timeout.rs
romac: > I have no idea why we pass a RebroadcastTimeout expiration to the driver, as apply_timeout in the driver just ignores it.

Good catch, let's address this in a follow-up PR.

romac: > The same should happen to the introduced timeout: it is not a consensus timeout, it should not interact with the driver.

True, note that this timeout is indeed not passed down to the driver because of the `return` statements below.

code/crates/core-consensus/src/handle/vote.rs
```diff
@@ -70,6 +70,11 @@ where
 
     debug_assert_eq!(consensus_height, vote_height);
 
+    // Only process this vote if we have not yet seen it.
+    if state.driver.votes().has_vote(&signed_vote) {
+        return Ok(());
+    }
+
```
romac, line 77: Good catch, there is indeed no need to verify the vote if we've already seen it.

code/crates/starknet/host/src/actor.rs
```diff
@@ -636,12 +632,9 @@ async fn store_proposed_value(
     }
 }
 
-async fn on_decided(
+async fn on_decided_notification(
```
romac: For consistency, I would keep the original name here.

code/crates/core-consensus/src/handle/driver.rs
```diff
@@ -141,8 +141,6 @@ where
 
         DriverInput::TimeoutElapsed(_) => (),
 
-        DriverInput::TransitionToFinalize => (),
```
romac: 🎉

code/crates/test/app/src/app.rs
```diff
@@ -214,13 +215,36 @@ pub async fn run(state: &mut State, channels: &mut Channels<TestContext>) -> eyr
...
+                    signatures = certificate.commit_signatures.len(),
+                    "Consensus has decided on value, awaiting Finalized message..."
+                );
+
+                // Storing now so Sync can see it
+                if let Err(e) = state.store_decided(certificate).await {
+                    error!("Failed to store decided value: {e}");
+                }
+
+                sleep(Duration::from_millis(500)).await;
```
romac: We probably don't want to sleep in both Decided and Finalized, but only in the latter, right?

romac: And perhaps only if we are not using the target time?

code/crates/core-consensus/src/handle/vote.rs
```diff
@@ -84,20 +89,43 @@ where
...
+        co,
+        Effect::WalAppend(
+            signed_vote.height(),
+            WalEntry::ConsensusMsg(SignedConsensusMsg::Vote(signed_vote.clone())),
+            Default::default()
+        )
+    );
 
-        apply_driver_input(co, state, metrics, DriverInput::Vote(signed_vote)).await?;
+    // Directly add the precommit to the vote keeper without going through state machine
```
romac, line 101: It's probably worth documenting why this is needed

code/crates/core-consensus/src/handle/decide.rs
romac: I think we can leave it as is for now and revisit later.

code/crates/core-consensus/src/handle/timeout.rs
sergio-mena: I just pushed a commit addressing this

sergio-mena: Please take a look to make sure that's what you folks want

(@romac I had addressed it already when I saw your comment proposing to fix it in a follow-up PR)

code/crates/starknet/host/src/actor.rs
sergio-mena: 👍

code/crates/test/app/src/app.rs
sergio-mena: I don't really understand the goal of the sleep, so trusting your intuition here

sergio-mena: > only if we are not using the target time?

Having second thoughts here. What if we configure a 10ms target time in the tests?
I'd leave the sleep as is in the Finalized.
WDYT?

code/crates/core-consensus/src/handle/vote.rs
sergio-mena: Not sure what you mean... do you mean in order to keep the state machine orthogonal to non-tendermint-algo changes?

code/crates/app-channel/src/msgs.rs
sergio-mena: OK, will do after lunch

code/crates/test/app/src/app.rs
romac: > I don't really understand the goal of the sleep, so trusting your intuition here

It's to emulate stable block times, otherwise the test app produces blocks much too fast.

romac: Can we sleep in Finalized only if target time is disabled?

code/crates/core-consensus/src/handle/vote.rs
romac: I mean why why apply the vote directly in the vote keeper without going through `apply_driver_input`

romac: sorry misread what you typed, yes that's what I meant :)

code/crates/app-channel/src/msgs.rs
sergio-mena: I pushed a commit with this. If think that, as a side effect, we're now collecting evidence also during finalization period, but I'm not 100% sure

code/crates/test/app/src/app.rs
sergio-mena: Sure, I trust you on this one

code/examples/channel/src/app.rs
```diff
@@ -241,6 +242,22 @@ pub async fn run(state: &mut State, channels: &mut Channels<TestContext>) -> eyr
...
+                info!(
+                    height = %certificate.height,
+                    round = %certificate.round,
+                    value = %certificate.value_id,
+                    signatures = certificate.commit_signatures.len(),
+                    "Consensus has decided on value, awaiting Finalized..."
+                );
+
+                // Sleep a bit to slow down the app.
+                sleep(Duration::from_millis(500)).await;
```
cason, line 257: I don't think this has any effect on Malachite.

code/crates/test/tests/unit/certificates/commit.rs
```diff
@@ -177,3 +177,31 @@ fn commit_certificate_with_mixed_valid_and_invalid_votes() {
...
+    // Extended certificate
+    CertificateTest::<Commit>::new()
+        .with_validators([20, 20, 20, 20, 20, 20, 20])
+        .with_votes(1..7, VoteType::Precommit)
+        .expect_valid();
+
+    // Full certificate
+    CertificateTest::<Commit>::new()
+        .with_validators([20, 20, 20, 20, 20, 20, 20])
+        .with_votes(0..7, VoteType::Precommit)
```
cason, line 199: The range being exclusive for the last confuses me all the time.

code/crates/app-channel/src/connector.rs
```diff
@@ -214,7 +226,7 @@ where
                 tokio::spawn(async move {
                     if let Ok(next) = rx.await {
                         if let Err(e) = reply_to.send(next) {
-                            error!("Failed to send next height and validator set: {e}");
+                            error!("Finalized: Failed to send next height and validator set: {e}");
```
cason:
```suggestion
                            error!("Finalized: connector failed to send StartHeight: {e}");
```

code/crates/core-consensus/src/handle.rs
cason: Dead code, can remove.

code/crates/core-consensus/src/handle/decide.rs
cason: > We shouldn't assert on verify_commit_certificate in Finalilze? I left it for sanity/defensive programming

Yes, this is becoming Comet. Ok, it does not harm - except for the cost of N signature verifications, which we are verifying for the THIRD time.

cason: Should we send only extensions on `Finalize`?

code/crates/core-consensus/src/handle/decide.rs
cason: I would remove. We are producing 3 logs messages in sequence: decide, finalize immediately, finalize. We should log if we don't do this immediately.

code/crates/core-consensus/src/handle/timeout.rs
```diff
@@ -33,22 +34,27 @@ where
...
-            co,
-            Effect::WalAppend(height, WalEntry::Timeout(timeout), Default::default())
-        );
-    }
-
-    apply_driver_input(co, state, metrics, DriverInput::TimeoutElapsed(timeout)).await?;
-
-    if matches!(timeout.kind, TimeoutKind::Rebroadcast) {
-        on_rebroadcast_timeout(co, state, metrics).await?;
+    match timeout.kind {
```
cason, line 37: 👍

code/crates/core-consensus/src/handle/vote.rs
cason: I still don't like this, but this can be left to a follow-up PR.

code/crates/core-consensus/src/state.rs
```diff
@@ -35,6 +36,15 @@ where
...
```
cason, line 44: I don't think we need to have an option here, this data can be useful for logging etc.

code/crates/core-consensus/src/handle.rs
cason: I don't know how Malachite can receive a `StartHeight` without sending a `Finalize` with a response channel to receive it.

The same for `Propose`:  it is a reply to `GetValue`, isn't it?

If there is a way to receive, well, maybe double checking does not harm.

cason:
> IMPORTANT1: Since the application gets the minimal commit certificate at decide time (just as before), it can make it available for Sync to have access to that current height during finalization period.

This is a good point, I hadn't seen it when leaving my review. I think `Decide` MUST include a valid certificate. The question are vote extensions. What is the point of delivering them twice, upon `Decide` and `Finalize`.

Taking a step back, it appears to the that:

1. `Decide` is the consensus output for a `Height`
2. `Finalize` contains all the additional data that consensus can deliver to the application, namely extended `Commit`, evidences, _and_ vote extensions.

romac:
Some unrelated integration tests are failing, haven't manage to figure out why yet.

cason:
`start_late_parallel_requests_with_batching` is flaky and fails also locally. I could not make it foo fail in `main` though.

code/crates/core-consensus/src/handle/decide.rs
romac: > Ok, it does not harm - except for the cost of N signature verifications, which we are verifying for the THIRD time.

Agree we should strive to minimize duplicated work, especially signature verification as it is potentially quite expensive.

It would be nice to have a way to statically track whether a commit certificate has already been verified or not. We can leave this for a follow-up PR.

code/examples/channel/src/app.rs
romac: It will delay the processing of the Finalized message, but it's redundant. Let's remove this sleep and keep the one in Finalized.

romac:
I wonder if we have broken something by not calling `middleware.on_commit` in `Decided`?

code/crates/core-consensus/src/handle/decide.rs
```diff
@@ -75,23 +75,49 @@ where
...
-        votes: state.driver.take_vote_evidence(),
-    };
-
     perform!(
         co,
-        Effect::Decide(certificate, extensions, evidence, Default::default())
+        Effect::Decide(certificate.clone(), extensions.clone(), Default::default())
     );
 
+    let Some(target_time) = state.target_time else {
```
ancazamfir, line 80: If we have `target_time` configured and we decide via sync we should also finalize immediately? Maybe check `existing_certificate.is_none()`

code/crates/core-consensus/src/handle/finalize.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let height = state.height();
+
+    let Some((proposal_round, decided_value)) = state.decided_value() else {
+        return Err(Error::DecisionNotFound(height, state.round()));
+    };
+
+    let decided_id = decided_value.id();
+
+    let mut commits = state.restore_precommits(height, proposal_round, &decided_value);
```
ancazamfir, line 28: This breaks sync as there will be no precommits stored just the certificate. Here we need to combine them for now (*). You will see something like
```
 [/consensus,/liveness,/proposal_parts], outbound, 0
2026-02-06T10:32:38.975546Z  INFO test{id=1}:node{id=3}:node{moniker=node-3}:consensus{height=1 round=0}: Timeout elapsed step=FinalizeHeight(9.985000583s) timeout.round=0 height=1 round=0

thread 'value_sync::start_late_1' (9830221) panicked at /Users/anca.zamfir/arc/malachite/code/crates/core-consensus/src/handle/finalize.rs:32:5:
Finalize: Commit certificate is not valid
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2026-02-06T10:32:38.977141Z ERROR test{id=1}:node{id=3}:node{moniker=node-3}:node: Actor 0.15 has failed: Finalize: Commit certificate is not valid
```
I think we should have some sync tests using `target_time`. I have this local change in `start_late()`, test fails
```
diff --git a/code/crates/test/tests/it/value_sync.rs b/code/crates/test/tests/it/value_sync.rs
index 3c76d9aa..f04293b8 100644
--- a/code/crates/test/tests/it/value_sync.rs
+++ b/code/crates/test/tests/it/value_sync.rs
@@ -200,6 +200,7 @@ pub async fn start_late() {
             Duration::from_secs(30),
             TestParams {
                 enable_value_sync: true,
+                target_time: Some(Duration::from_secs(10)),
                 ..Default::default()
             },
         )
```         

Then there is the matter of ignoring `target_time` if we decide via sync (i.e. in decide we have a commit certificate already) which might make this issue moot. But we should still make some sync tests using target_time

code/crates/app-channel/src/msgs.rs
ancazamfir: regarding the Sync issue here is what I was thinking. A peer will not know we have decided `H` until it receives a Status with `H`. The problem is that sync receives a `Decided` event from consensus around the same time as app gets it. Sync then sets the tip height to `H`.  If ticker fires before finalize then this peer will send a `Status(H)` to us. At that point we query the peer's (app) value at `H`. 

Although unlikely, the peer's app might have not fully process de `Decided` when sync asks it for value at `H` and might fail (this has nothing to do with this PR and we should figure out if/how we want to fix this). 

The more likely case is that the peer has the value and ideally the certificate, that's why I mentioned when talking with @sergio-mena that it's good to make it part of `Decided` message.

In general, imho is good to keep the existing Decided with certificate and evidence. Then in Finalize we send again the same or extended. The app can figure out what to do (stores as is, or the diffs, etc)

ancazamfir:
> `start_late_parallel_requests_with_batching` is flaky and fails also locally. I could not make it foo fail in `main` though.

I am not able to make it fail, have you changed anything in the test or it just fails as is? And in all runs?

cason:
> > `start_late_parallel_requests_with_batching` is flaky and fails also locally. I could not make it foo fail in `main` though.
> 
> I am not able to make it fail, have you changed anything in the test or it just fails as is? And in all runs?

It is flaky, fails 1/4 of the times here:

```
malachite/code/crates/test/tests/it $ while cargo test start_late_parallel_requests_with_batching --test it; do done
    Finished `test` profile [optimized + debuginfo] target(s) in 0.17s
     Running tests/it/main.rs (/Users/daniel.cason/arc/malachite/code/target/debug/deps/it-eafe2d40df6ed909)

running 1 test
test value_sync::start_late_parallel_requests_with_batching ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 67 filtered out; finished in 14.69s

    Finished `test` profile [optimized + debuginfo] target(s) in 0.16s
     Running tests/it/main.rs (/Users/daniel.cason/arc/malachite/code/target/debug/deps/it-eafe2d40df6ed909)

running 1 test
error: test failed, to rerun pass `--test it`

Caused by:
  process didn't exit successfully: `/Users/daniel.cason/arc/malachite/code/target/debug/deps/it-eafe2d40df6ed909 start_late_parallel_requests_with_batching` (exit status: 1)
note: test exited abnormally; to see the full output pass --no-capture to the harness.
```

code/crates/core-consensus/src/handle/finalize.rs
cason:
```
    // FIXME: there is actual no guarantee that associated vote extensions can be found,
    // in particular when deciding via sync, see: https://circlepay.atlassian.net/browse/CCHAIN-915.
```

This is in the current code, the problem is not new. We don't support vote extensions, in few words.

code/crates/core-consensus/src/handle/decide.rs
cason: So, this is part of the logic that is supposed to reside in the application. If we are lagging behind, we should not pass a target time - in the same way that we currently don't use the stable block times logic in the application.

code/crates/core-consensus/src/handle/finalize.rs
ancazamfir: this is not because of vote extensions. it's empty commit certificate in finalize.

code/crates/core-consensus/src/handle/decide.rs
ancazamfir: I see. So it is possible target times is passed at StartHeight and we sync that height. Then we will assert in finalize with empty certificate.

code/crates/core-consensus/src/handle.rs
ancazamfir: We should move these checks in the handlers. We want to buffer inputs from higher heights (including sync responses which are dropped here). This happens in the individual handlers that drop lower heights, buffer higher heights, process current_heights.
The `- => {` match arm bellow is the reason the sync tests fail in this branch.
If we want to keep it here for some reason I miss, let's add
```
Input::SyncValueResponse(value) => {
                return on_value_response(co, state, metrics, value).await
            }
```
We should also allow `Proposal`s, maybe round certificates.

ancazamfir:
> `start_late_parallel_requests_with_batching` is flaky and fails also locally. I could not make it foo fail in `main` though.

I finally managed to reproduce, the failure is introduced in the PR branch, see https://github.com/circlefin/malachite/pull/1447/changes#r2777494755

code/crates/core-consensus/src/handle.rs
cason: I can confirm that by removing this code block, the errors are not produced.

cason: > If we want to keep it here for some reason I miss, let's add

I can't see any reason to keep it here.

cason:
Issues with the flaky equivocation test. Trying again.

code/crates/core-consensus/src/handle/finalize.rs
cason: You are right. We should check for an existing `Commit` certificate here too. The problem is that we never extend it...

code/crates/core-consensus/src/handle.rs
sergio-mena: Thanks for this. I was never sure if these checks should be here. I thought that, at worst, the would be redundant. But form this convo, I see they're not redundant but potentially harmful.

code/crates/core-consensus/src/handle/decide.rs
sergio-mena: The idea is that the target height time inside malachite shouldn't depend on what the application does.
So, if we're deciding because of sync, we shouldn't wait.
I hadn't realised about this condition when working on this PR, so that's definitely a gap in the PR.
For the record, @cason and I are physically at the same place, and agreed on a solution in `decide.rs`. He's testing it and will push a couple of commits

code/crates/core-consensus/src/handle/finalize.rs
sergio-mena: Thanks folks for spotting this! 🚀🚀🚀
As said above, I hadn't realised about this case when working on this PR. So it's a gap in its logic.

cason: @sergio-mena kind of agreed with you that decisions via Sync => target time can be ignored and we finalize immediately. We have an updated test unit that attest this behaviour (fails with this code), and a workaround.

I am not much of a fan of this, but... it should work.

sergio-mena: For the record, when rushing to prepare the meeting with the Data Engineering folks on Fri, I spun up a couple of Quake testnets and was alarmed by the "crankiness" of the system, but didn't have the time to investigate.

The symptoms I was seeing match your findings here. So, once @cason pushes the fix that we discussed (in `decide.rs`), I will test again with quake to make sure the "crankiness" has disappeared

code/crates/core-consensus/src/handle.rs
sergio-mena: Update on this: @cason just removed those "if"s

code/crates/app-channel/src/msgs.rs
sergio-mena: @ancazamfir do you want me to open a ticket for this? (I agree we could tackle this in another PR, as this PR is blocking 3 more PRs on `circle-chain-reth`)

code/crates/core-consensus/src/handle/finalize.rs
sergio-mena: Talked in person. Resolving

code/crates/core-consensus/src/handle/decide.rs
sergio-mena: I agree, I filed [CCHAIN-1061](https://circlepay.atlassian.net/browse/CCHAIN-1061)

Resolving

sergio-mena:
@romac will you have the time tomorrow to take a final look and, if nothing else looks off, approve? Thanks 🙏

code/crates/core-consensus/src/handle/vote.rs
```diff
@@ -84,20 +89,44 @@ where
         "Received vote",
     );
 
-    // Only process this vote if we have not yet seen it.
-    if !state.driver.votes().has_vote(&signed_vote) {
```
romac, line 88: What's the reasoning for removing this check?

romac:
@sergio-mena I took another look, left a question above, and pushed a few minor commits that you may want to take a look at.

Note also that you will have to push another commit for me to be able to approve in the end, since I am currently the last pusher.

The PR looks great, really good work there! 🏅

code/crates/core-consensus/src/handle/vote.rs
romac: For reference, this check was introduced in https://github.com/circlefin/malachite/pull/902 to avoid storing duplicate votes to the WAL.

sergio-mena:
Thanks a lot @romac, I reviewed the latest changes and they LGTM (thanks for the refactoring!)

I just pushed an empty commit. Let me know if that's not enough for you to approve (if you also want to merge directly, please go ahead 🙏)

code/crates/core-consensus/src/handle/vote.rs
sergio-mena: Hmmm, I didn't remove it, I moved it up... let me check

sergio-mena: Yes, I checked the latest version and you see:

* The hunk you refer to here, which is removing the check for duplicate votes (lines 87-88 in the old version of the file)
* The hunk above it (lines 73-77 in the new version of the file), which returns early if the vote is a duplicate

Anyway, can you please review the logic to make sure nothing's being overlooked?

romac: Ah my bad, didn't see it this time around. All good then :)

romac:
I did some testing with the channel app and a target time of 2s, but was never able to collect all precommits with 20 validators, only the 7 required + 3 or 4 additional ones. Does that match what you've seen in your own tests?

In any case, I am pre-approving already as that's not a blocker.

sergio-mena:
> I did some testing with the channel app and a target time of 2s, but was never able to collect all precommits with 20 validators, only the 7 required + 3 or 4 additional ones. Does that match what you've seen in your own tests?
> 
> In any case, I am pre-approving already as that's not a blocker.

@romac I was heavily testing this on Wed (early morning for me) using Quake. I used [circle-chain-reth#947](https://github.com/crcl-main/circle-chain-reth/pull/947) (which uses this patch). Apart from a glitch that lasted around 1000 heights, after a late validator caught up (I set it up like this in quake), all validators were collecting full certificates for 1000s of heights.
If you think it's worth it, I can try to repro that "glitch" again... I didn't check the logs at that moment, although I should have.

romac:
> If you think it's worth it, I can try to repro that "glitch" again... I didn't check the logs at that moment, although I should have.

I will try to reproduce again myself and will keep you posted. It's very likely that I was either doing something wrong or that the issue lies in the Malachite channel-based example app that I was using.

cason:
> > If you think it's worth it, I can try to repro that "glitch" again... I didn't check the logs at that moment, although I should have.
> 
> I will try to reproduce again myself and will keep you posted. It's very likely that I was either doing something wrong or that the issue lies in the Malachite channel-based example app that I was using.

I think the issue is that the channel app sleeps 500ms after a decision, not?

PR #1448, opened 2026-02-04
feat(sync): Add credit-based scoring strategy

romac:
Add a simple credit-based strategy for scoring peers in the sync protocol.

- Maintain an integer `credit` per peer.
- Fast success increases credit more than slow success.
- Failures and timeouts reduce credit.
- Credits are clamped to `[min_credit, max_credit]`.
- Score is a normalized mapping of `credit -> [0.0, 1.0]`.

Credit is updated based on the sync result, via these deltas:

```rust
pub credit_fast_success: i32,
pub credit_slow_success: i32,
pub credit_failure: i32,
pub credit_timeout: i32,
```

## TODO
- [ ] Find good default values for the credit deltas

Current values for credit strategy:

```rust
CreditConfig {
    slow_threshold: Duration::from_millis(500),
    credit_fast_success: 2,
    credit_slow_success: 0,
    credit_failure: -2,
    credit_timeout: -4,
    min_credit: -20,
    max_credit: 20,
}
```

**Notes:**

- Credit ranges from `-20` to `20`
- Initial credit is derived from the initial score of `0.5` and is therefore `0`
- Timeouts are more heavily penalized than failures, as those hurt sync performance more than failures
- Slow successes do not change credit

PR #1449, opened 2026-02-04
refactor(sync): Refactor the sync scoring system

romac:
Follow-up to #1448

Actual changes:
- https://github.com/circlefin/malachite/pull/1449/commits/7f8f82de6cd116054e0dfc61d29b1d4dd7f44841
- https://github.com/circlefin/malachite/pull/1449/commits/828c3f89d4901ae04443476b70753af0aba96e26

### DIFF
--- a/code/crates/sync/src/scoring.rs
+++ b/code/crates/sync/src/scoring.rs
@@ -177,7 +177,10 @@ impl<S: ScoringStrategy> PeerScorer<S> {
             return None;
         }
 
-        let scores = peers.iter().map(|id| self.get_score(id).max(0.0));
+        // Use a minimum floor so that even the lowest-scoring peers
+        // retain a non-zero probability of being selected.
+        const MIN_WEIGHT: f64 = 0.01;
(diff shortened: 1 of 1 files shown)


code/crates/sync/src/scoring.rs
```diff
@@ -150,7 +177,10 @@ impl PeerScorer {
             return None;
         }
 
-        let scores = peers.iter().map(|id| self.get_score(id).max(0.0));
+        // Use a minimum floor so that even the lowest-scoring peers
+        // retain a non-zero probability of being selected.
+        const MIN_WEIGHT: f64 = 0.01;
+        let scores = peers.iter().map(|id| self.get_score(id).max(MIN_WEIGHT));
```
romac, line 183: I wonder whether we want to keep this behavior or not? I've only added it so that the credit strategy would not cause scores to drop to 0, to match the emergent property of the EMA scoring system that scores are never quite zero (cf. See the pre-existing `all_peers_selectable` test).

At some point, we probably want to be ban/disconnect from peers whose score has dropped to zero.

romac:
Development of this PR has moved to a different repository.

PR #1436, opened 2026-02-03
docs(CCHAIN-793): ADR for Proof-of-Validator protocol

nenadmilosevic95:
Closes #1444. 

## Summary
This PR introduces ADR-006: Proof-of-Validator (PoV), a protocol that enables validator nodes to cryptographically prove their validator status to peers in the network.

### Use cases:
- Debugging and observability: classify peers as validators vs other nodes
- Connection prioritization: prefer connections to validators for reliable consensus message delivery
- Mesh formation optimization: build gossipsub mesh overlay that prioritizes validator peers
### Key design decisions:
- Store-and-re-evaluate approach: proofs are stored once and re-evaluated when validator set updates arrive
- One proof per peer lifetime: anti-spam protection with bounded storage

**Story**: [https://circlepay.atlassian.net/browse/CCHAIN-793](https://circlepay.atlassian.net/jira/software/projects/CCHAIN/boards/1021?selectedIssue=CCHAIN-793)
**Old PR**:  https://github.com/circlefin/circle-chain-consensus/pull/103

PR #1450, opened 2026-02-04
feat(CCHAIN-740): Implement validator proof protocol per ADR-006

ancazamfir:
Implementation of https://github.com/circlefin/malachite/pull/1436

Uses libp2p-stream protocol to satisfy the protocol spec and as proposed by @romac

### DIFF
--- a/code/crates/signing/src/ext.rs
+++ b/code/crates/signing/src/ext.rs
@@ -104,14 +104,14 @@ where
         thresholds: ThresholdParams,
     ) -> Result<(), CertificateError<Ctx>>;
 
-    /// Sign a validator certificate binding the given public key to the given peer ID.
+    /// Sign a validator proof binding the given public key to the given peer ID.
     async fn sign_validator_proof(
         &self,
         public_key: Vec<u8>,
         peer_id: Vec<u8>,
     ) -> Result<ValidatorProof<Ctx>, Error>;
 
-    /// Verify a validator certificate's signature using the public key included in the certificate.
+    /// Verify a validator proof's signature using the public key included in the certificate.
     ///
     /// This allows immediate verification without needing to look up the public key from the validator set.
     async fn verify_validator_proof(


code/crates/core-types/src/validator_proof.rs
```diff
@@ -0,0 +1,58 @@
...
+///
+/// This allows nodes to cryptographically verify that a peer claiming to be
+/// a validator actually controls the corresponding consensus private key.
+///
+/// The proof binds a network ID (peer_id) to a consensus public key,
+/// signed by the corresponding consensus private key. This allows immediate
+/// signature verification without needing to look up the public key from the
+/// validator set.
+#[derive_where(Clone, Debug, PartialEq, Eq)]
+pub struct ValidatorProof<Ctx: Context> {
```
adizere, line 22: Just familiarizing myself with the ADR and this implementation: Do we need to version this data type? Seems to go through the wire unversioned.

ancazamfir: Currently I use the same approach as for sync and consensus, using protocol names that have version in their names. Haven't added it to defaults yet. But if two nodes use different versions, streams will not be established.
Currently hardcoded with:
```
pub const PROTOCOL_NAME: StreamProtocol = StreamProtocol::new("/malachitebft-validator-proof/v1")
```

ancazamfir: cce504e1331081a5d621d0b10c40fa2f3c205a7f

ancazamfir:
For testing I used the example/channel app, init and started nodes wiht [testnet_manager.bash ](https://gist.github.com/ancazamfir/e0f05aed2ae426e81fac0f59399b4c77) script. 
Current settings in the script:
```
    export MALACHITE__CONSENSUS__P2P__PROTOCOL__TYPE="gossipsub"
    export MALACHITE__CONSENSUS__P2P__PROTOCOL__ENABLE_PEER_SCORING="false"
    export MALACHITE__VALIDATOR_ROTATION__ENABLED="true"
    export MALACHITE__VALIDATOR_ROTATION__ROTATION_PERIOD="10"
    export MALACHITE__VALIDATOR_ROTATION__SELECTION_SIZE="3"
    export MALACHITE__CONSENSUS__P2P__PROTOCOL__ENABLE_EXPLICIT_PEERING="false"
    export MALACHITE__CONSENSUS__P2P__PROTOCOL__ENABLE_FLOOD_PUBLISH="true"
```
With these settings the validator set rotates every 10 heights and 3 validators are selected from the validators crates.

Testing uses the metrics to determine node classification, for this PR the `validator` vs `persistent` or `full-node` is of interest. Used [check_mesh_vals.py ](https://gist.github.com/ancazamfir/c52e9076f5b99f088be59e5ebb9ce1ba)to get and process the metrics.

- init and start the nodes
```
./scripts/testnet_manager.bash init --validators 5 --full-nodes 10 --home nodes
./scripts/testnet_manager.bash start --nodes all --home node
```

- check the node types, verify that every 10 heights the nodes are properly (re)classified after validator rotation
``` ./scripts/check_mesh_vals.py
...
Moniker         Type         Cons  Prop  Live  Expl  Peers  InPeers  OutPeers  Conns  InConns  OutConns
-----------------------------------------------------------------------------------------------------------------------------
app-1           validator    11    11    11    0     11     7        4         15     11       4
app-2           validator    10    8     11    0     11     7        4         15     11       4
app-3           validator    6     11    7     0     11     7        4         15     11       4
-----------------------------------------------------------------------------------------------------------------------------
app-0           persistent   11    11    11    0     11     7        4         15     11       4
app-4           persistent   10    7     8     0     11     7        4         15     11       4
-----------------------------------------------------------------------------------------------------------------------------
app-10          full_node    4     4     4     0     5      0        5         5      0        5
app-11          full_node    4     4     4     0     5      0        5         5      0        5
app-5           full_node    4     4     4     0     5      0        5         5      0        5
app-6           full_node    4     4     4     0     5      0        5         5      0        5
app-7           full_node    4     4     4     0     5      0        5         5      0        5
app-8           full_node    4     4     4     0     5      0        5         5      0        5
app-9           full_node    4     4     4     0     5      0        5         5      0        5
```

code/crates/network/src/validator_proof/README.md
```diff
@@ -0,0 +1,467 @@
...
+
+This module implements a one-way protocol that allows validators to prove their identity to peers. When a validator successfully proves their identity, peers may upgrade their GossipSub score, giving priority to validator messages in mesh formation and message propagation. In the future, this may also be used for connection prioritization (e.g., preferring connections to validators when slots are limited).
+
+See ADR-006 (adr-006-proof-of-validator.md) for the design rationale and protocol specification.
+
+## Overview
+
+When peers connect, they don't know if the other peer is a validator. The Identify protocol provides a peer's moniker and listen address, but validator status must be cryptographically proven.
+
+Each validator holds a pre-signed proof containing their consensus public key and libp2p peer ID, signed with their consensus key. Validators send this proof:
```
nenadmilosevic95:
```suggestion
Each validator holds a pre-signed proof containing their consensus public key and libp2p peer ID, signed with their consensus private key. Validators send this proof:
```

code/crates/network/src/validator_proof/README.md
```diff
@@ -0,0 +1,467 @@
...
+
+## Validator Proof Related State
+
+The validator proof state is split between two locations:
+
+**`validator_proof::Behaviour`** (`behaviour.rs`) — connection-scoped session state:
+
+| Field | Type | Purpose |
+|-------|------|---------|
+| `proof_bytes` | `Option<Bytes>` | Our proof to send (set when we're a validator, cleared when not) |
```
nenadmilosevic95: I don't see why we need to clear it. And after looking at the code I don't think we even clear it ever?

code/crates/core-types/src/validator_proof.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// The validator's consensus public key (raw bytes)
+    pub public_key: Vec<u8>,
+    /// The libp2p peer ID bytes
+    pub peer_id: Vec<u8>,
+    /// Signature over (public_key, peer_id) using the validator's consensus key
```
nenadmilosevic95:
```suggestion
    /// Signature over (public_key, peer_id) using the validator's consensus private key
```

code/crates/network/src/validator_proof/types.rs
```diff
@@ -0,0 +1,23 @@
...
+/// Internal verification result.
+///
+/// This is used internally for tracking - the wire protocol is one-way
+/// and doesn't send any response.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[repr(u8)]
+pub enum ProofVerificationResult {
+    /// Proof received and verified successfully.
+    Valid = 0,
+    /// Proof validation failed (decode, peer_id mismatch, or invalid signature).
```
nenadmilosevic95, line 14:
```suggestion
    /// Proof validation failed (peer_id mismatch or invalid signature).
```
right?

code/crates/network/src/validator_proof/behaviour.rs
```diff
@@ -0,0 +1,309 @@
...
+        cx: &mut task::Context<'_>,
+    ) -> Poll<ToSwarm<Self::ToSwarm, libp2p::swarm::THandlerInEvent<Self>>> {
+        // Check for events from protocol tasks
+        if let Poll::Ready(Some(event)) = self.events_rx.poll_recv(cx) {
+            match &event {
+                // On send failure, allow retry by removing from sent set
+                Event::ProofSendFailed { peer, .. } => {
+                    self.proofs_sent.remove(peer);
+                    return Poll::Ready(ToSwarm::GenerateEvent(event));
+                }
```
nenadmilosevic95: Potential edge case: if `write_proof` succeeds but `close()` fails, this allows a retry, but the receiver may already have the proof in `proofs_received` and would disconnect us for anti-spam. Is this scenario even possible, and if so should we handle it?

code/crates/network/src/validator_proof/protocol.rs
```diff
@@ -0,0 +1,86 @@
...
+
+async fn handle_incoming_streams(
+    mut streams: stream::IncomingStreams,
+    events_tx: mpsc::UnboundedSender<Event>,
+) {
+    while let Some((peer, stream)) = streams.next().await {
+        debug!(%peer, "Accepted incoming validator proof stream");
+
+        let events_tx = events_tx.clone();
+        tokio::spawn(async move {
```
nenadmilosevic95, line 39: No timeout on `recv_proof` — if a peer opens a stream but never sends data, the task hangs until the connection closes. Should we add a `tokio::time::timeout` here to clean up stale streams and put some large value?

code/crates/network/src/validator_proof/behaviour.rs
```diff
@@ -0,0 +1,691 @@
...
+    /// Channel for receiving events from protocol tasks.
+    events_rx: mpsc::UnboundedReceiver<Event>,
+    events_tx: mpsc::UnboundedSender<Event>,
+
+    /// Track active connections per peer.
+    /// Only send proof on first connection, only clean up when all connections close.
+    connections: HashMap<PeerId, HashSet<ConnectionId>>,
+
+    /// Track peers we've sent proofs to (to avoid duplicates).
+    proofs_sent: HashSet<PeerId>,
```
nenadmilosevic95: Do we need both `connections` and `proofs_sent`? `send_proof` is only called from `on_connection_established` which already checks first connection via connections, and they're cleaned up together in `on_connection_closed`, so `proofs_sent` is always redundant here.

Also we could use libp2p's `other_established` (0 = first) and `remaining_established` (0 = last) instead of tracking connections ourselves — that would let us drop both and just keep `proofs_received`. WDYT?

code/crates/network/src/validator_proof/behaviour.rs
```diff
@@ -0,0 +1,691 @@
...
+    fn poll(
+        &mut self,
+        cx: &mut task::Context<'_>,
+    ) -> Poll<ToSwarm<Self::ToSwarm, libp2p::swarm::THandlerInEvent<Self>>> {
+        // Check for events from protocol tasks
+        if let Poll::Ready(Some(event)) = self.events_rx.poll_recv(cx) {
+            match &event {
+                // On send failure, allow retry by removing from sent set
+                Event::ProofSendFailed { peer, .. } => {
+                    self.proofs_sent.remove(peer);
```
nenadmilosevic95: This removes from `proofs_sent` to "allow retry" but I don't see the retry. Is there any specific retry mechanism here?

code/crates/network/src/state.rs
```diff
@@ -30,10 +30,24 @@ pub struct NetworkStateDump {
...
     pub address: String,
+    /// Public key bytes (for matching validator proofs)
+    pub public_key: Vec<u8>,
     /// Voting power
     pub voting_power: u64,
 }
 
+impl ValidatorInfo {
+    /// Returns the address if the public key matches, None otherwise.
+    /// Used to look up the address when we only have public key bytes (from proof).
```
nenadmilosevic95:
```suggestion
    /// Returns the address if the public key matches, None otherwise.
```
Nit: I would leave it like this, the second part is confusing

code/crates/network/src/state.rs
```diff
@@ -631,11 +698,269 @@ fn apply_peer_type_change(
...
     peer_info.peer_type = new_type;
     peer_info.score = new_score;
 
+    // Update metrics (marks old stale if labels changed)
     metrics
         .update_peer_labels(peer_id, old_peer_info, peer_info)
         .then_some(new_score)
 }
+
+#[cfg(test)]
```
nenadmilosevic95, line 711: There's no test for the full `pending_verified_proofs → update_peer flow`. Maybe we can add something like this (Claude generated):

```
    #[test]
    fn pending_proof_applied_when_identify_completes() {
        let mut state = test_state();
        let peer_id = libp2p::PeerId::random();
        let public_key = vec![1, 2, 3];

        // Add to validator set
        state.validator_set.insert(ValidatorInfo {
            address: "buffered_val".to_string(),
            public_key: public_key.clone(),
            voting_power: 100,
        });

        // Proof arrives before Identify → should be buffered
        let result = state.record_verified_proof(&peer_id, public_key.clone());
        assert!(result.is_none());
        assert!(state.pending_verified_proofs.contains_key(&peer_id));

        // Simulate Identify completing
        let info = identify::Info {
            public_key: libp2p::identity::Keypair::generate_ed25519().public(),
            protocol_version: "test/1.0".to_string(),
            agent_version: "moniker=test-peer".to_string(),
            listen_addrs: vec!["/ip4/10.0.0.1/tcp/26656".parse().unwrap()],
            protocols: vec![],
            observed_addr: "/ip4/127.0.0.1/tcp/0".parse().unwrap(),
            signed_peer_record: None,
        };
        let conn_id = libp2p::swarm::ConnectionId::new_unchecked(42);
        let score = state.update_peer(peer_id, conn_id, &info);

        // Proof should be applied from buffer
        assert!(state.pending_verified_proofs.is_empty());
        let peer = &state.peer_info[&peer_id];
        assert!(peer.peer_type.is_validator());
        assert_eq!(peer.consensus_public_key.as_deref(), Some(&public_key[..]));
        assert_eq!(peer.consensus_address.as_deref(), Some("buffered_val"));
        assert_eq!(score, VALIDATOR_SCORE);
    }
```

code/crates/network/src/lib.rs
```diff
@@ -375,6 +446,10 @@ async fn run(
     mut rx_ctrl: mpsc::Receiver<CtrlMsg>,
     tx_event: mpsc::Sender<Event>,
 ) {
+    // Note: We don't set the validator proof here at startup.
+    // `is_validator` is initially false, the proof will be set when we receive
+    // the validator set via UpdateValidatorSet and confirm we're in it.
+
```
nenadmilosevic95: Is this comment a leftover?

code/crates/signing/src/ext.rs
```diff
@@ -399,4 +415,26 @@ where
...
+    }
+
+    async fn verify_validator_proof(
+        &self,
+        proof: &ValidatorProof<Ctx>,
+    ) -> Result<VerificationResult, Error> {
+        let signing_bytes = ValidatorProof::<Ctx>::signing_bytes(&proof.public_key, &proof.peer_id);
+        // Decode the public key from the certificate bytes
+        let public_key = Ctx::SigningScheme::decode_public_key(&proof.public_key)
+            .map_err(|e| Error::from_source(format!("Invalid public key in certificate: {e}")))?;
```
nenadmilosevic95:
```suggestion
        // Decode the public key from the proof bytes
        let public_key = Ctx::SigningScheme::decode_public_key(&proof.public_key)
            .map_err(|e| Error::from_source(format!("Invalid public key in validator proof: {e}")))?;
```

BREAKING_CHANGES.md
```diff
@@ -7,9 +7,17 @@
...
 ### `malachitebft-core-types`
 
+- Added new `ValidatorProof<Ctx>` type for the Proof-of-Validator protocol (ADR-006)
 - Added new associated type `Timeouts` to the `Context` trait (use `LinearTimeouts` for default implementation) ([#1227](https://github.com/circlefin/malachite/pull/1227))
 - Remove `initial_validator_set` and `initial_height` fields from `Params` struct ([#1190](https://github.com/circlefin/malachite/pull/1190))
 
+### `malachitebft-signing`
+
+- Added two new methods to `SigningProviderExt` trait for Proof-of-Validator (ADR-006):
+  - `sign_validator_certificate(&self, public_key: Vec<u8>, peer_id: Vec<u8>) -> Result<ValidatorProof<Ctx>, Error>`
```
nenadmilosevic95:
```suggestion
  - `sign_validator_proof(&self, public_key: Vec<u8>, peer_id: Vec<u8>) -> Result<ValidatorProof<Ctx>, Error>`
```

BREAKING_CHANGES.md
```diff
@@ -7,9 +7,17 @@
...
+  - `verify_validator_certificate(&self, certificate: &ValidatorProof<Ctx>) -> Result<VerificationResult, Error>`
```
nenadmilosevic95:
```suggestion
  - `verify_validator_proof(&self, proof: &ValidatorProof<Ctx>) -> Result<VerificationResult, Error>`
```

code/crates/network/src/state.rs
ancazamfir: Will remove, but to clairfy, when a validator proof arrives, it only contains public key bytes. This function checks if those bytes match this ValidatorInfo's public key, and if so returns the address. It's used in `record_verified_proof()` and `reclassify_peers()` to map a proof's public key to a validator's address for display/metrics.

code/crates/network/src/validator_proof/behaviour.rs
ancazamfir: Good ideea, can do this now that we don't send the proof at start (vs when we become a validator) when we don't have connections established. Also forgot completely about `other_established`.

ancazamfir: 148c46d023e4606fb190e930b891528d513bf719

code/crates/network/src/validator_proof/behaviour.rs
ancazamfir: Removed as part of 148c46d023e4606fb190e930b891528d513bf719.
We discussed this during the review that we do not currently retry but we could. Should we?

code/crates/network/src/validator_proof/behaviour.rs
ancazamfir: It is theoretically possible. With the new code the gap shifts though, e.g. if send() fails (or open_stream fails), and peer never gets the proof. We don't retry (and we removed the `proofs_sent` that would be needed for that). So the peer stays connected but we're never classified as a validator by them.

In practice I believe that if open_stream or send fails, the underlying connection is degraded and will likley be dropped, trigger a reconnect and fresh proof exchange. But not guaranteed.

We have these options:

1. Accept the risk, if the send fails, the connection is likely unhealthy anyway and will reset
2. Retry on failure, i.e. on `ProofSendFailed`, schedule a retry (e.g., after a short delay). This adds complexity but closes the gap
3. Re-add `proofs_sent` tracking, so `ProofSendFailed` removes from the set, and we add a periodic "retry pending peers" check. This is back to more state though

I propose we do 1 for now, wdyt? We can do 2 later.

code/crates/network/src/validator_proof/protocol.rs
ancazamfir: There is a timeout inside `read_proof` in codec.rs, see `READ_TIMEOUT`  5 seconds. So stale streams are cleaned up, if a peer opens a stream but never sends, it times out and emits `ProofReceiveFailed`, which disconnects the peer.

code/crates/network/src/validator_proof/types.rs
ancazamfir: yes, that's out of date.
71e351d9e983c4247f79bd15c1df56a9f184c4e2

code/crates/network/src/lib.rs
ancazamfir: fixed in 148c46d023e4606fb190e930b891528d513bf719

code/crates/network/src/state.rs
ancazamfir: 63f7151e6c58fe392299d15d35d72731bf844125

code/crates/network/src/validator_proof/protocol.rs
nenadmilosevic95: Yes, this comment is stale. I think Romain added this while I was reviewing the old version

code/crates/network/src/validator_proof/README.md
ancazamfir: 148c46d023e4606fb190e930b891528d513bf719

code/crates/signing/src/ext.rs
```diff
@@ -102,6 +103,21 @@ where
         validator_set: &Ctx::ValidatorSet,
         thresholds: ThresholdParams,
     ) -> Result<(), CertificateError<Ctx>>;
+
+    /// Sign a validator certificate binding the given public key to the given peer ID.
```
nenadmilosevic95:
```suggestion
    /// Sign a validator proof binding the given public key to the given peer ID.
```

code/crates/signing/src/ext.rs
```diff
@@ -102,6 +103,21 @@ where
...
+    async fn sign_validator_proof(
+        &self,
+        public_key: Vec<u8>,
+        peer_id: Vec<u8>,
+    ) -> Result<ValidatorProof<Ctx>, Error>;
+
+    /// Verify a validator certificate's signature using the public key included in the certificate.
```
nenadmilosevic95:
```suggestion
    /// Verify a validator proof's signature using the public key included in the certificate.
```

code/crates/network/src/validator_proof/behaviour.rs
nenadmilosevic95: Yes, let's do like that, we can track if it is happening and if we need 2. Since we don't use it for any critical tasks for now like msg filtering it should be ok!

code/crates/network/src/validator_proof/behaviour.rs
nenadmilosevic95: Ok, let's skip it for now

PR #1451, opened 2026-02-04
fix(sync): Replace the latency “cliff” in EMA scoring with a continuous curve

romac:
Right now quality drops from 1.0 to e^-1 immediately at the threshold. This commit makes it continuous after the threshold.

**Before:**

```
threshold = 1.00s
  t(s)     quality
--------------------
  0.90    1.000000
  1.00    1.000000
  1.10    0.332871   <-- cliff
  1.20    0.301194
  1.30    0.272532
  1.40    0.246597
  1.50    0.223130
  1.60    0.201897
  1.70    0.182684
  1.80    0.165299
  1.90    0.149569
  2.00    0.135335
  2.10    0.122456
  2.20    0.110803
  2.30    0.100259
  2.40    0.090718
```

**After:**

```
threshold = 1.00s
  t(s)     quality
--------------------
  0.90    1.000000
  1.00    1.000000
  1.10    0.904837   <-- no cliff
  1.20    0.818731
  1.30    0.740818
  1.40    0.670320
  1.50    0.606531
  1.60    0.548812
  1.70    0.496585
  1.80    0.449329
  1.90    0.406570
  2.00    0.367879
  2.10    0.332871
  2.20    0.301194
  2.30    0.272532
  2.40    0.246597
```

### DIFF
--- a/code/crates/sync/src/scoring/ema.rs
+++ b/code/crates/sync/src/scoring/ema.rs
@@ -90,19 +90,17 @@ impl ScoringStrategy for ExponentialMovingAverage {
                     (-(response_time_secs - threshold_secs) / threshold_secs).exp()
                 };
 
-                println!(
-                    "Response time: {:.3}s, Quality: {:.3}, Previous Score: {:.3}",
-                    response_time_secs, quality, previous_score
-                );
-
-                println!(
-                    "       Updating score: alpha_success: {:.3}, New Score: {:.3}",
-                    self.alpha_success,
-                    self.alpha_success * quality + (1.0 - self.alpha_success) * previous_score
-                );
-
                 // Update score with EMA using alpha_success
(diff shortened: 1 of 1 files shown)


code/crates/sync/src/scoring.rs
```diff
@@ -528,7 +527,9 @@ mod tests {
     }
 
     // Property: Score updates should be monotonic for sequences of same result type
+    // Note: This test does not apply to EMA anymore, but it can be useful for testing other strategies that have more deterministic score changes.
     #[test]
+    #[ignore]
```
romac, line 532: One notable difference with the sharp dropoff we had before is that response that is only a little slower than the threshold is still deemed of fairly high quality and might therefore improve a previously bad score.

ancazamfir (approved):
👍

PR #1452, opened 2026-02-04
feat(sync): Add eager status update mode for sync actor

romac:
## Summary

Introduces a new status update mode that sends a status update as soon as a new height is started rather than at fixed intervals.

To enable this mode, set `status_update_interval = 0`.

## Testing

For testing in Arc, use my `eager-sync-updates` branch

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -49,6 +49,9 @@
 - Refactor sync actor to notify consensus of sync responses
 - Support batch retrieval of decided values
 - Validate value request ranges before processing
+- Introduce a new mode that sends a status update as soon as a new height is started rather than at a fixed interval ([#1452](https://github.com/circlefin/malachite/pull/1452))
+  To enable this mode, set `status_update_interval = 0`.
+
 
 ## 0.6.0
 


code/crates/engine/src/sync.rs
```diff
@@ -126,7 +126,13 @@ impl<Ctx: Context> From<TimeoutElapsed<Timeout>> for Msg<Ctx> {
 
 #[derive(Debug)]
 pub struct Params {
+    /// Interval at which to update other peers of our status
+    /// If set to 0s, status updates are sent eagerly right after each decision.
+    /// Default: 5s
```
cason, line 131: 2s?

code/crates/engine/src/sync.rs
```diff
@@ -499,6 +527,33 @@ where
...
+        StatusUpdateMode::OnDecision
+    } else {
+        info!("Using status update mode: Interval");
+
+        // One-time uniform adjustment factor [-1%, +1%]
+        const ADJ_RATE: f64 = 0.01;
+        let adjustment = rng.gen_range(-ADJ_RATE..=ADJ_RATE);
+
+        let ticker = tokio::spawn(
+            ticker(interval, sync.clone(), adjustment, || Msg::Tick).in_current_span(),
```
cason, line 545: Just saw this logic here.

There is no actual randomness, we have a random period, which is `interval +- rand(-adjustment, +adjustment)`, namely very similar to each other.

Plus, the code assume that `sleep` sleeps the exactly amount of time required. Which is not the case. Probably `sleep` has an error of about 1%, as the duration, so....

code/crates/engine/src/sync.rs
romac: The default value on line 136/142 is 5s at the moment. Way too high if you ask me, but that's what it is.

code/crates/engine/src/sync.rs
romac: cf. discussion in the original PR and Jira ticket

code/crates/engine/src/sync.rs
```diff
@@ -425,6 +448,11 @@ where
             Msg::Decided(height) => {
                 self.process_input(&myself, state, sync::Input::Decided(height))
                     .await?;
+
+                // If in OnDecision mode, schedule the next tick with some jitter
+                if let StatusUpdateMode::OnDecision = &state.status_update_mode {
+                    myself.send_after(state.random_jitter(), || Msg::Tick);
```
cason: Do we really need this random jitter?

I have the impression that the decision process, pushing the message to the sync actor, etc. already contains enough "imprecision" (in terms of random timing) to cover it.

And in any case we are broadcasting a small message, I can hardly see a relevant network effect of it.

romac: @ancazamfir Any concerns about removing the jitter here?

ancazamfir: yes we can remove it in this case.

ancazamfir: But not sure if Decided is the best place here. We want to make sure the app has the block and executed the payload. Maybe we bcast in StartHeight with the tip height.

romac: Yeah I was also concerned about that. I thought about StartHeight but figured we might be able to do better if we had a way for the app to acknowledge the Decided message, so that we'd know it has persisted the decision. Right now the reply to Decided moves us to the next height, but perhaps after #1447 (which splits decision and finalization), we might be able to wait for reply on Decided and send then?

In any case, I will move it to StartHeight for now and maybe we can revisit after #1447 is merged.

romac: - https://github.com/circlefin/malachite/pull/1452/commits/0e61c106688e83b4ed69235286ca9a67ae388b3a
- https://github.com/circlefin/malachite/pull/1452/commits/e0939bd24c057f6bd14733870e352d017df04734

code/crates/engine/src/sync.rs
```diff
@@ -440,6 +432,13 @@ where
...
             // (Re)Started a new height
             Msg::StartedHeight(height, restart) => {
+                // If in OnDecision mode, send a status update for the previous decision,
+                // now that we know for sure that the application has stored the decided value.
+                if let StatusUpdateMode::OnDecision = &state.status_update_mode {
+                    self.process_input(&myself, state, sync::Input::SendStatusUpdate)
+                        .await?;
+                }
+
                 self.process_input(&myself, state, sync::Input::StartedHeight(height, restart))
```
cason: Shouldn't we first update our `tip_height` then send the `Status` message? Namely, the if should after this line instead of before?

code/crates/sync/src/handle.rs
```diff
@@ -17,8 +17,8 @@ use crate::{
 
 #[derive_where(Debug)]
 pub enum Input<Ctx: Context> {
-    /// A tick has occurred
-    Tick,
+    /// Scheduled tick for sending status update to peers
```
cason:
```suggestion
    /// Periodical event triggering the broadcast of a status update
```

code/crates/engine/src/sync.rs
romac: Good catch, thanks! 99216502adb5e29ef8dc60ade15ce218c3c21f28

code/crates/engine/src/sync.rs
```diff
@@ -139,6 +145,15 @@ impl Default for Params {
     }
 }
 
+/// The mode for sending status updates
+enum StatusUpdateMode {
+    /// Send status updates at regular intervals
+    Interval(JoinHandle<()>), // the ticker task handle
+
+    /// Send status updates only upon decisions
+    OnDecision,
```
ancazamfir: Is it on StartHeight?
```suggestion
    /// Send status updates with tip height when starting a new height
    OnStartedHeight,
```
If you agree, more related changes are needed

romac: Right, it's a leftover from the initial implementation.

ae6d864d790e87d2d23d24f84fc7483cfb29f758

PR #1453, opened 2026-02-04
fix(test): Fix `equivocation_two_vals_same_pk` integration test being flaky

romac:


cason (commented):
This is still a workaround, although it probably works.

Ideally, we should wait after the consensus instance was concluded before starting a new one. And make sure that equivocations observed during this interval are reported (I am afraid they are not right now...).

sergio-mena (approved):

ancazamfir (approved):

PR #1467, opened 2026-02-09
feat(sync): Queue sync responses for future heights in the Sync actor

romac:
Instead of buffering sync responses in the `core-consensus` input queue, sync responses are buffered directly in the Sync actor. This prevents sync responses and consensus messages from contending over the input queue.

  ## Changes

  1. Add a `BoundedQueue<Height, BufferedValue>` that buffers responses for heights ahead of consensus, draining them when consensus starts a new height. Its capacity is defined as `2 * sync.parallel_requests * sync.batch_size` to allow for extra room.
  2. Each buffered value is tagged with the `OutboundRequestId` that produced it. On `InvalidValue`, all buffered values originating from the same request as the invalid height are removed before re-requesting from a different peer. This prevents stale values from a bad peer from being drained to consensus when the height advances.
  3. The `ProcessSyncResponse` message now carries a single `CoreValueResponse` instead of a `(PeerId, sync::ValueResponse)`. The batch-processing loop and `process_sync_response` helper are removed.
  4. The sync actor is notified *after* consensus enters the `Running` phase (post-WAL replay), so pending sync values are only forwarded when consensus is ready.
  5. The consensus `queue_capacity` is no longer derived from `sync.parallel_requests * sync.batch_size`; it gets a proper default of 10 via `default_queue_capacity()`.
  6. Add `BoundedQueue::retain()` for filtering values by predicate, and `RawDecidedValue::height()` / `to_core()` helpers. The `OutboundRequestId` is threaded through the  `ProcessValueResponse` effect so it is available at the buffering site.

## Metrics

Adds two new metrics:
- `sync_queue_heights`: Number of heights in the sync input queue
- `sync_queue_size`: Total number of inputs across all heights in the sync input queue

## Logging

This PR improves logging in the bounded queue, as the current logs are misleading:

- `Bounded queue is full, dropping value` => `Bounded queue is full, evicting highest index`
  - The original message leads to think that the value was not inserted in the queue, whereas the actual behavior is that the entries with the highest index in the queue are evicted and the value is inserted.
- `Bounded queue is full, no value is inserted` => `Bounded queue is full, rejecting value`
  - Correct, but the new message is clearer imho

--- 

**Story:** https://circlepay.atlassian.net/browse/CCHAIN-1080

### DIFF
--- a/code/crates/discovery/src/handlers/close.rs
+++ b/code/crates/discovery/src/handlers/close.rs
@@ -103,6 +103,9 @@ where
         // Remove signed peer record (no longer connected, record may be stale)
         self.signed_peer_records.remove(&peer_id);
 
+        // Clear rate limiter state for this peer
+        self.rate_limiter.remove_peer(&peer_id);
+
         // Clear connect_request done_on to allow re-upgrading the peer on reconnection
         self.controller.connect_request.remove_done_on(&peer_id);
 
--- a/code/crates/discovery/src/handlers/peers_request.rs
+++ b/code/crates/discovery/src/handlers/peers_request.rs
@@ -29,6 +29,48 @@ where
             || request_data.retry.count() != 0
     }
 
+    /// Check rate limit for an incoming peers request.
+    ///
+    /// Returns `true` if the request should be served, `false` if rate limited.
+    /// When rate limited, logs a warning and disconnects the peer if they've
+    /// exceeded the maximum violation count.
+    fn check_rate_limit(&mut self, swarm: &mut Swarm<C>, peer: &PeerId) -> bool {
+        let result = self.rate_limiter.check_request(peer);
+
+        if result.is_allowed() {
+            return true;
+        }
+
+        let violation_count = self.rate_limiter.violation_count(peer);
+        let should_disconnect = result.should_disconnect();
+
+        warn!(
+            %peer,
+            violation_count,
+            should_disconnect,
+            "Rate limiting: ignoring peers request (exceeded {} requests in {:?})",
+            self.rate_limiter.max_requests_per_window(),
+            self.rate_limiter.rate_window()
+        );
+
+        // TODO: When reputation system is implemented:
+        // - Report ReputationPenalty::ExcessivePeersRequests (-4096) to reputation system
+        // - Reputation system will handle backoff tiers and eventual banning
+        // - Maybe keep the immediate disconnect for max violations as fast path
+
+        if should_disconnect {
+            warn!(
+                %peer,
+                violation_count,
+                max_violations = self.rate_limiter.max_violations(),
+                "Disconnecting peer due to excessive peers request violations"
+            );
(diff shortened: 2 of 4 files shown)


code/crates/engine/src/consensus.rs
```diff
@@ -611,41 +606,25 @@ where
...
-                        // At this point, `process_sync_response` has already sent a message
-                        // about an invalid value, etc. to the sync actor. The sync actor
-                        // will then, re-request this range again from some peer.
-                        // Because of this, in case of failing to process the response, we need
-                        // to exit early this loop to avoid issuing multiple parallel requests
-                        // for the same range of values. There's also no benefit in processing
-                        // the rest of the values.
-                        error!(%start_height, %height, "Failed to process sync response: {e:?}");
-
-                        break;
```
romac, line 650: > [!IMPORTANT]
> The old code would stop processing the batch of synced values if one value in the batch failed validation, while the new code processes all values unconditionally.

I need to think on whether the old behavior was necessary, and if so, how to restore it.

ancazamfir: this was only if the certificate validation was failing i think. If block was invalid (we would find out later), we would still buffer the rest of sync values.

ancazamfir: In the new code, when a batch of sync values arrives, the sync actor immediately buffers values for future heights (H+1, H+2) before the current height (H) is validated. If H's validation fails, sync re-requests the range from a different peer, but the already-buffered H+1, H+2 from the bad peer remain in the queue. 

When the new batch arrives from a good peer, those values are appended to the same height buckets, creating duplicates. Later, when consensus advances to H+1, all buffered values for that height are drained and sent to consensus, both the bad peer's (which may fail validation again) and the good peer's (which will succeed). This is wasteful but not a incorrect. 

One possible fix to deduplicate, would be to remove buffered values from a peer when sync receives InvalidValue. We could buffer the request id:
```rust
struct BufferedValue<Ctx: Context> {
    request_id: OutboundRequestId,
    value: CoreValueResponse<Ctx>,
}

type SyncQueue<Ctx> = BoundedQueue<<Ctx as Context>::Height, BufferedValue<Ctx>>;
```
and, on `InvalidValue`, remove all matching values before re-requesting those heights. Something like:

```rust
Msg::InvalidValue(peer, height) => {
    // Remove buffered values from the same request
    if let Some((request_id, _)) = state.sync.get_request_id_by(height) {
        state.sync_queue.retain(|_, buffered| buffered.request_id != request_id);
        self.metrics.sync_queue_updated(state.sync_queue.len(), state.sync_queue.size());
    }
    self.process_input(&myself, state, sync::Input::InvalidValue(peer, height))
        .await?
}

```

Probably nice to have as old code still had the duplicates issue.

romac: 547c3ab27132561cc8e9ba600d160b0c41221ee1

PR #1473, opened 2026-02-11
feat(deps): Update `libp2p-scatter` to v0.4.0-rc.1.

romac:
Update `libp2p-scatter` to v0.4.0-rc.1.

[📖 Changelog](https://github.com/romac/libp2p-scatter/blob/main/CHANGELOG.md#v040-rc1)

This release of `libp2p-scatter` should significantly improve performance by reusing long-lived bidirectional substreams for message exchange, rather than opening new ones for each message, something we previously tried to address in https://github.com/circlefin/malachite/pull/1117.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -38,6 +38,7 @@
 - Add network metrics for peer identification and tracking
 - Add transport level connection limits
 - Limit the number of peers that can connect from same IP address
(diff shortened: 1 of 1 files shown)


ancazamfir:
> This release of `libp2p-scatter` should significantly improve performance by reusing long-lived bidirectional substreams for message exchange, rather than opening new ones for each message,

Just to clarify, this is for Status messages only, right? And consensus if configured with `PubSubProtocol::Broadcast`?

romac:
Yes, that release note is for libp2p-scatter itself, which we only use in
those cases.

PR #1477, opened 2026-02-16
ci: Add check for verified commits

romac:


cason:
Can you push a non-verified commit to test if the workflow does work?

romac:
> Can you push a non-verified commit to test if the workflow does work?

We can't run workflows that use `pull_request_target` from the branch in which they are modified or added as those run from the main branch. So we need to merge this before we can test it.

adizere (approved):

PR #1481, opened 2026-02-17
fix(driver): schedule timeout upon `CommitCertificate` for invalid value

cason:
Closes: #1482 

Minor fix for the logic commented below, which is not applied if the proposal is not valid, only if not found:

https://github.com/circlefin/malachite/blob/ea0809d45a3cb6f242c359bedf3f3c312b350fdc/code/crates/core-driver/src/mux.rs#L225-L228

Test unit updated in order to fail in the case a `CommitCertificate` is received for an invalid value `v`.

### DIFF
--- a/code/crates/core-driver/src/driver.rs
+++ b/code/crates/core-driver/src/driver.rs
@@ -184,6 +184,18 @@ where
         &mut self.vote_keeper
     }
 
+    /// Return precommits for the given round and value from the vote keeper
+    pub fn restore_precommits(
+        &self,
+        round: Round,
+        value_id: &ValueId<Ctx>,
+    ) -> Vec<SignedVote<Ctx>> {
+        self.vote_keeper
+            .per_round(round)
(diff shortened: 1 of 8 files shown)


code/crates/core-driver/src/driver.rs
```diff
@@ -436,7 +436,7 @@ where
 
         match self.store_and_multiplex_commit_certificate(certificate) {
             Some(round_input) => self.apply_input(round, round_input),
-            None => Ok(None),
+            None => Ok(None), // FIXME: Any scenario where this can happen?
```
romac: I think we can now change the return type of `store_and_multiplex_commit_certificate` to `RoundInput<Ctx>` and simplify this.

cason: [ec385bb](https://github.com/circlefin/malachite/pull/1481/commits/ec385bb45f6c27668393bba6568abc57c237cf63)

romac (approved):

PR #1491, opened 2026-02-20
feat(CCHAIN-1105): Add GossipSub explicit peering 

ancazamfir:
Closes: #XXX

### Summary

Add support for GossipSub explicit peering to ensure reliable message delivery between validators, bypassing mesh formation dynamics.

- Add `enable_explicit_peering` and `enable_flood_publish` configuration options to GossipSubConfig
- When explicit peering is enabled, persistent peers are automatically added as explicit peers in GossipSub
- Add `explicit_peers` metric to track active explicit peer connections

GossipSub mesh formation is probabilistic and can result in validators not being directly meshed with each other, requiring messages to traverse through full nodes or rely on IHAVE/IWANT gossip.

### Configuration

Enable via environment variable:

```bash
MALACHITE__CONSENSUS__P2P__PROTOCOL__ENABLE_EXPLICIT_PEERING=true
MALACHITE__CONSENSUS__P2P__PROTOCOL__ENABLE_FLOOD_PUBLISH=true
```

Or in config.toml:

```toml
[consensus.p2p.protocol]
enable_explicit_peering = true
enable_flood_publish = true
```

### Behavior
When `explicit_peering` is enabled, gossipsub publishes and forwards all messages to all persistent peers.
When  `flood_publish` is enabled, gossipsub publishes all messages to all connected peers. The forwarder behavior doesn't change, it still forwards to mesh peers. 

One thing to note is that libp2p gossipsub [assumes symmetric explicit peering.](https://github.com/libp2p/rust-libp2p/blob/libp2p-gossipsub-v0.49.0/protocols/gossipsub/src/behaviour.rs#L1394-L1398)

In detail, let's assume full node F has validator V in persistent_peers but V does NOT have F as persistent. This is what happens:

When F identifies V, F calls gossipsub.add_explicit_peer(V). So V is in F's explicit_peers set.
V does NOT have F in persistent_peers. V never calls add_explicit_peer(F).
V's heartbeat sees mesh=0 (below mesh_n_low=4). V picks F as a candidate to graft (F is connected, subscribed to /consensus, not explicit from V's perspective). V sends GRAFT to F.
F's gossipsub receives the GRAFT and runs handle_graft(). First thing it checks (see above code snippet):
   if self.explicit_peers.contains(peer_id) {  // peer_id = V
       // "ignoring request from direct peer"
       // send PRUNE for all topics
   }
V is in F's explicit_peers → F sends PRUNE back to V.
V receives the PRUNE and F enters 60-second backoff on V's side.
Next heartbeat, V tries another full node, same result. In my setup with 5 validators and 10 full nodes, eventually all full nodes are backed off. After 60s they expire, cycle repeats.

This behavior was one of the reasons to allow flood_publish enabled.

## Testing

- [x] Start testnet with explicit peering enabled
- [x] Verify validators have zero mesh peers but consensus progresses
- [x] Check that `explicit_peers` metric shows expected connections
- [x] Check duplicates
    In a 20 node fully meshed network:
        - explicit peering only: `Avg per node: 172 unique, 3018 duplicates (94% duplicate rate)`
        - flood publish only:   `Avg per node: 172 unique, 1088 duplicates (86% duplicate rate)`
---

#### Contribution eligibility

#### For all contributors

### DIFF
--- a/code/crates/network/src/state.rs
+++ b/code/crates/network/src/state.rs
@@ -107,7 +107,7 @@ impl PeerInfo {
         } else {
             &self.consensus_address
         };
-        let explicit = if self.is_explicit { "explicit" } else { "" };
+        let explicit = if self.is_explicit { "explicit" } else { "-" };
         format!(
             "{}, {}, {}, {}, {}, {}, {}, {}, {}",
             self.address,
@@ -462,7 +462,7 @@ impl State {
     }
 
     /// Format the peer information for logging (scrapable format):
-    ///  Address, Moniker, PeerId, Mesh, Dir, Type, Score
+    ///  Address, Moniker, Type, PeerId, ConsensusAddr, Mesh, Dir, Score, Explicit
     pub fn format_peer_info(&self) -> String {
         let mut lines = Vec::new();
 


code/crates/network/src/lib.rs
```diff
@@ -609,6 +613,37 @@ fn set_peer_score(swarm: &mut swarm::Swarm<Behaviour>, peer_id: libp2p::PeerId,
...
+    swarm: &mut swarm::Swarm<Behaviour>,
+    state: &State,
+    peer_id: libp2p::PeerId,
+    connection_id: libp2p::swarm::ConnectionId,
+    info: &identify::Info,
+    enable_explicit_peering: bool,
+) {
+    if !enable_explicit_peering {
+        return;
+    }
```
romac: Let's move this check to the call site?

ancazamfir: 13d40a04fae39645f000a6968d076a06bed509da

code/crates/config/src/lib.rs
```diff
@@ -298,12 +298,22 @@ pub struct GossipSubConfig {
 
     /// Enable peer scoring to prioritize nodes based on their type in mesh formation
     enable_peer_scoring: bool,
+
+    /// Enable explicit peering for persistent peers.
+    /// When enabled, persistent peers are added as explicit peers in GossipSub,
+    /// ensuring guaranteed message delivery outside the mesh.
+    /// This eliminates mesh partitioning and backoff issues.
+    enable_explicit_peering: bool,
```
nenadmilosevic95, line 305: Should we make a comment somewhere what explicit peering means? also for flood publish?

code/crates/network/src/behaviour.rs
```diff
@@ -168,6 +168,19 @@ fn message_id(message: &gossipsub::Message) -> gossipsub::MessageId {
 }
 
 fn gossipsub_config(config: GossipSubConfig, max_transmit_size: usize) -> gossipsub::Config {
+    // When explicit peering is enabled, flood_publish is forced to false since explicit peers
+    // handle peer-to-peer communication. Otherwise use the configured flood_publish value.
+    let flood_publish = if config.enable_explicit_peering {
```
nenadmilosevic95: maybe we can move this check to `GossipSubConfig::adjust()` ? Also I am double thinking do we need to force this, what will be the consequence of not having this check? I guess the only difference would be that when broadcasting a msg the node will send also directly to peers that are not persistent(explicit peers) but to which it has a connection, basically to nodes that connected to this node?

code/crates/network/src/lib.rs
```diff
@@ -688,6 +718,17 @@ async fn handle_swarm_event(
                 .handle_closed_connection(swarm, peer_id, connection_id);
 
             if num_established == 0 {
+                // Mark explicit peer as stale if this peer was one
+                if config.gossipsub.enable_explicit_peering {
```
nenadmilosevic95, line 743: why we don't remove explicit peer here?

code/crates/network/src/lib.rs
```diff
@@ -609,6 +613,32 @@ fn set_peer_score(swarm: &mut swarm::Swarm<Behaviour>, peer_id: libp2p::PeerId,
     }
 }
 
+/// Add a persistent peer as an explicit peer in gossipsub (if explicit peering is enabled).
+/// Explicit peers receive messages unconditionally, outside the mesh.
+/// This ensures reliable validator-to-validator communication.
+fn add_explicit_peer_to_gossipsub(
```
nenadmilosevic95, line 620: Small suggestion: `add_explicit_peer_to_gossipsub` takes `connection_id` and `info` just to call `state.peer_type()`, but by the time it's called, `state.update_peer()` has already run and populated `state.peer_info`. So the function could just read from `state.peer_info` directly:


```rust
fn add_explicit_peer_to_gossipsub(
    swarm: &mut swarm::Swarm<Behaviour>,
    state: &State,
    peer_id: libp2p::PeerId,
) {
    let Some(peer_info) = state.peer_info.get(&peer_id) else {
        return;
    };

    if peer_info.peer_type.is_persistent() {
        if let Some(gossipsub) = swarm.behaviour_mut().gossipsub.as_mut() {
            gossipsub.add_explicit_peer(&peer_id);
            state.metrics.record_explicit_peer(&peer_id, &peer_info.moniker);
            info!("Added persistent peer {peer_id} as explicit peer in gossipsub");
        }
    }
}
```

code/crates/network/src/metrics.rs
```diff
@@ -77,6 +84,8 @@ pub(crate) struct Metrics {
     discovered_peers: Family<PeerInfoLabels, Gauge>,
     /// Per-peer, per-topic mesh membership (1 = in mesh, 0 = not in mesh)
     peer_mesh_membership: Family<MeshMembershipLabels, Gauge>,
+    /// Explicit peers in gossipsub (1 = active, i64::MIN = disconnected/stale)
```
nenadmilosevic95, line 96: Nit: why not 0 here instead of i64::MIN?

nenadmilosevic95 (approved):
LGTM! Left few minor comments.

code/crates/network/src/behaviour.rs
ancazamfir: > maybe we can move this check to GossipSubConfig::adjust() ?

Yes, the only reason I had to keep it here was for the `warn!` as config create doesn't have tracing dep. But can move it in `adjust()`

ancazamfir: > Also I am double thinking do we need to force this, what will be the consequence of not having this check? I guess the only difference would be that when broadcasting a msg the node will send also directly to peers that are not persistent(explicit peers) but to which it has a connection, basically to nodes that connected to this node?

Agreed we should remove the restriction. Also, maybe consider explicit peering to all connected peers (not just explicit peers)? wdyt?

nenadmilosevic95: @ancazamfir and I discussed this in a meeting and have decided to leave it as it is for the moment, namely explicit peering only for persistent peers. We'll revisit this if needed in the future.

ancazamfir:
While testing I found that libp2p gossipsub [assumes symmetric explicit peering.](https://github.com/libp2p/rust-libp2p/blob/libp2p-gossipsub-v0.49.0/protocols/gossipsub/src/behaviour.rs#L1394-L1398)

In detail, let's assume full node `F` has validator  `V` in `persistent_peers` but `V` does NOT have `F` as persistent. This is what happens:

1. When `F` identifies `V`, `F` calls `gossipsub.add_explicit_peer(V)`. So `V` is in `F`'s explicit_peers set.
2. `V` does NOT have `F` in persistent_peers. `V` never calls `add_explicit_peer(F)`. 
3. `V`'s heartbeat sees `mesh=0` (below `mesh_n_low=4`). `V` picks `F` as a candidate to graft (`F` is connected, subscribed to /consensus, not explicit from `V`'s perspective). `V` sends `GRAFT` to `F`.
4. `F`'s gossipsub receives the `GRAFT` and runs `handle_graft()`. First thing it checks (see above code snippet):
```
   if self.explicit_peers.contains(peer_id) {  // peer_id = V
       // "ignoring request from direct peer"
       // send PRUNE for all topics
   }
```

5. `V` is in `F`'s explicit_peers → `F` sends `PRUNE` back to `V`.
6. `V` receives the `PRUNE` and `F` enters 60-second backoff on `V`'s side.

Next heartbeat, V tries another full node, same result. In my setup with 5 validators and 10 full nodes, eventually all full nodes are backed off. After 60s they expire, cycle repeats.

This behavior gives one more reason to allow flood_publish enabled.

code/crates/config/src/lib.rs
ancazamfir: 7bf173aeb2aed7f50b224410a9161a5a5fed09c7

code/crates/network/src/lib.rs
ancazamfir: good catch, completely missed this.
7bf173aeb2aed7f50b224410a9161a5a5fed09c7

code/crates/network/src/lib.rs
ancazamfir: 7bf173aeb2aed7f50b224410a9161a5a5fed09c7

code/crates/network/src/metrics.rs
ancazamfir: in general we use `i64:MIN` for stale metrics.

nenadmilosevic95:
> While testing I found that libp2p gossipsub [assumes symmetric explicit peering.](https://github.com/libp2p/rust-libp2p/blob/libp2p-gossipsub-v0.49.0/protocols/gossipsub/src/behaviour.rs#L1394-L1398)
> 
> In detail, let's assume full node `F` has validator `V` in `persistent_peers` but `V` does NOT have `F` as persistent. This is what happens:
> 
> 1. When `F` identifies `V`, `F` calls `gossipsub.add_explicit_peer(V)`. So `V` is in `F`'s explicit_peers set.
> 2. `V` does NOT have `F` in persistent_peers. `V` never calls `add_explicit_peer(F)`.
> 3. `V`'s heartbeat sees `mesh=0` (below `mesh_n_low=4`). `V` picks `F` as a candidate to graft (`F` is connected, subscribed to /consensus, not explicit from `V`'s perspective). `V` sends `GRAFT` to `F`.
> 4. `F`'s gossipsub receives the `GRAFT` and runs `handle_graft()`. First thing it checks (see above code snippet):
> 
> ```
>    if self.explicit_peers.contains(peer_id) {  // peer_id = V
>        // "ignoring request from direct peer"
>        // send PRUNE for all topics
>    }
> ```
> 
> 5. `V` is in `F`'s explicit_peers → `F` sends `PRUNE` back to `V`.
> 6. `V` receives the `PRUNE` and `F` enters 60-second backoff on `V`'s side.
> 
> Next heartbeat, V tries another full node, same result. In my setup with 5 validators and 10 full nodes, eventually all full nodes are backed off. After 60s they expire, cycle repeats.
> 
> This behavior gives one more reason to allow flood_publish enabled.

Very important finding! Thanks for doing this! And yes we should definitely allow `flood_publish`

ancazamfir:
> Very important finding! Thanks for doing this! And yes we should definitely allow `flood_publish`

Also we should consider adding option to enable explicit peering for all connected peers (not only persistent). This pretty much nicks out gossipsub mesh. No need to enable `flood_publish`. We could have something like "none" (disabled), "persistent", "all" for `explicit_peering`

code/crates/network/src/behaviour.rs
ancazamfir: 6549596259a5d6010a37bdb5a57bc644040a85ca

code/crates/config/src/lib.rs
```diff
@@ -349,6 +362,10 @@ impl GossipSubConfig {
         {
             self.mesh_outbound_min = max(1, min(self.mesh_n / 2, self.mesh_n_low - 1));
         }
+
+        // Both flood_publish and explicit_peering can be enabled together.
+        // flood_publish sends to all known peers on publish, explicit peering ensures
```
nenadmilosevic95, line 368: Nitpick: explicit peering description doesn't seem precise enough to me, I would rather do it like this - https://github.com/nenadmilosevic95/malachite/commit/28f444bc4b5f5f8471b6ae79223a7ec2d3f5b00c

ancazamfir: 28f444bc4b5f5f8471b6ae79223a7ec2d3f5b00c

code/crates/network/src/state.rs
```diff
@@ -106,16 +107,18 @@ impl PeerInfo {
         } else {
             &self.consensus_address
         };
+        let explicit = if self.is_explicit { "explicit" } else { "" };
```
nenadmilosevic95: Minor:
```suggestion
        let explicit = if self.is_explicit { "explicit" } else { "-" };
```

code/crates/network/src/state.rs
nenadmilosevic95: Claude caught that this comment is not up to date.

code/crates/network/src/lib.rs
```diff
@@ -611,6 +615,51 @@ fn set_peer_score(swarm: &mut swarm::Swarm<Behaviour>, peer_id: libp2p::PeerId,
...
+    };
+
+    if peer_info.peer_type.is_persistent() {
+        if let Some(gossipsub) = swarm.behaviour_mut().gossipsub.as_mut() {
+            gossipsub.add_explicit_peer(&peer_id);
+            state
+                .metrics
+                .record_explicit_peer(&peer_id, &peer_info.moniker);
+            peer_info.is_explicit = true;
+            info!("Added persistent peer {peer_id} as explicit peer in gossipsub");
```
nenadmilosevic95, line 636: Super minor note: this will log twice if both sides dial each other simultaneously? I think it is fine to leave it just making a note

code/crates/network/src/state.rs
ancazamfir: 2d01ea35c163e04bfe02913b8e544aaf17c1f7b1

code/crates/network/src/state.rs
ancazamfir: 2d01ea35c163e04bfe02913b8e544aaf17c1f7b1

code/crates/network/src/metrics.rs
```diff
@@ -222,6 +239,24 @@ impl Metrics {
...
+        self.explicit_peers.get_or_create(&labels).set(1);
+    }
+
+    /// Mark an explicit peer as stale (disconnected)
+    pub(crate) fn mark_explicit_peer_stale(&self, peer_id: &PeerId, moniker: &str) {
+        let labels = ExplicitPeerLabels {
+            peer_id: peer_id.to_string(),
+            peer_moniker: moniker.to_string(),
+        };
+        self.explicit_peers.get_or_create(&labels).set(i64::MIN);
```
romac, line 257: Just curious if there is any specific reason for not using zero here?

romac: Ah just saw your reply there: https://github.com/circlefin/malachite/pull/1491#discussion_r2852276816

PR #1497, opened 2026-02-23
fix(vote-keeper): Filter duplicate votes

ancazamfir:
Closes: #XXX

---

#### Contribution eligibility

#### For all contributors

### DIFF
--- a/code/crates/core-driver/tests/it/extra.rs
+++ b/code/crates/core-driver/tests/it/extra.rs
@@ -3727,13 +3727,7 @@ fn commit_certificate_from_driver_verifies_after_reapplied_votes_from_round_cert
         },
         TestStep {
             desc: "We receive the proposal (from v1)",
-            input: proposal_input(
-                Round::new(0),
-                value.clone(),
-                Round::Nil,
-                Validity::Valid,
-                validators[0].address,
-            ),
+            input: proposal_input_from_proposal(proposal.clone(), Validity::Valid),
             expected_outputs: vec![prevote_output(Round::new(0), value.clone(), &my_addr)],
             expected_round: Round::new(0),
             new_state: prevote_state(Round::new(0)),
@@ -3777,17 +3771,6 @@ fn commit_certificate_from_driver_verifies_after_reapplied_votes_from_round_cert
         TestStep {
             desc: "We receive v1's precommit for value",
             input: Input::Vote(precommit_v1.clone()),
-            expected_outputs: vec![],
-            expected_round: Round::new(0),
-            new_state: precommit_state_with_proposal_and_locked_and_valid(
-                Round::new(0),
-                proposal.clone(),
-            ),
-        },
-        // v4 prevotes nil
-        TestStep {
-            desc: "We receive v4's prevote for nil",
-            input: prevote_nil_input_at(round, &validators[3].address),
             expected_outputs: vec![],
             expected_round: Round::new(0),
             new_state: precommit_state_with_proposal_and_locked_and_valid(
@@ -3872,7 +3855,7 @@ fn commit_certificate_from_driver_verifies_after_reapplied_votes_from_round_cert
 
     assert!(
         result.is_ok(),
-        "commit certificate built from driver vote keeper after re-applied votes must verify (no DuplicateVote): {:?}",
+        "commit certificate built from driver vote keeper must verify: {:?}",
         result.err()
     );
 }


cason:
Another example of https://circlepay.atlassian.net/browse/CCHAIN-889

code/crates/core-driver/tests/it/extra.rs
```diff
@@ -3646,6 +3650,233 @@ fn sync_decision_certificate_then_proposal() {
...
+            input: prevote_input_at(round, value.clone(), &validators[2].address),
+            expected_outputs: vec![precommit_output(Round::new(0), value.clone(), &my_addr)],
+            expected_round: Round::new(0),
+            new_state: precommit_state_with_proposal_and_locked_and_valid(
+                Round::new(0),
+                proposal.clone(),
+            ),
+        },
+        TestStep {
+            desc: "We (v2) get our precommit for value",
```
cason, line 3761: Is this needed?

cason: Apparently, it is. Is `core-consensus` that applies the outputs that the driver itself produces?

code/crates/core-driver/src/driver.rs
```diff
@@ -184,6 +184,27 @@ where
         &mut self.vote_keeper
     }
 
+    /// Return precommits for the given round and value from the vote keeper
+    pub fn restore_precommits(
```
cason, line 188: Can you explain why duplicates are filtered out here, but not in the existing method?

code/crates/core-driver/tests/it/extra.rs
```diff
@@ -3646,6 +3650,233 @@ fn sync_decision_certificate_then_proposal() {
...
+        TestStep {
+            desc: "Start round 0, we (v2) are not the proposer",
+            input: new_round_input(Round::new(0), validators[0].address),
+            expected_outputs: vec![start_propose_timer_output(Round::new(0))],
+            expected_round: Round::new(0),
+            new_state: propose_state(Round::new(0)),
+        },
+        TestStep {
+            desc: "We receive the proposal (from v1)",
+            input: proposal_input(
```
cason:
```suggestion
            input: proposal_input_from_proposal(proposal.clone(), Validity::Valid),
```

cason: Then you can remove all content til line 3737.

code/crates/core-votekeeper/src/keeper.rs
```diff
@@ -104,6 +104,8 @@ where
                     conflicting: vote,
                 });
             }
+            // Do not add duplicate vote
```
cason, line 107: How is it possible that the vote keeper test never caught that?

code/crates/core-driver/src/driver.rs
cason: Ok, the fix is in the vote keeper.

romac (approved):

code/crates/core-consensus/src/handle/decide.rs
```diff
@@ -43,16 +43,17 @@ where
...
-            certificate.clone(),
-            state.driver.validator_set().clone(),
-            state.params.threshold_params,
-        )
-        .await?
-        .is_ok(),
-        "Decide: Commit certificate is not valid"
+        result.is_ok(),
+        "Decide: Commit certificate is not valid: {:?}",
+        result.err()
```
romac: Nitpick: for a slightly cleaner error message (without the `Option` wrapping the error returned by `err()`):

```suggestion
    let result = verify_commit_certificate(
        co,
        certificate.clone(),
        state.driver.validator_set().clone(),
        state.params.threshold_params,
    )
    .await?;
    
    if let Err(e) = result {
        panic!("Decide: Commit certificate is not valid: {e:?}");
     }
```

Note: this patch above won't apply cleanly, need to apply it manually.

code/crates/core-driver/tests/it/extra.rs
```diff
@@ -3646,6 +3650,233 @@ fn sync_decision_certificate_then_proposal() {
...
+            desc: "We receive v1's precommit for value",
+            input: Input::Vote(precommit_v1.clone()),
+            expected_outputs: vec![],
+            expected_round: Round::new(0),
+            new_state: precommit_state_with_proposal_and_locked_and_valid(
+                Round::new(0),
+                proposal.clone(),
+            ),
+        },
+        // v4 prevotes nil
```
cason: This step can be removed

code/crates/core-driver/tests/it/extra.rs
```diff
@@ -3646,6 +3650,233 @@ fn sync_decision_certificate_then_proposal() {
...
+    let result = block_on(signers[0].verify_commit_certificate(
+        &ctx,
+        &certificate,
+        &validator_set,
+        ThresholdParams::default(),
+    ));
+
+    assert!(
+        result.is_ok(),
+        "commit certificate built from driver vote keeper after re-applied votes must verify (no DuplicateVote): {:?}",
```
cason:
```suggestion
        "commit certificate built from driver vote keeper must verify: {:?}",
```

cason: The error is a bit blah.

ancazamfir:
> Another example of https://circlepay.atlassian.net/browse/CCHAIN-889

Couldn't find where this is covered there, could you point out? 
Here we receive a round certificate (Precommit) and we process votes one by one and this results (later) in building a commit certificate with duplicate votes... as were not filtering dups. We do not store the received cert, we build our own when we move to next round.

code/crates/core-votekeeper/src/keeper.rs
cason: Because we don't build certificates there....

cason (commented):
Thank you for that :)

cason:
Suggestion, change the title to: `fix(vote-keeper): Do not store duplicate votes`.

Since we do filter then, in the sense that we don't count their voting power twice.

cason:
> > Another example of https://circlepay.atlassian.net/browse/CCHAIN-889
> 
> Couldn't find where this is covered there, could you point out? Here we receive a round certificate (Precommit) and we process votes one by one and this results (later) in building a commit certificate with duplicate votes... as were not filtering dups. We do not store the received cert, we build our own when we move to next round.

Certificates should be produced by the vote keeper. In this case, we would need to cover this procedure with test units and would not end up in this scenario. Notice, again, that we filter out duplicate votes, namely we ignore the associate voting power, but we do store them because at some point we introduce this because we needed to deal with equivocations.

code/crates/core-consensus/src/handle/decide.rs
ancazamfir: c3f177cf8bc65a095c1561d96436d5feb3a32562

code/crates/core-driver/tests/it/extra.rs
ancazamfir: Not sure i understand, we have to create the proposal someplace.

ancazamfir: ah sorry, yes we have set it up earlier.

ancazamfir: 8880de39162cd7c8206288d8bf9204f4a3cedd8e

code/crates/core-driver/tests/it/extra.rs
ancazamfir: 8880de39162cd7c8206288d8bf9204f4a3cedd8e

code/crates/core-driver/tests/it/extra.rs
ancazamfir: 8880de39162cd7c8206288d8bf9204f4a3cedd8e

ancazamfir:
> Since we do filter then, in the sense that we don't count their voting power twice.

Even before this fix we were not counting the power twice, the problem was that we were storing dups in `received_votes`

ancazamfir:
> but we do store them because at some point we introduce this because we needed to deal with equivocations.

just to clarify, in main, we do not store equivocating votes in`received_votes`, and with this PR now we also don't store duplicate votes.

PR #1498, opened 2026-02-24
feat: Stop `Node` actor on any child failure/termination

romac:


romac:
Development of this PR has moved to a different repository.

Issue #1500, opened 2026-02-24
fix(driver): Prevent scheduling identical timeouts


cason:
In some cases, due to the Mux logic, the same timeout, in particular `TimeoutPrecommit` is scheduled several times:

https://github.com/circlefin/malachite/blob/8d8c224c53e095f1f10ab575e0711ecc8168bf51/code/crates/core-driver/tests/it/extra.rs#L2970-L2986

Maybe the simples solution is to just keep track of scheduled timeouts and drop duplicated `ScheduleTimeout` outputs.

PR #1501, opened 2026-02-24
feat(driver): do not schedule the same timeout twice

cason:
Closes: #1500

As per title.

This can happen to several reasons, but mostly because `TimeoutPrecommit` is re-scheduled at every round step change.

The solution consists on creating a `scheduled_timeouts` vector at the driver, and only adding a new timeout to it if not already present. The vector is cleaned upon a new height, and old entries are removed upon a new height.

### DIFF
--- a/code/crates/core-driver/src/driver.rs
+++ b/code/crates/core-driver/src/driver.rs
@@ -463,11 +463,8 @@ where
         }
 
         let round = certificate.round;
-
-        match self.store_and_multiplex_commit_certificate(certificate) {
-            Some(round_input) => self.apply_input(round, round_input),
-            None => Ok(None),
-        }
+        let round_input = self.store_and_multiplex_commit_certificate(certificate);
+        self.apply_input(round, round_input)
     }
 
     fn apply_polka_certificate(
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -24,8 +24,8 @@ jobs:
       code: ${{ steps.filter.outputs.code }}
       specs: ${{ steps.filter.outputs.specs }}
     steps:
-      - uses: actions/checkout@v4
-      - uses: dorny/paths-filter@v3
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: step-security/paths-filter@6eee183b0d2fd101d3f8ee2935c127bca14c5625 # v3.0.5
         id: filter
         with:
           filters: |
@@ -152,12 +152,12 @@ jobs:
       - name: Create coverage directory
         run: mkdir -p code/coverage
       - name: Download integration coverage
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
         with:
           name: integration-coverage
           path: code/coverage
       # - name: Download MBT coverage
-      #   uses: actions/download-artifact@v4
+      #   uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
       #   with:
       #     name: mbt-coverage
       #     path: code/coverage
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,8 +21,8 @@ jobs:
     outputs:
       code: ${{ steps.filter.outputs.code }}
     steps:
-      - uses: actions/checkout@v4
-      - uses: dorny/paths-filter@v3
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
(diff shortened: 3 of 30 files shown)


code/crates/core-driver/src/driver.rs
```diff
@@ -370,17 +375,29 @@ where
...
 
             RoundOutput::Decision(round, proposal) => outputs.push(Output::Decide(round, proposal)),
         }
     }
 
+    fn lift_timeout_output(&mut self, timeout: Timeout, outputs: &mut Vec<Output<Ctx>>) {
+        if timeout.round < self.round() || self.scheduled_timeouts.contains(&timeout) {
+            return;
+        }
+        // XXX: test if the driver produces **non**-consensus timeouts
```
cason: Should I change this to a warning or something like that? Apparently, we don't panic - at least in the test that have ran.

code/crates/core-driver/src/driver.rs
```diff
@@ -67,6 +67,8 @@ where
 
     /// The certificate that justifies moving to the `enter_round` specified in the `EnterRoundCertificate.
     pub round_certificate: Option<EnterRoundCertificate<Ctx>>,
+
+    scheduled_timeouts: Vec<Timeout>,
```
romac: Let's add a doc comment here to explain this field's purpose

code/crates/core-driver/src/driver.rs
```diff
@@ -132,6 +135,8 @@ where
         // Reset additional internal state
         self.last_prevote = None;
         self.last_precommit = None;
+
+        self.scheduled_timeouts = vec![];
```
romac: To re-use the existing allocation:

```suggestion
        self.scheduled_timeouts.clear();
```

code/crates/core-types/src/timeout.rs
```diff
@@ -39,6 +39,14 @@ impl Timeout {
         Self { round, kind }
     }
 
+    /// Check if this timeout is for a consensus step (propose, prevote, or precommit).
+    pub fn is_consensus(&self) -> bool {
+        match self.kind {
+            TimeoutKind::Propose | TimeoutKind::Prevote | TimeoutKind::Precommit => true,
+            _ => false,
+        }
```
romac:
```suggestion
        matches!(self.kind,
            TimeoutKind::Propose | TimeoutKind::Prevote | TimeoutKind::Precommit)
```

code/crates/core-driver/src/driver.rs
romac: I would perhaps change this to a `debug_assert!`

code/crates/core-driver/src/driver.rs
```diff
@@ -370,17 +376,29 @@ where
...
+                self.lift_timeout_output(timeout, outputs);
                 outputs.push(Output::GetValue(height, round, timeout));
             }
 
             RoundOutput::Decision(round, proposal) => outputs.push(Output::Decide(round, proposal)),
         }
     }
 
+    fn lift_timeout_output(&mut self, timeout: Timeout, outputs: &mut Vec<Output<Ctx>>) {
+        if timeout.round < self.round() || self.scheduled_timeouts.contains(&timeout) {
```
nenadmilosevic95, line 391: Minor: Can this ever be true `timeout.round < self.round()`? Timeouts should only be for the current round. If purely defensive, I'd add a `warn!` here so we notice if it ever triggers.

code/crates/core-driver/src/driver.rs
```diff
@@ -482,6 +500,10 @@ where
         // Update the proposer for the new round
         self.proposer = Some(proposer);
 
+        // Remove useless timeouts from previous rounds
+        self.scheduled_timeouts
+            .retain(|timeout| timeout.round >= round);
```
nenadmilosevic95, line 505: Similar here, could this just be clear()? It is minor ofc, I just want to understand if we expect to see any future rounds here, since I don't see how

code/crates/core-driver/src/driver.rs
cason: Good point.

code/crates/core-driver/src/driver.rs
cason: Also a good point. But the concern here is to limit the growth of this set.

cason: I wonder if we can schedule timeouts for future rounds - otherwise, `clear()` is the way to go.

cason:
> IMHO the state machine check might be even cleaner, since the Tendermint pseudo-code states "for the first time" and the state machine should align with it, but this works well too.

I will try to implement this version.

cason:
> > IMHO the state machine check might be even cleaner, since the Tendermint pseudo-code states "for the first time" and the state machine should align with it, but this works well too.
> 
> I will try to implement this version.

https://github.com/circlefin/malachite/pull/1508

romac:
Superseded by #1508

PR #1508, opened 2026-02-27
feat(smr): do not schedule the same timeout twice

cason:
Closes: #1500

Alternative to https://github.com/circlefin/malachite/pull/1501.

Creates a `scheduled_timeouts` boolean array in the state machine `State`, which is reset to `false` at every round. Before scheduling a timeout for a given `TimeoutKind`, checks if the corresponding slot of the array is taken. If not, mark is as taken an return the timeout as output. Otherwise, don't produce a duplicate `Timeout` event.

In other to prevent tests from breaking, `scheduled_timeouts` is not considered for `PartialEq` comparisons.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -27,7 +27,7 @@ jobs:
       pull-requests: write
     steps:
       - name: Check signatures and cleanup comments
-        uses: actions/github-script@v7
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
         with:
           script: |
             // 1. Fetch all commits in the PR (handling pagination)
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,6 +10,8 @@ jobs:
   publish:
     name: Publish to crates.io
     runs-on: github-hosted-large
+    permissions:
+      id-token: write
     steps:
       - name: Checkout
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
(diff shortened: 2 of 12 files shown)


code/crates/core-state-machine/src/lib.rs
```diff
@@ -8,8 +8,6 @@
     rustdoc::private_intra_doc_links,
     variant_size_differences
 )]
-// no_std compatibility
```
cason, line 11: Because of `derivative` import.

romac: Let's not add non-essential dependencies to the core crate if we can help it, especially if they are not no_std compatible. A manual PartialEq impl is preferable imho.

Perhaps don't even need `PartialEq` on state? Worth looking into.

romac: Ah we need for tests, then let's do a manual impl

code/crates/core-state-machine/src/state.rs
```diff
@@ -73,12 +73,33 @@ where
     /// It may be different, lower or higher, than the state machine round.
     pub decision: Option<RoundValue<Ctx::Value>>,
 
+    /// Timeouts already scheduled for the current round.
+    /// Intended to avoid scheduling the same timeout multiple times.
+    pub scheduled_timeouts: [bool; 3],
```
romac: I just realized that derive-where supports skipping fields for some traits, so we can maybe just use


```suggestion
    #[derive_where(skip(EqHashOrd))]
    pub scheduled_timeouts: [bool; 3],
```

?

code/crates/core-state-machine/src/state.rs
```diff
@@ -27,6 +27,53 @@ impl<Value> RoundValue<Value> {
...
+            debug_assert!(false, "Only Propose, Prevote, and Precommit timeouts should be checked here. Got: {timeout:?}");
+
+            // Untracked timeouts are not scheduled and always return false.
+            false
+        }
+    }
+
+    /// Helper to map a `TimeoutKind` to its specific bitmask.
+    /// Returns `None` for timeouts that are not tracked per-round.
+    const fn mask(timeout: TimeoutKind) -> Option<u8> {
```
cason, line 67: An alternative solution is to use `Step` as index, see: https://github.com/cason/malachite/pull/5

code/crates/core-state-machine/src/state_machine.rs
```diff
@@ -293,6 +293,7 @@ where
             );
             debug_trace!(state, Line::L18);
 
+            debug_assert!(state.check_timeout(TimeoutKind::Propose));
```
nenadmilosevic95, line 296: This line will be stripped entirely in release builds since it's inside a `debug_assert!`?

romac: Yes

nenadmilosevic95: I guess it is fine because it is a propose timeout, although it will be nicer if it is consistent with prevote and precommit. But not critical anyway

PR #1512, opened 2026-02-27
chore(ci): Configure trusted crate publishing

jscaltreto:
Closes: #1513

This configures trusted publishing of crates to crates.io

The version of the rust-lang/crates-io-auth-action action corresponds to the latest version, [v1.0.3](https://github.com/rust-lang/crates-io-auth-action/releases/tag/v1.0.3).

Ref: https://crates.io/docs/trusted-publishing

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -41,7 +41,6 @@
 - Add network metrics for peer identification and tracking
 - Add transport level connection limits
 - Limit the number of peers that can connect from same IP address
-- Update `libp2p-scatter` to v0.4.0-rc.1 ([#1473](https://github.com/circlefin/malachite/pull/1473))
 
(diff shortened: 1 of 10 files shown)


.github/workflows/publish.yml
```diff
@@ -10,15 +10,20 @@ jobs:
   publish:
     name: Publish to crates.io
     runs-on: github-hosted-large
+    permissions:
+      id-token: write
```
romac, line 14: I believe we also need to create a specific environment to use for running the action and specify it here:

```suggestion
    environment: Publish
    permissions:
      id-token: write
```

romac: Ah no, re-reading the crates.io docs, it seems to be optional.

PR #1543, opened 2026-04-01
fix(sync): reject sync responses with non-contiguous heights

bastienfaivre:
Closes: #1541

## Summary

When a sync response contains values with non-contiguous certificate heights (e.g., request 1..=10, response has certificates for heights 1,2,5..12), the existing validation passes because `end_height` is computed from `start_height + values.len() - 1` without inspecting actual certificate heights. Values get buffered by their real certificate height, causing sync to get stuck when consensus advances to a missing height and finds nothing in the queue.

This PR adds a contiguity check in `on_value_response()` that verifies each value's `certificate.height == start_height + index` before processing. Non-contiguous responses are rejected and re-requested from a different peer.

The check only reads `certificate.height` (a direct struct field access) — no value decoding is needed.

## Tests

- **`test_validate_value_response_heights`**: Unit test on the extracted `validate_value_response_heights()` function. Covers valid contiguous responses, empty responses, gaps in heights, duplicate heights, wrong start height, and reversed order.
- **`test_non_contiguous_response_rejected_by_sync_handler`**: Handler-level test that drives the sync state machine coroutine with a crafted non-contiguous response (request 1..=10, get heights 1,2,5..12). Verifies the response triggers a `SendValueRequest` effect (re-request to another peer) and no `ProcessValueResponse` effect (response not forwarded to consensus).

---

#### Contribution eligibility

- [x] I am a core contributor, OR I have been explicitly assigned to the linked issue
- [x] I have read [CONTRIBUTING.md](/CONTRIBUTING.md) and my PR complies with all requirements
- [x] I understand that PRs not meeting these requirements will be closed without review

#### For all contributors

### DIFF
--- a/code/crates/sync/src/handle.rs
+++ b/code/crates/sync/src/handle.rs
@@ -149,7 +149,28 @@ where
         return on_invalid_value_response(co, state, metrics, request_id, peer_id).await;
     }
 
+    if !validate_value_response_heights::<Ctx>(&response) {
+        warn!(
+            %request_id, %peer_id,
+            "Response contains non-contiguous heights"
+        );
+
+        return on_invalid_value_response(co, state, metrics, request_id, peer_id).await;
+    }
+
     on_valid_value_response(co, state, metrics, request_id, peer_id, response).await
+}
+
+/// Validate that each value in the response has the expected height,
+/// ie. heights are contiguous starting from `start_height`.
+fn validate_value_response_heights<Ctx>(response: &ValueResponse<Ctx>) -> bool
+where
+    Ctx: Context,
+{
+    response.values.iter().enumerate().all(|(i, value)| {
+        let expected = response.start_height.increment_by(i as u64);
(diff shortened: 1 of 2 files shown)


code/crates/sync/src/handle.rs
```diff
@@ -1248,4 +1275,183 @@ mod tests {
...
+                make_raw_decided_value(11),
+                make_raw_decided_value(12),
+            ],
+        );
+
+        let input = Input::ValueResponse(request_id, peer_a, Some(response));
+        let metrics = Metrics::default();
+
+        // Drive the sync handler through the coroutine
+        let mut gen = crate::co::Gen::new(|co| handle(co, &mut state, &metrics, input));
```
romac: Do you think we could write this test using the process! macro instead of relying on the internals of the coroutine system?

romac (approved):

PR #1560, opened 2026-05-19
ci(pr-gate): bug fixes, vouch-style allow/deny, extracted scripts

romac:
Closes: #XXX

Reworks the PR Gate workflow that auto-closes external PRs lacking issue assignment. Preserves the existing contribution model (claim an issue, then PR) but fixes correctness bugs in the gate, borrows two ideas from [mitchellh/vouch](https://github.com/mitchellh/vouch), and extracts the inline workflow JavaScript into testable modules with CI coverage.

## Summary

**Bug fixes in existing checks**

- **CODEOWNERS team handles** were truncated by the `/@[\w-]+/g` regex — `@org/team` silently became `@org`, treating the org name as a codeowner username. Now skipped via negative lookahead until proper team expansion is needed.
- **`/assign` composition** returned on the first codeowner comment containing `/assign`, so codeowner A's later `/assign @alice` would invalidate codeowner B's earlier `/assign @bob`. And `/assign @alice @bob` in one comment only registered `@bob` (last match wins inside the comment). Now: union across every codeowner comment and every `/assign` occurrence.

**New capabilities (borrowed from vouch)**

- **`.github/DENOUNCED`** — flat file listing usernames whose PRs are auto-closed regardless of any other signal. First step in the ladder. Supports optional `: reason` that surfaces in the close comment.
- **`.github/VOUCHED`** — flat file of trusted external contributors who bypass the issue-assignment requirement. Complements (does not replace) the issue-claim flow.
- **Repo-collaborator bypass** — `repos.checkCollaborator` now runs before the org-membership check. Catches outside collaborators with explicit repo access (read/triage/write/admin) who aren't necessarily org members.

**Hardening / cleanup**

- Declared explicit workflow `permissions:` (contents: read, issues: write, pull-requests: write) so a future repo-wide default downgrade doesn't silently break the close step.
- Moved four inline rejection-message heredocs to `.github/pr-gate-messages/*.md` with `{{placeholder}}` substitution. Templates are still loaded from `base.ref` via the contents API for `pull_request_target` safety.
- Side effect of the template move: the previous messages embedded 12 spaces of indent in YAML/JS, which GitHub's markdown renderer treats as code blocks. New templates are left-aligned, so rejection comments are readable.
- Consolidated the eight-step `if:` ladder into one `Evaluate eligibility` script and one `Close ineligible PR` step gated by `steps.evaluate.outputs.status == 'close'`. Adding a future check is now one block in one script instead of inserting a step plus editing every subsequent step's condition.

**Extracted scripts + tests**

- Moved the two inline `actions/github-script` bodies into `.github/scripts/pr-gate/{evaluate,close}.js` as CommonJS modules. The workflow sparse-checks out only that directory and `require()`s each module. `actions/checkout` under `pull_request_target` defaults to the base ref, so the `require()`'d code is base-trusted; trust-input files (DENOUNCED/VOUCHED/CODEOWNERS/templates) are still fetched via the REST API against `base.ref` for defense in depth.
- Added `evaluate.test.js` (27 cases) and `close.test.js` (7 cases) using Node 20's built-in `node:test` runner — no `package.json`, no deps. Tests cover every branch of the decision tree, including the specific regression cases for the CODEOWNERS regex and `/assign` composition fixes.
- Added `.github/workflows/pr-gate-scripts.yml` to run `node --test` on changes to the gate workflow or its scripts. Read-only permissions, path filters so unrelated PRs don't trigger it.

## Net change

`pr-gate.yml`: 307 → 55 lines. Total workflow + scripts + tests + templates + lists: a small net increase, but the change is now structured code with unit coverage instead of a 300-line YAML heredoc.

## Commits (logical order, one per fix)

| commit | what |
|---|---|
| `e1021a2f` | declare explicit workflow permissions |
| `294928ee` | skip team handles in CODEOWNERS parsing |
| `5cb9ffb5` | compose `/assign` across codeowners and within comments |
| `c4abb3f0` | extract rejection messages into template files |
| `eb2d9b71` | allow outside collaborators with explicit repo access |
| `a16e2a0e` | add denounce list to auto-close known bad actors |
| `d8c04ba8` | add vouched allowlist for trusted external contributors |
| `f8d3ec8e` | consolidate eligibility ladder into a single script |
| `7332f7c7` | extract workflow scripts to `.github/scripts/pr-gate/` |
| `22d1d6e8` | add tests for `evaluate.js` and `close.js` |
| `c725d08a` | add CI to run the pr-gate script tests |

## Test plan

- [x] `node --test` from `.github/scripts/pr-gate/` — 34 tests pass
- [x] `python3 -c 'import yaml; yaml.safe_load(...)'` on both workflow files — valid YAML
- [ ] Once merged, watch the first external PR exercise the gate and confirm the close comment renders cleanly (no more accidental code-block indentation)
- [ ] Verify `pr-gate-scripts.yml` runs on a follow-up PR touching any file under `.github/scripts/pr-gate/`

## Out of scope (worth a follow-up)

- Team-handle expansion in CODEOWNERS — currently skipped; could be resolved via `orgs.listMembersForTeam` if the project ever moves to team-based ownership.
- `/unassign @user` syntax — revocation today is "close the PR".

---

#### Contribution eligibility

- [x] I am a core contributor, OR I have been explicitly assigned to the linked issue
- [x] I have read [CONTRIBUTING.md](/CONTRIBUTING.md) and my PR complies with all requirements
- [x] I understand that PRs not meeting these requirements will be closed without review

#### For all contributors

### DIFF
--- a/.github/DENOUNCED
+++ b/.github/DENOUNCED
@@ -1,0 +1,12 @@
+# Denounced contributors.
+#
+# Users listed here have their pull requests automatically closed by the
+# PR Gate workflow regardless of any other signal (codeowner status, org
+# membership, repo collaborator role, issue assignment, /assign tag).
+#
+# Format: one GitHub username per line, with an optional reason after a
+# colon. Comments start with `#`. Blank lines are ignored.
+#
+# Example:
+#   baduser
+#   spammer: opens AI-generated PRs that ignore review feedback
--- a/.github/VOUCHED
+++ b/.github/VOUCHED
@@ -1,0 +1,18 @@
+# Vouched contributors.
+#
+# Users listed here bypass the issue-assignment requirement of the PR
+# Gate workflow. Their pull requests are admitted as if they were org
+# members or repo collaborators. This is the path for trusted external
+# contributors who have demonstrated they understand the project well
+# enough that requiring a fresh issue claim for every PR is overhead.
+#
+# Codeowners, repo collaborators, and circlefin org members are already
+# trusted through earlier checks — they do not need to be vouched here.
+#
+# Format: one GitHub username per line, with an optional reason after a
+# colon for maintainer reference. Comments start with `#`. Blank lines
+# are ignored.
+#
+# Example:
+#   trusted-contributor
+#   long-time-helper: maintainer of the upstream foo crate
--- a/.github/pr-gate-messages/default.md
+++ b/.github/pr-gate-messages/default.md
@@ -1,0 +1,7 @@
+Hi @{{prAuthor}},
+
+Thank you for your interest in contributing to Malachite.
+
+This PR has been automatically closed because it does not meet our contribution requirements.
+
+Please see our [CONTRIBUTING.md](https://github.com/{{owner}}/{{repo}}/blob/main/CONTRIBUTING.md) for details on how to contribute properly.
--- a/.github/pr-gate-messages/denounced.md
+++ b/.github/pr-gate-messages/denounced.md
@@ -1,0 +1,5 @@
+Hi @{{prAuthor}},
+
+This PR has been automatically closed because you are listed in this repository's [`DENOUNCED`](https://github.com/{{owner}}/{{repo}}/blob/main/.github/DENOUNCED) file.{{reasonBlock}}
+
+If you believe this is in error, please reach out to the maintainers.
--- a/.github/pr-gate-messages/issue_not_found.md
+++ b/.github/pr-gate-messages/issue_not_found.md
@@ -1,0 +1,12 @@
+Hi @{{prAuthor}},
+
+Thank you for your interest in contributing to Malachite.
+
+This PR has been automatically closed because the referenced issue could not be found. Please ensure you reference a valid, existing issue using the format `Closes: #XXX`.
+
+**To contribute properly:**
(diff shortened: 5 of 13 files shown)


osr21:
Two observations from reading through the full implementation:

**`parseCodeowners()` runs the regex on comment lines**

`parseCodeowners()` applies `CODEOWNER_HANDLE_REGEX` to the raw file content before stripping `#` comment lines. A CODEOWNERS file with commented-out names — e.g. `# Former maintainer: @alice` or `# Reviewed by: @bob` — would silently add those users to the codeowner set, letting them `/assign` others and tag external PRs as eligible.

The current CODEOWNERS (`* @ancazamfir @romac`) has no comments, so there's no current impact. But it's a cheap fix to future-proof:

```js
function parseCodeowners(content) {
  if (!content) return [];
  const uncommented = content
    .split('\n')
    .filter(line => !line.trimStart().startsWith('#'))
    .join('\n');
  return (uncommented.match(CODEOWNER_HANDLE_REGEX) || []).map(c => c.substring(1).toLowerCase());
}
```

**`Closes: #XXX` placeholder in this PR's own body**

The PR description has an unfilled issue reference. If submitted by an external contributor, the gate being fixed here would close it for `no_issue_reference`. Worth replacing with a real issue number (or `N/A` with a note that this is an internal workflow change) before undrafting — partly to close the loop on the checklist item, partly to avoid the gate exercising the fix on its own author.

